PR #50, opened 2026-05-20
fix(tokens): emit shadow primitives in bundled @nexus/tailwind

aishvaryagupta:
## Summary

- Bundled `generate-tailwind-package.js` now recognizes themed primitive modes (`-light` / `-dark` pairs introduced by #34). Shadow tokens for the selected mode emit into `variables.css` as `:root` declarations (light) plus a `.dark` override block.
- File-existence invariant replaces the silent `log.warn`-then-skip: missing primitive files throw instead of producing a broken bundle.
- `partitionThemedModes` hoisted from `generate-modular.js` to shared `utils.js` so both pipelines use the same partitioning.
- One-character typo fix: `shadows.json:18` `xs.spread` referenced `{2xs.layer-1.spread}` — corrected to `{xs.layer-1.spread}`.

After this PR, `nx:shadow-*` utilities render real shadows everywhere they're used (Card, Dialog, Switch, Dropdown, Select, Input). Before, all 42 `var(--nx-shadow-*)` references in `nexus.css` resolved to empty — components silently rendered flat.

## GitHub Issue

Closes #38

## Test Plan

- [x] `yarn tokens:tailwind` runs without warnings; `packages/tailwind/variables.css` gains 160 `--nx-shadow-*` declarations (80 light in `:root`, 80 dark in `.dark`)
- [x] `yarn test:unit` — 65 tests pass (52 existing + 13 new)
- [x] New `generate-tailwind-package.test.js` symmetrically asserts every `var(--nx-shadow-*)` in `nexus.css` resolves to a matching decl in `:root` of `variables.css`
- [ ] Manual: open Storybook → Card / Dialog / Switch / Dropdown / Select / Input — confirm visible shadows in light and dark themes
- [ ] Accept Chromatic visual diffs on the above stories — see note below

## Notes for reviewers

Chromatic will flag visual changes on every component story using \`nx:shadow-*\`. Those diffs **are** the bug fix landing (shadows now actually render) — please accept them as intentional, not unrelated drift.

PR #52, opened 2026-05-20
fix(tokens): run prettier on generator CSS output

### LINKED ISSUE #51: Token generators should run prettier on emitted CSS
## Problem

The CSS artifacts in `packages/tailwind/` (`nexus.css`, `variables.css`, etc.) are committed in **prettier-formatted** state, but the token generators (`packages/core/scripts/generate-tailwind-package.js`, `generate-modular.js`) emit **raw, unformatted** CSS.

This means running `yarn tokens:tailwind` or `yarn tokens:modular` on a clean tree produces a confusing ~56-line diff in `nexus.css` plus trailing-whitespace diffs in other files — none of which reflect actual token changes. The drift is purely a formatting mismatch.

## Reproduction

```bash
git checkout main
yarn tokens:tailwind
git diff packages/tailwind/
```

You'll see diffs like:
- `nexus.css`: multi-line `--shadow-sm: var(...) var(...), var(...) var(...);` declarations collapsed to single-line.
- `variables.css` and others: trailing blank-line differences before `}`.

Running `npx prettier --write packages/tailwind/*.css` after generation brings the artifacts back to byte-identical state with what is committed.

## Root cause

Commit `b75691d` ("updated remaining files", Jan 10 2026) ran prettier over the generator output and committed the formatted result. The generators were never updated to format their own output, so the two states have drifted ever since.

## Impact

- Anyone regenerating tokens sees a noisy diff unrelated to their changes.
- The noise can mask real changes or get accidentally committed.
- New contributors hit this footgun on first regeneration.
- Recently cost investigation time during PR #50 review-feedback work, where the drift looked (briefly) like it might have been caused by a code refactor.

## Proposed fix

In `packages/core/scripts/generate-tailwind-package.js` and `generate-modular.js`, format the emitted CSS string with the `prettier` API before writing each file to disk. Use the repo's existing `.prettierrc.json` config so the output exactly matches what `npx prettier --write` would produce.

Small follow-up to file: ~10 lines per script, plus a `prettier` import.

## Out of scope

- Doesn't touch any token values or generator logic — purely a write-time formatting step.
- Not bundled into PR #50, which is about shadow primitive emission, not formatting.

aishvaryagupta:
## Summary

- Both token generators (`generate-tailwind-package.js`, `generate-modular.js`) now run prettier in place on every emitted `.css` file before exiting, using a shared `formatDistDirectory` helper in `utils.js`.
- The committed CSS in `packages/tailwind/` had been prettier-formatted, but the generators emitted raw output — so regenerating tokens produced a noisy ~56-line whitespace diff in `nexus.css` plus trailing-newline diffs elsewhere, all of which masked real token changes and recently cost investigation time during PR #50.
- Adds a regression test that asserts every file emitted by `generateTailwindPackage` is prettier-idempotent.

## GitHub Issue

Closes #51

## Test Plan

- [x] `yarn test:unit` — 65 passed, 1 skipped (6 in `generate-tailwind-package.test.js` including the new regression case)
- [x] `yarn tokens:tailwind` on a clean tree — produces zero diff in `packages/tailwind/`
- [x] `yarn tokens:modular` on a clean tree — produces zero diff in `apps/playground/public/themes/` and `apps/playground/src/styles/`
- [x] `npx eslint packages/core/scripts/` — clean
- [x] `npx prettier --check` on touched source files — clean

### DIFF
--- a/packages/core/scripts/__tests__/generate-tailwind-package.test.js
+++ b/packages/core/scripts/__tests__/generate-tailwind-package.test.js
@@ -1,10 +1,14 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import { afterEach, beforeAll, describe, expect, it } from 'vitest';
+import * as prettier from 'prettier';
+import { fileURLToPath } from 'url';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import { generateTailwindPackage } from '../generate-tailwind-package.js';
 import { DEFAULT_CONFIG } from '../utils.js';
+
+const TEST_DIR = path.dirname(fileURLToPath(import.meta.url));
 
 const tmpDirs = [];
 
@@ -28,7 +32,7 @@ function extractBlock(css, openSelector) {
   return match[1];
 }
 
-afterEach(() => {
+afterAll(() => {
   while (tmpDirs.length > 0) {
     const dir = tmpDirs.pop();
     fs.rmSync(dir, { recursive: true, force: true });
@@ -41,14 +45,14 @@ describe('generateTailwindPackage', () => {
   let nexusCSS;
   let warnings;
 
-  beforeAll(() => {
+  beforeAll(async () => {
     warnings = [];
     const originalWarn = console.warn;
     console.warn = (...args) => warnings.push(args.map(String).join(' '));
 
     try {
       distDir = makeTmpDir();
-      generateTailwindPackage(DEFAULT_CONFIG, { distDir });
+      await generateTailwindPackage(DEFAULT_CONFIG, { distDir });
     } finally {
       console.warn = originalWarn;
     }
@@ -108,5 +112,28 @@ describe('generateTailwindPackage', () => {
   it('emits zero `File not found` warnings for the default config', () => {
     const fileNotFound = warnings.filter((w) => /File not found/.test(w));
     expect(fileNotFound).toEqual([]);
+  });
+
+  // Regenerating tokens used to produce a noisy whitespace diff against the
+  // committed CSS because the generator emitted raw output but the committed
+  // files were prettier-formatted. The generator now formats every emitted
+  // file, so prettier --check must agree it has nothing to change.
+  it('emits prettier-formatted CSS (idempotent under prettier)', async () => {
+    const config = await prettier.resolveConfig(TEST_DIR);
+    const files = fs
+      .readdirSync(distDir)
+      .filter((name) => name.endsWith('.css'));
+
+    expect(files.length).toBeGreaterThan(0);
+
+    for (const name of files) {
+      const filePath = path.join(distDir, name);
+      const content = fs.readFileSync(filePath, 'utf8');
+      const reformatted = await prettier.format(content, {
+        ...config,
+        filepath: filePath,
+      });
+      expect(reformatted, `${name} is not prettier-formatted`).toBe(content);
+    }
   });
 });
--- a/packages/core/scripts/utils.js
+++ b/packages/core/scripts/utils.js
@@ -1,5 +1,7 @@
 import fs from 'fs';
 import path from 'path';
+import * as prettier from 'prettier';
+import { fileURLToPath } from 'url';
 
 import {
   hexToOklchMechanical,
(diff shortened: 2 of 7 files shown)


packages/core/scripts/__tests__/generate-tailwind-package.test.js
```diff
@@ -109,4 +113,27 @@ describe('generateTailwindPackage', () => {
     const fileNotFound = warnings.filter((w) => /File not found/.test(w));
     expect(fileNotFound).toEqual([]);
   });
+
+  // Regenerating tokens used to produce a noisy whitespace diff against the
+  // committed CSS because the generator emitted raw output but the committed
+  // files were prettier-formatted. The generator now formats every emitted
+  // file, so prettier --check must agree it has nothing to change.
+  it('emits prettier-formatted CSS (idempotent under prettier)', async () => {
```
examlly-tech, line 121: **Minor — coverage gap:** This regression guards `generateTailwindPackage` only. `generate-modular.js` received the same fix but has no equivalent test, so the same drift could re-emerge there silently. Export `main` from generate-modular and add a sibling test.

packages/core/scripts/utils.js
```diff
@@ -1115,3 +1117,26 @@ export function generateBaseLayerCSS() {
...
+
+/**
+ * Format every `.css` file in `distDir` in place with prettier, using the
+ * repo's `.prettierrc`. Resolves the config from the script's own location so
+ * callers can write to a temporary dist (e.g. tests) without losing config
+ * resolution.
+ */
+export async function formatDistDirectory(distDir) {
+  const config = await prettier.resolveConfig(SCRIPTS_DIR);
+  const files = fs.readdirSync(distDir).filter((name) => name.endsWith('.css'));
```
examlly-tech: **Minor — non-recursive walk:** `readdirSync` returns only direct children; any future nested layout under `distDir` would be silently skipped. The helper's name implies 'format every CSS in dist'. Walk recursively or assert flatness.

examlly-tech (commented):

### Issues

|---|----------|------|-------|------------|
| 1 | ⚠️ Minor | `packages/core/scripts/__tests__/generate-tailwind-package.test.js:121` | The prettier-idempotency regression covers only the bundled generator; `generate-modular.js` got the same fix but has no equivalent test. | Add a sibling test that runs the modular generator against a temp dir and asserts every emitted file is prettier-idempotent. |
| 2 | ⚠️ Minor | `packages/core/scripts/utils.js:1131` | `formatDistDirectory` is non-recursive — it silently skips any nested subdirectory if the dist layout ever grows one. | Walk recursively or assert flatness explicitly. |

### Architectural Concerns

**Undeclared dependency on `prettier`** — `packages/core/scripts/utils.js` imports `prettier`, but `packages/core/package.json` declares no dependency on it. Resolution works today only via workspace hoist from the root devDependency, so `@nexus/core` cannot build in isolation; declare `prettier` in the package's devDependencies to make the contract self-contained.

**Generator-internal formatting vs root `yarn format`** (`packages/core/scripts/utils.js:1129`) — Confirm in-generator prettier is the intended ownership boundary: it can't be bypassed, but it duplicates the responsibility already covered by the root `yarn format` script and adds prettier as a build-time dep of `@nexus/core`.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/52
- **GitHub Issue:** #51

packages/core/scripts/utils.js
```diff
@@ -1115,3 +1117,26 @@ export function generateBaseLayerCSS() {
...
+
+const SCRIPTS_DIR = path.dirname(fileURLToPath(import.meta.url));
+
+/**
+ * Format every `.css` file in `distDir` in place with prettier, using the
+ * repo's `.prettierrc`. Resolves the config from the script's own location so
+ * callers can write to a temporary dist (e.g. tests) without losing config
+ * resolution.
+ */
+export async function formatDistDirectory(distDir) {
```
examlly-tech: **Minor — naming:** the function formats `.css` files inside the directory rather than the directory itself; `formatDistCssFiles(distDir)` reads truer to behavior.

examlly-tech (commented):
## SDE2 Review — Verdict: COMMENT

### Issues

|---|----------|------|-------|------------|
| 1 | ⚠️ Minor | `packages/core/scripts/utils.js:1129` | `formatDistDirectory` formats files inside the directory, not the directory itself. | Rename to `formatDistCssFiles` to match what the function does. |

### Verified

- Prettier 3.7.4 API usage (`resolveConfig` + `format` with `filepath`) is correct.
- `resolveConfig` returning `null` (e.g. if `.prettierrc` were removed) spreads safely into `format`; falls back to prettier defaults.
- `afterEach` → `afterAll` switch is necessary because the new prettier test reads from disk inside `it`; existing tests remain read-only — no isolation regression.
- Top-level `await main()` in `generate-modular.js` requires Node ≥ 14.8 (repo pins Node 20 via `.nvmrc`).
- `dist/modular` and `dist/tailwind` have no subdirectories — flat `readdirSync().filter('.css')` is sufficient today.
- End-to-end: `yarn tokens:tailwind` and `yarn tokens:modular` both produce zero diff against committed CSS; all 65 unit tests pass.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/52
- **GitHub Issue:** #51

examlly-tech (approved):

### Unresolved Issues

All previous architectural issues resolved.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/52
- **GitHub Issue:** #51

examlly-tech (approved):

### Unresolved Issues

All previous issues resolved.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/52
- **GitHub Issue:** #51

PR #59, opened 2026-05-21
chore(workspace): fix Node/vitest/tsconfig/turbo/prettier/husky drift

aishvaryagupta:
## Summary

Seven independent config-drift fixes that compound:

- **Node pin** unified via `node-version-file: '.nvmrc'` on all 10 CI jobs; `package.json` engines + `.nvmrc` anchored at `20.19.0`
- **vitest** unified to `4.0.18` (test-utils was pinned at `4.0.17`); 84-line lockfile dedup
- **tsconfig project references** expanded to include `packages/test-utils` (composite added); `tsc --build --dry` from root now covers both TS-bearing packages
- **`noUncheckedIndexedAccess`** enabled at `tsconfig.base.json`; 14 type errors fixed across 4 story files by narrowing `Record<string, X>` shapes to literal keyset unions
- **`apps/playground/tsconfig.json`** now extends `tsconfig.base.json`, bringing the playground under the strict graph
- **`turbo.json`** dead `.next/**` outputs stripped from `build` task
- **`.prettierignore`** now excludes auto-generated CSS at three paths (12 files total)
- **`.husky/pre-commit`** stripped to a single `npx lint-staged` call — husky v9's wrapper already auto-sources `~/.config/husky/init.sh`

## Locked decisions (worth flagging)

| Decision | Choice | Why |
|----------|--------|-----|
| `engines.node` shape | kept `">=20.19.0"` (already canonical) | Preserves semver semantics for engines; CI now reads `.nvmrc` directly so the three sources (`.nvmrc`, CI, `engines.node`) collapse to one source-of-truth read at runtime. |
| tsconfig references count | **2** (`react` + `test-utils`), not literal "all four" | `packages/core` is `.js` build scripts only; `packages/tailwind` is pure CSS. Adding empty-graph tsconfigs for them is anti-quality noise. |
| `exactOptionalPropertyTypes` | **skipped** | Issue body mentions both flags, but acceptance criteria narrows to `noUncheckedIndexedAccess`. Belongs in its own PR with its own type-error ripple. |
| Other root scripts in turbo | **kept outside** | `format`, `test:*`, `playground`, `chromatic`, `audit:*`, `tokens:*` don't fit turbo's input/output cache model. Adding declarations that no script invokes would be the same dead-config anti-pattern we're removing. |

## Type fixes detail (Phase 4)

All 14 errors resolved by tightening types so indexed access is type-safe by construction — cleaner than scattering `!` assertions:

[...]

PR #53, opened 2026-05-20
fix(tokens): pipeline polish — schema regex, audit pairs, rounding, dedup

aishvaryagupta:
## Summary

Five small but real correctness fixes in the design-token pipeline:

- **Audit-contrast pairs** — added `muted-light-foreground ↔ muted-light` and `disabled-foreground ↔ disabled` to `BASE_PAIRS` at the existing Lc 45 incidental tier. Per `tokens.md` thresholds are non-negotiable, so failures were fixed by re-pointing semantic references:
  - LIGHT `disabled-foreground`: `shade.300` → `shade.400` (all 5 base palettes)
  - DARK `muted-light-foreground`: `shade.500` → `shade.300` (all 5 base palettes)
  - DARK `disabled-foreground`: `shade.400` → `shade.300` (all 5 base palettes)
  - DARK `muted-light` (bg): `shade.800` → `shade.900` (all 5 base palettes) — mirrors the LIGHT pattern where `muted-light` = `background`, restoring distinction from `muted` (which stays at `shade.800`).
  - `BASE_PAIRS`/`BRAND_PAIRS` switched from positional `[text, bg, minLc]` tuples to `{text, bg, minLc, tier}` objects so the APCA tier (`body`/`ui`/`incidental`) travels with each pair instead of living only in `tokens.md`.
- **Float-32 rounding** — `formatTokenValue`'s dimension branch now rounds `value.value` to 4 decimals, stripping Figma's `-0.800000011920929` export junk to `-0.8`. Source JSON deliberately left as-imported so future Figma re-exports don't regress on read.
- **`line-height: auto` → `normal`** — invalid CSS mapped to a well-formed default inside `resolveTypographyProperty`. `typography-code-inline` now emits `line-height: normal`.
- **Resolver dedup** — `resolveReferenceWithNxPrefix` and `resolveValueWithNxPrefix` were byte-equivalent to their non-prefixed counterparts (the `nx-` was baked into the primitiveMap `cssName`). Deleted both; rewired callers to `resolveValue`; `resolveTypographyProperty` collapsed from 16 lines to 3.
- **Dormant schema removed** — `packages/core/tokens.schema.json` was never read by any script. Deleted along with stale references in `.claude/rules/tokens.md` and `packages/core/scripts/TESTING_ACTION_PLAN.md`.

`yarn audit:contrast` passes 100/100 pairs. Generated CSS in `packages/tailwind/` and `apps/playground/` regenerated for the new references.

## GitHub Issue

Closes #45

[...]

Issue #45, opened 2026-05-20
Token pipeline polish: schema regex, audit-contrast pairs, letter-spacing rounding, line-height: auto


aishvaryagupta:
## Summary

Smaller pipeline issues that the structural shadow fix (PR A) doesn't cover:

- **`tokens.schema.json:7,60` regex rejects every shipped key** that starts with a digit (`50`–`950`, `0_5`, `2xs`, `2xl`, `3xl`). The "validator" the rules cite validates nothing.
- **`audit-contrast.js:20-27` missing two foreground pairs** that `tokens.md:48-52` documents (`muted-light-foreground ↔ muted-light`, `disabled-foreground ↔ disabled`). May be silently failing accessibility.
- **Letter-spacing float artifacts** — `typography-vega.json` and `typography-lyra.json` ship Figma export junk like `-0.800000011920929px`. Propagates verbatim into `variables.css:341-346`. Belongs in the generator (`utils.js` should round before emit).
- **`line-height: auto`** emitted via `utils.js:604` from `typography.json:195` for `typography-code-inline`. Invalid CSS — browsers ignore it. Map `auto` → `normal` in the generator, or fix the source token.
- **Two byte-equivalent reference resolvers** at `utils.js:171-209` (`resolveReference` vs `resolveReferenceWithNxPrefix`) — delete one.

## Scope clarification — `audit-contrast.js` ownership (LOCKED IN)

This PR is the **sole place** `packages/core/scripts/audit-contrast.js` is modified across the audit-followup work. PR E (#42) was originally going to share this fix; the scope has been split so that **only this PR touches the script** and **only PR E touches the CI YAML**.

**What this PR does to `audit-contrast.js`:**

- Adds the two missing foreground pairs to `BASE_PAIRS` (lines 20-27):
  - `muted-light-foreground ↔ muted-light`  →  `|Lc| ≥ 45` (incidental tier)
  - `disabled-foreground ↔ disabled`  →  `|Lc| ≥ 45` (incidental tier)
- If those pairs newly fail APCA after being added, fix them by either:
  - Adjusting the semantic-token reference (which shade fills which role) in `base-{palette}-{theme}.json`, **or**
  - Tuning the perceptual L-grid values in `packages/core/scripts/lib/perceptual-grid.json`
- **Do not lower the APCA thresholds.** The 75/60/45 tiers come from APCA's published guidance and are not negotiable per-finding (see `.claude/rules/tokens.md` "APCA contrast gate").

[...]

Issue #43, opened 2026-05-20
Workspace config drift: Node pinning, vitest versions, tsconfig graph, turbo coverage, prettierignore


aishvaryagupta:
## Summary

Multiple small config-drift issues that compound:

- **Node version pinning**: `.nvmrc: 20`, `package.json engines: ">=20.19.0"`, CI `'20'` — three pins, none agreeing precisely
- **`vitest` version**: root pins `4.0.18`, `packages/test-utils` pins `4.0.17`
- **Root `tsconfig.json`** references only `packages/react`; `packages/core`, `packages/tailwind`, `packages/test-utils` are invisible to `tsc --build` from root
- **`tsconfig.base.json`** is `strict: true` but missing `noUncheckedIndexedAccess` and `exactOptionalPropertyTypes` (both valuable for a library shipping to consumers)
- **`turbo.json`** has dead Next.js outputs (no Next.js in repo) and most root scripts (`audit:contrast`, `tokens:tailwind`, `tokens:modular`, `playground`, `chromatic`, `format`, `test:*`) aren't declared — bypass turbo orchestration entirely
- **`.prettierignore`** doesn't exclude auto-generated CSS at `packages/tailwind/*.css`, `apps/playground/public/themes/*.css`, `apps/playground/src/styles/*.css` — every `prettier --write` reformats them despite "DO NOT EDIT" headers
- **`.husky/pre-commit`** hardcodes `~/.nvm/nvm.sh` — breaks for Volta, fnm, asdf, system Node users

## Acceptance criteria

- [ ] One canonical Node pin (likely `20.19.0`), referenced consistently by `.nvmrc`, `package.json engines`, and CI
- [ ] `vitest` version unified across workspace (single canonical pin)
- [ ] Root `tsconfig.json` references all four packages via project references
- [ ] `noUncheckedIndexedAccess` enabled at the base config
- [ ] `turbo.json`: remove dead Next.js outputs; add tasks for scripts that benefit from caching/parallelism
- [ ] `.prettierignore` covers all three generated-CSS paths
- [ ] `.husky/pre-commit` works without nvm specifically

## Audit reference

W-34, W-35, W-36, W-37, W-38, W-39, W-40 — see `reports/audit-report.html`

aishvaryagupta:
## ✅ Done · merged in PR #59

[...]

PR #64, opened 2026-05-21
ci: enforce quality gates promised by .claude/rules/

### LINKED ISSUE #42: CI doesn't enforce quality gates promised by .claude/rules/ (contrast, visual regression, lint paths)
## Summary

The rules promise CI enforcement of gates that CI never runs:

| Gate | Rule reference | Local script | CI status |
| --- | --- | --- | --- |
| APCA contrast | `tokens.md` | `yarn audit:contrast` | **Missing** |
| Visual regression | `storybook.md` | `yarn chromatic:ci` | **Missing** |
| Regenerate-and-diff (token CSS) | implied | — | **Missing** |
| Lint | `code-quality.md` | `yarn lint` | Partial — diff-only, diverges from local |
| Format | Prettier config | `yarn format` | Partial — diff-only |

**Also:** `.github/workflows/ci.yml:255-296,298-346` has two CI jobs (`test-storybook` and `test-coverage`) that do nearly the same thing with separate Playwright installs. Consolidate.

**And:** CI runs `npx eslint` directly on changed files. Local `yarn lint` runs `turbo run lint`, which only lints the `react` package (only package with the script). Two diverging lint paths.

## Scope clarification — `audit-contrast.js` is NOT modified here (LOCKED IN)

The script-level fix for `audit-contrast.js` (adding the two missing APCA foreground pairs) is **owned entirely by PR H (#45)**. This PR only wires `yarn audit:contrast` into CI.

**Sequencing:** PR H (#45) must merge **before** this PR starts. Verify on `main` that `audit-contrast.js` `BASE_PAIRS` contains the `muted-light-*` and `disabled-*` pairs, and that `yarn audit:contrast` exits 0 locally, before opening this PR.

**Also:** this PR conflicts with **PR F (#43)** on `.github/workflows/ci.yml` (F updates the Node version pin; this PR adds new jobs). PR F should land first; this PR rebases against F's YAML changes.

## Acceptance criteria

- [ ] PR H (#45) has merged; `yarn audit:contrast` exits 0 on `main`
- [ ] PR F (#43) has merged; CI YAML's Node pinning is settled
- [ ] CI runs `yarn audit:contrast` on every PR; fails on threshold violation
- [ ] CI runs `yarn chromatic:ci` on PRs (informational on main)
- [ ] CI runs a regenerate-and-diff step: `yarn tokens:tailwind && git diff --exit-code` — catches a regression like the PR A bundled-shadow issue
- [ ] `test-storybook` and `test-coverage` jobs consolidated (single Playwright install)
- [ ] Local `yarn lint` and CI lint hit the same files with the same rules

## Audit reference

C-03, W-01, W-41, W-42; Figure 6.1 (CI gate matrix) — see `reports/audit-report.html`

aishvaryagupta:
## Summary

- **Lint + format alignment** — local `yarn lint`/`yarn format:check` and the corresponding CI jobs now invoke identical commands against the same files. Dropped per-package `lint` scripts and the turbo `lint` task; root `yarn lint` is `eslint . --max-warnings 0`. CI jobs collapsed from diff-based scripts to one-liners.
- **Fixed broken regenerate-and-diff** — `audit-tokens` job ran `yarn workspace @nexus/core build:tailwind` (writes to `packages/core/dist/tailwind/`, untracked) then checked `git diff -- packages/tailwind` (tracked). The build step never touched the diffed directory, so the check passed vacuously and never caught stale token CSS. Switched to `yarn tokens:tailwind` (root script does build + copy); also extended to cover `yarn tokens:modular` outputs in `apps/playground/**` and uses `git status --porcelain` to catch untracked new files.
- **Consolidated Playwright test jobs** — removed `test-storybook`. `test-coverage` already exercises both `unit` and `storybook` vitest projects under Playwright with coverage. `test-unit` (jsdom, no Playwright) stays as the fast feedback path.
- **Removed chromatic entirely** (revert of an earlier phase) — visual regression is no longer in scope for this PR. The corresponding AC on the issue is dropped (see issue comment). Chromatic CI job, scripts, devDeps, storybook addon, config file, modes file, story `chromatic: { ... }` parameters, and Chromatic sections in `.claude/rules/storybook.md`, `.claude/rules/testing-react.md`, and `CONTRIBUTING.md` all removed.
- **Decoupled tests from build** — `test-unit`/`test-coverage`/`build-storybook` no longer depend on the `build` job or download build artifacts. Vitest reads `src/` and Storybook reads `src/**/*.stories.*`; only `bundle-size` needs `packages/react/dist/`. Verified locally.
- **`CI Status` aggregator job** — new single required-check job (`if: always()`, `needs:` every other job). Treats skipped jobs (filtered out by paths-filter) as success; fails when any required job failed or was cancelled. Closes the "required-check trap" where docs-only or `.claude/**`-only PRs would block on skipped checks.
- Reformatted 3 `.claude/skills/shadcn/` markdown files (Tailwind class reordering via `prettier-plugin-tailwindcss`) — they were the only previously-tracked failures surfaced by the broader format glob.

## GitHub Issue

Closes #42 (with the visual-regression AC dropped per discussion)

## Pre-merge Checklist

- [ ] **Branch protection update REQUIRED** — change the required status check on `main` to `CI Status` (the new aggregator). Drop all per-job required-check entries (`Format Check`, `Lint`, `Build`, `TypeScript Check`, `Unit Tests`, `Audit Tokens`, `Test Coverage`, `Build Storybook`, `Bundle Size`). The aggregator is the single point of contact for branch protection going forward; failures in any individual job propagate up through it.

## Follow-up tracked issues

- #66 — model `core → tailwind` workspace edge via turbo, drop fake devDep on `@nexus/react`
- #67 — fix `test-coverage` instrumentation (reports 0% vs 70% threshold) OR rename the job + drop thresholds
- #68 — restore a fast-fail browser-mode story-test signal after the `test-storybook`/`test-coverage` consolidation

## Notes

- PR #43 and #45 referenced in the issue body never existed as numbered PRs; equivalent work landed under #53 (`audit-contrast` BASE_PAIRS fix) and #59 (Node/vitest/workspace drift fix). Prereqs verified by reading code (`yarn audit:contrast` passes 220/220 locally; CI uses `node-version-file: '.nvmrc'`).
- Two reports under `reports/` (`configurability-vs-discipline.md`, `surface-options.html`) remain untracked WIP and are not part of this PR.

## Test Plan

- [ ] CI `lint` job runs `yarn lint` and exits 0
- [ ] CI `format-check` job runs `yarn format:check` and exits 0
- [ ] CI `audit-tokens` job: regenerate-and-diff catches token CSS drift (verify by modifying a token JSON in a follow-up branch and watching this job fail)
- [ ] CI `test-storybook` job no longer exists; `test-coverage` runs the full surface
- [ ] No chromatic job in CI; `yarn build-storybook` and `yarn test:storybook` still pass locally
- [ ] CI `CI Status` aggregator passes on a docs-only PR (i.e., touching only `.md` files or `apps/docs/**` or `.claude/**`)


.github/workflows/ci.yml
```diff
@@ -107,31 +88,8 @@ jobs:
...
-
-          if [ -n "$CHANGED_FILES" ]; then
-            echo "Linting changed files:"
-            echo "$CHANGED_FILES"
-            echo "$CHANGED_FILES" | xargs npx eslint --max-warnings 0
-          else
-            echo "No matching files changed, skipping lint"
-          fi
+      - name: Lint
+        run: yarn lint
```
examlly-tech: **Minor — no ESLint cache:** Full-tree `eslint .` runs cold on every PR (previously per-package via turbo cache). Add `--cache --cache-location .eslintcache` and an `actions/cache@v4` step keyed on `yarn.lock` + `eslint.config.js` to keep wall time bounded as the repo grows.

.github/workflows/ci.yml
```diff
@@ -271,9 +229,9 @@ jobs:
 
       - name: Regenerate @nexus/tailwind and check freshness
         run: |
-          yarn workspace @nexus/core build:tailwind
+          yarn tokens:tailwind
           if ! git diff --exit-code -- packages/tailwind; then
-            echo "::error::packages/tailwind/ is stale — run 'yarn workspace @nexus/core build:tailwind' and commit the result."
+            echo "::error::packages/tailwind/ is stale — run 'yarn tokens:tailwind' and commit the result."
             exit 1
           fi
```
examlly-tech, line 295: **Minor — single step bundles three audits:** Diff-check, contrast, and class-refs run in sequence; first failure short-circuits the rest. Fine at current scale, but splitting into separate steps would surface independent failure signals.

examlly-tech (commented):
## Principal Architect Review — Verdict: APPROVED

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:91-92` | Full-tree `yarn lint` runs no-cache on every PR; will become the slowest job as the repo grows. | Add ESLint flat-config cache via `--cache --cache-location .eslintcache` plus `actions/cache@v4` keyed on `yarn.lock` + `eslint.config.js`. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:212-242` | `audit-tokens` runs three independent audits (`tokens:tailwind`/diff, `audit:contrast`, `audit:class-refs`) sequentially as one step group; any failure masks the others in a single run. | Acceptable for current scale; revisit as separate jobs only if audit runtime climbs. |

### Architectural Concerns

**Loss of fast Storybook signal** (`.github/workflows/ci.yml:244-292`) — `test-coverage` is now the only browser-test path and runs coverage instrumentation on every PR (slower than the deleted `test-storybook`). `test-unit` (jsdom) preserves fast feedback for hook/util changes, but component-only PRs now wait on the heavier coverage path. With `test-storybook` removed, `test-coverage` is the sole jsdom-vs-browser distinction — confirm `test-unit` stays jsdom-only so component-only PRs still have a fast signal pre-coverage.

**Full-tree lint vs. affected-package model** (`.github/workflows/ci.yml:73-92`, `package.json:13`) — Lint and format jobs are now full-tree while build/typecheck/test remain `--filter='...[origin/base]'`; the workflow has two scaling models in one graph. Intentional and correct for now (local/CI parity), but document the asymmetry so future contributors don't try to "fix" it.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.prettierignore` | New eslint ignores (`storybook-static`, `build`, `out`) not mirrored in `.prettierignore`; `format:check` still scans `storybook-static/*.json`. Currently passes only because the generated JSON happens to be prettier-valid — fragile to future storybook output changes. | Add the three patterns to `.prettierignore` to keep lint and format scopes aligned. |
| 2 | ⚠️ Minor | `docs/plans/oklch-migration.md:51,61,78` | Plan doc still references `packages/react/src/storybook/modes.ts` and `@chromatic-com/storybook` addon as "already wired"; both are deleted by this PR. A reader following the path references hits dead links. | Strike Chromatic lines from §5/§6 or add a status banner that visual regression was later dropped. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

_Inline comments would target `.prettierignore` and `docs/plans/oklch-migration.md` — both unchanged in this PR, so the Reviews API can't pin them. Findings are surfaced here instead._

.github/workflows/ci.yml
```diff
@@ -271,61 +247,20 @@ jobs:
...
-        run: npx playwright install chromium --with-deps
-
-      - name: Install Playwright system deps
-        if: steps.playwright-cache.outputs.cache-hit == 'true'
-        run: npx playwright install-deps chromium
-
-      - name: Run Storybook tests
-        run: yarn test:storybook
-
   test-coverage:
```
examlly-tech, line 309: **Architectural — branch protection coupling:** The `test-storybook` → `test-coverage` required-check rename is documented only in the PR body. If main's protection-rule list isn't updated at the exact merge moment, future PRs block on a vanished check. Coordinate the rule change with merge, or open a tracked issue.

examlly-tech (commented):
## Principal Architect Review — Verdict: APPROVED

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:233-234` | `audit-tokens` waits on `build` (`needs: [changes, build]` + `build.result == 'success'`) but never downloads its artifacts; it regenerates CSS via `yarn tokens:tailwind` and runs standalone node scripts. | Drop `build` from `needs` and remove the `build.result` gate so contrast/class-ref signal surfaces even when an unrelated React build fails. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:147-148` | `--filter='@nexus/react'` is hardcoded in `build` to compensate for `@nexus/core` not being a workspace dep of `@nexus/react`. The inline comment names the workaround. | Add `@nexus/core` as a `devDependency` of `@nexus/react` so Turbo's affected graph picks up token-CSS regeneration naturally; drop the manual filter. |

### Architectural Concerns

**Branch protection drift is an unguarded merge-time hazard** (`.github/workflows/ci.yml:264`) — Required-check rename (`test-storybook` → `test-coverage`) is described only in the PR body. If the rename isn't applied to main's protection rules at the exact merge moment, every subsequent PR blocks on a check that no longer exists. Land the protection-rule change in the same window as the merge, or open a tracked issue.

**Workspace graph doesn't model token-CSS consumption** (`packages/react/package.json`) — `@nexus/react` consumes `@nexus/tailwind` CSS produced by `@nexus/core`'s token scripts, but `core` is not a dep of `react` anywhere. The `bundle-size` workaround filter (Issue 2) is the visible symptom; the audit-tokens job's independence (Issue 1) is another. Make `@nexus/core` a `devDependency` of `@nexus/react`, or extract token-CSS generation into a dedicated workspace both depend on.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -107,31 +97,17 @@ jobs:
...
-
-          if [ -n "$CHANGED_FILES" ]; then
-            echo "Linting changed files:"
-            echo "$CHANGED_FILES"
-            echo "$CHANGED_FILES" | xargs npx eslint --max-warnings 0
-          else
-            echo "No matching files changed, skipping lint"
-          fi
+      - name: Lint
+        run: yarn lint --cache --cache-location .eslintcache
```
examlly-tech: **Minor — eslint cache inert in CI:** Default `--cache-strategy` is `metadata` (mtime), which `git checkout` invalidates on every run. The cache file restores fine but every source is re-linted. Append `--cache-strategy content` so ESLint hashes file contents and the cache actually skips unchanged files.

.github/workflows/ci.yml
```diff
@@ -271,61 +247,20 @@ jobs:
...
-          yarn workspace @nexus/core build:tailwind
+          yarn tokens:tailwind
           if ! git diff --exit-code -- packages/tailwind; then
-            echo "::error::packages/tailwind/ is stale — run 'yarn workspace @nexus/core build:tailwind' and commit the result."
+            echo "::error::packages/tailwind/ is stale — run 'yarn tokens:tailwind' and commit the result."
             exit 1
           fi
 
       - name: APCA contrast audit
+        if: ${{ !cancelled() }}
```
examlly-tech: **Minor — guard semantics:** `!cancelled()` lets these audits run even when `yarn tokens:tailwind` itself crashed mid-write, not only when the freshness diff fails. If the intent is `run audits regardless of stale-tokens result`, fine; if it's `run audits unless the regen step crashed`, gate via step `id` + `steps.<id>.conclusion == 'failure'` instead.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:110` | ESLint cache will be inert in CI. Default `--cache-strategy metadata` keys on file mtime, which `git checkout` rewrites on every run — cache restores fine, but every source is re-linted. | Append `--cache-strategy content` to the lint command so ESLint hashes file contents. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:257,261` | `if: ${{ !cancelled() }}` runs the contrast/class-ref audits even when `yarn tokens:tailwind` itself crashed (partial-write state), not only when the freshness diff fails. | If the intent is "run audits regardless of stale-tokens result", document that; if it's "only when the diff step fails", gate by step `id` + `steps.<id>.conclusion == 'failure'` instead. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -254,8 +229,8 @@ jobs:
   audit-tokens:
     name: Audit Tokens
     runs-on: ubuntu-latest
-    needs: [changes, build]
-    if: needs.changes.outputs.react == 'true' && needs.build.result == 'success'
+    needs: changes
+    if: needs.changes.outputs.react == 'true'
```
examlly-tech: **Minor — `ci` output unused here:** `audit-tokens` gates on `react` only, so a workflow-only PR that refactors this job's regenerate / freshness / audit steps skips it in CI. Add `|| needs.changes.outputs.ci == 'true'` so CI self-validates audit-pipeline changes.

packages/react/package.json
```diff
@@ -19,11 +19,11 @@
     "build": "vite build",
     "clean": "rm -rf dist",
     "typecheck": "tsc --noEmit",
-    "lint": "eslint src/",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build"
   },
   "devDependencies": {
+    "@nexus/core": "*",
```
examlly-tech, line 31: **Minor — dep doesn't reflect consumption:** Nothing in `packages/react/src` imports `@nexus/core`; this devDep exists solely to seed turbo's affected graph. Real edge: `core scripts → packages/tailwind/*.css → @nexus/tailwind → @nexus/react`. Either annotate inline or express via turbo `inputs` / `globalDependencies`.

.github/workflows/ci.yml
```diff
@@ -269,63 +244,30 @@ jobs:
...
-        if: steps.playwright-cache.outputs.cache-hit == 'true'
-        run: npx playwright install-deps chromium
-
-      - name: Run Storybook tests
-        run: yarn test:storybook
-
+  # NOTE: branch protection on `main` lists this job as a required status check
+  # (renamed from `test-storybook` when that job was consolidated into this one).
+  # If you rename or remove `test-coverage`, update the protection rule in the
+  # same change — otherwise every subsequent PR blocks on a vanished check.
```
examlly-tech: **Minor — fragile protection coupling:** The contract relies on a future renamer reading this comment. Safer: treat `test-coverage` as a stable CI public-API name (rename-freeze) or move the protection contract to a file the protection-rule owner already monitors (CODEOWNERS-style).

examlly-tech (commented):
## Principal Architect Review — Verdict: APPROVED

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:233` | `audit-tokens` gates on `react` only. A workflow-only PR refactoring the audit step (or its scripts) skips the job in CI — it self-validates only when `packages/react/**` or sibling react-graph paths change. | Add `\|\| needs.changes.outputs.ci == 'true'` to this `if:` (and the other react-only jobs) so CI-only edits self-validate. |
| 2 | ⚠️ Minor | `packages/react/package.json:26` | `@nexus/core` is declared as a devDep of `@nexus/react` purely to seed turbo's affected graph; nothing in `packages/react/src` imports it. The real coupling is `core scripts → packages/tailwind/*.css → @nexus/tailwind → @nexus/react`. | Either leave a one-line note in `packages/react/package.json` explaining the dep is a turbo-graph hint, or express the coupling via `turbo.json` `globalDependencies` / per-task `inputs` so the declaration matches actual consumption. |

### Architectural Concerns

**Branch-protection coupling is fragile** (`.github/workflows/ci.yml:267-270`) — The inline note relies on the future renamer noticing the comment immediately above the job. Treat `test-coverage` as a stable CI public-API name (rename-freeze) or move the contract to a file the protection-rule owner already monitors (e.g. CODEOWNERS-style).

**`test-unit` partially overlaps `test-coverage`** (`.github/workflows/ci.yml:203`) — `test:coverage` runs both vitest projects (unit + storybook), so `test-unit` re-executes the unit project. Defensible as a fast-fail signal before the Playwright run, but the overlap should be a deliberate, documented decision in the YAML, not implicit.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -168,8 +144,7 @@ jobs:
       - name: Build affected packages
         run: |
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            # Force @nexus/react: bundle-size needs its dist even when only core/scripts changes (core isn't a workspace dep of react).
-            yarn turbo build --filter='...[origin/${{ github.base_ref }}]' --filter='@nexus/react'
+            yarn turbo build --filter='...[origin/${{ github.base_ref }}]'
```
examlly-tech: **Minor — shell injection class:** `${{ github.base_ref }}` interpolates directly into the `run:` block. Branch names allow quotes/metacharacters, so a crafted PR base could inject shell. Move to `env: BASE_REF: ${{ github.base_ref }}` and use `"$BASE_REF"`. Same for the typecheck step.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:147,194` | `${{ github.base_ref }}` is interpolated directly into the `run:` shell block in both `build` and `typecheck`. Branch names can contain shell metacharacters (`'`, `"`, `;`, `$(...)`), so this is the classic GitHub-Actions script-injection class. | Pass via `env: BASE_REF: ${{ github.base_ref }}` on the step, then use `"$BASE_REF"` inside the script. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:170,207,234,268,322,355` | `&& needs.build.result == 'success'` duplicates the implicit `success()` semantics of `needs:`. GitHub Actions already short-circuits a dependent job when any `needs:` entry fails. | Drop the `&& needs.build.result == 'success'` clause; keep only the `changes.outputs.*` check. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:156-158` | `.turbo` is packed into `build-artifacts` and downloaded by every downstream React job, but none of them run turbo (vitest, size-limit, storybook directly). | Drop `.turbo` from the artifact `path:`. Turbo's own cache is restored separately by the `build` job's cache step and isn't needed downstream. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -14,6 +14,18 @@ permissions:
...
 
+# ============================================
+# Required status checks (branch protection on `main`)
+# ============================================
+# The following job names are listed as required checks in branch protection
+# rules. Renaming or removing any of them requires a coordinated protection-rule
+# update in the same change — otherwise every subsequent PR blocks on a check
+# that no longer exists.
+#
+#   - test-coverage   (renamed from test-storybook when consolidated)
```
examlly-tech: **Minor — required-checks list incomplete:** Only `test-coverage` is named. If `lint`/`format-check`/`build`/`typecheck`/`audit-tokens` are also required (they must be, or PRs touching only non-react packages would merge unchecked), list them so any rename gets the same coordination notice.

.github/workflows/ci.yml
```diff
@@ -225,11 +222,15 @@ jobs:
   # React Package Jobs
   # ============================================
 
+  # Runs the `unit` vitest project under jsdom — a deliberate fast-fail signal
+  # before the slower `test-coverage` job (which runs both projects, including
+  # `unit`, under Playwright with coverage instrumentation). The unit project
```
examlly-tech: **Minor — comment inaccurate:** `test-coverage` does not run the `unit` project under Playwright. `vitest.config.ts:29` pins `unit` to `environment: 'jsdom'`; only the `storybook` project uses Playwright. The duplication is fast-fail without coverage overhead, not jsdom-vs-Playwright.

packages/react/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@nexus/react",
   "version": "0.0.1",
   "description": "React components built with Radix UI and Tailwind CSS",
+  "//": "devDependencies includes @nexus/core as a turbo-affected-graph hint, not a code import. Turbo's --filter='...[origin/base]' resolves affected packages via package.json deps; without this edge, a core/scripts change that doesn't regenerate packages/tailwind/*.css skips @nexus/react#build and bundle-size loses its dist input.",
```
examlly-tech: **Architectural — wrong relationship modeled:** Annotation documents the symptom but `@nexus/tailwind` is the generated artifact of `@nexus/core`. The structurally correct fix is `@nexus/tailwind dependsOn @nexus/core#build:tailwind` in turbo, not threading a fake devDep through a published manifest. Worth a follow-up issue.

**Readability:** If keeping the annotation, switch to array form (`"//": ["line 1", "line 2"]`) — the current 320-char single line is hard to scan.

.github/workflows/ci.yml
```diff
@@ -269,68 +270,33 @@ jobs:
...
-        if: steps.playwright-cache.outputs.cache-hit == 'true'
-        run: npx playwright install-deps chromium
-
-      - name: Run Storybook tests
-        run: yarn test:storybook
-
+  # Required status check — see "Required status checks" block at top of file
+  # before renaming or removing.
   test-coverage:
     name: Test Coverage
```
examlly-tech, line 310: **Architectural — coverage gate is performative:** PR body notes `yarn test:coverage` reports 0% because browser-mode vitest doesn't instrument. The 70% thresholds in `vitest.config.ts:83` therefore never fire, yet this job is named `Test Coverage` and listed as a required check. Either fix instrumentation, rename the job, or drop the thresholds.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `turbo.json:8` | `build:tailwind` task declares cross-package `outputs: ["../tailwind/*.css"]` but is never invoked via `turbo run`. The root `tokens:tailwind` script bypasses turbo with `cd && cp`. Dead graph edge. | Either route `tokens:tailwind` through turbo (gains caching + a real graph edge) or delete this task definition. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:34-58` | The `changes` filter has no entry for `apps/**`. A breaking change under `apps/playground/` triggers zero CI jobs — silent skip. | Add an `apps` filter (or expand `packages` to include `apps/**`), or explicitly document the intentional exclusion. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:18-27` | The required-checks header lists only `test-coverage`. If `lint`/`format-check`/`build`/`typecheck`/`audit-tokens` are also required (they must be), the rename-coordination notice doesn't cover them. | List every required job by name so future renames trigger the same coordination instinct. |
| 4 | ⚠️ Minor | `.github/workflows/ci.yml:226-228` | Comment above `test-unit:` claims `test-coverage` runs the `unit` project "under Playwright with coverage instrumentation". `vitest.config.ts:29` pins the `unit` project to `environment: 'jsdom'`; only the `storybook` project uses Playwright. | Reword: duplication is fast-fail (no coverage overhead) vs full coverage run, not a jsdom-vs-Playwright distinction. |

### Architectural Concerns

**Workspace graph models the `core → tailwind` relationship in the wrong place** (`packages/react/package.json:5`, `turbo.json:8`) — The `"//"` annotation documents the symptom (`@nexus/core` listed as `@nexus/react` devDep purely to influence turbo's affected resolver) but the actual relationship is `@nexus/tailwind` being a generated artifact of `@nexus/core`. The structurally correct fix is `@nexus/tailwind dependsOn @nexus/core#build:tailwind` (or moving generated CSS into `core/dist`); the current shape threads a fake devDep edge into a published package's manifest.

**`test-coverage` gate is performative** (`.github/workflows/ci.yml:296`, `vitest.config.ts:83`) — PR body flags that browser-mode vitest doesn't instrument, so `yarn test:coverage` reports 0% yet the 70% thresholds in `vitest.config.ts` don't fail the job. The job is named `Test Coverage` and listed as a required check, but it enforces nothing coverage-related. Either fix instrumentation, rename the job, or drop the thresholds so the contract matches what's enforced — at minimum open a tracked issue.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -269,68 +270,33 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
-      - name: Regenerate @nexus/tailwind and check freshness
+      - name: Regenerate @nexus/tailwind
+        id: regen
+        run: yarn tokens:tailwind
+
+      - name: Check token CSS freshness
+        if: ${{ !cancelled() && steps.regen.conclusion == 'success' }}
```
examlly-tech, line 282: **Minor — redundant guard:** `!cancelled()` adds nothing when paired with `steps.regen.conclusion == 'success'` — a cancelled job's regen conclusion is never `'success'`. Drop `!cancelled() &&` from all three audit `if:` clauses (lines 278, 286, 290) for a cleaner read.

.github/workflows/ci.yml
```diff
@@ -48,6 +60,15 @@ jobs:
   # ============================================
   # Always Run (on code changes)
   # ============================================
+  #
+  # format-check and lint run full-tree (no `--filter='...[origin/base]'`)
```
examlly-tech: **Minor — orphaned comment:** This block describes the lint/format-vs-build asymmetry but sits between unrelated jobs with no anchor. A drive-by reader can't tell which jobs it covers. Move directly above `format-check:` or prefix lines with `# format-check / lint —`.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:278,286,290` | `!cancelled()` clause is redundant with `steps.regen.conclusion == 'success'` — a cancelled job's regen conclusion is never `'success'`, so the explicit cancellation guard adds nothing. | Drop `!cancelled() &&`; keep only `steps.regen.conclusion == 'success'`. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:148-154,203-209` | Both `build` and `typecheck` define a turbo cache step with `key: turbo-${{ runner.os }}-${{ github.sha }}`. Two saves under the same key in the same run causes one to warn with "cache already exists". | Either differentiate per job (`turbo-build-…`/`turbo-typecheck-…`) or make typecheck restore-only via `actions/cache/restore@v4`. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:60-71` | The lint/format-vs-build asymmetry comment block sits between the `changes` job and `format-check` with no anchor, reading as if it belongs to whichever job follows. | Move directly above `format-check:` (the job it actually describes), or prefix lines with `# format-check / lint —`. |
| 4 | ⚠️ Minor | `packages/react/package.json:5` | The `"//"` field is a 320-char single-line string mixing rationale and mechanics — hard to read at the top of the file. | Convert to array form: `"//": ["@nexus/core is a turbo-affected-graph hint, not a code import", "Without it, core/scripts-only changes skip @nexus/react#build and bundle-size loses its dist input"]`. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -42,12 +64,22 @@ jobs:
               - 'packages/test-utils/**'
             packages:
               - 'packages/**'
+              - 'apps/**'
```
examlly-tech: **Architectural — document the asymmetry:** `apps/**` is correctly in `packages` but the `react` filter above excludes it. Today no app imports `@nexus/react`, so this is correct. Add a one-line comment here noting the assumption so the next editor sees it the moment an app starts consuming `@nexus/react` at runtime.

examlly-tech (commented):
## Principal Architect Review — Verdict: APPROVED

### Architectural Concerns

**Change-detection asymmetry between `packages` and `react` filters** (`.github/workflows/ci.yml:60-69`) — `apps/**` is now in the `packages` filter (correct for build/typecheck), but the `react` filter still lists only `packages/{react,tailwind,core,test-utils}/**`. Today this is correct (playground imports none of `@nexus/*`), but it's a silent skip the moment an app starts consuming `@nexus/react` — a regression test-coverage / bundle-size would otherwise catch wouldn't trigger. A one-line comment in the filter block documenting the assumption prevents the next editor from missing it.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:127` | ESLint cache primary key includes `${{ github.sha }}`. SHA changes per commit, so the primary key never hits — every run uploads a brand-new cache that no future run keys on directly (only `restore-keys` resolve). | Drop `-${{ github.sha }}` from the primary key. The `hashFiles('yarn.lock','eslint.config.js')` prefix is sufficient; SHA-scoping only earns its place for cross-job sharing in the same run (like the build → typecheck turbo cache), which the lint job doesn't do. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:336-345` | The Playwright system-deps step runs `npx playwright install-deps chromium` on **cache hit** (apt-only) while the cache-miss branch runs `playwright install --with-deps` (browsers + apt). System deps aren't cached either way — the cache-hit branch reinstalls them every run anyway. | Either always run `install-deps` once unconditionally, or rely on `ubuntu-latest` shipping the deps and drop the install-deps step entirely on cache hit. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:30,32` | Required-checks header annotates only `typecheck → "TypeScript Check"` and `test-unit → "Unit Tests"` with parenthetical job names. `audit-tokens`, `test-coverage`, `build-storybook`, `bundle-size` are not annotated, suggesting their `id` matches their job `name:` (which is true for `bundle-size` but not for the others). | Annotate every entry or none. Branch protection matches on job `name:` not id, so partial annotation is misleading. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -107,31 +120,17 @@ jobs:
...
         run: yarn install --frozen-lockfile
 
-      - name: Restore Turbo cache
+      - name: Restore ESLint cache
         uses: actions/cache@v4
         with:
-          path: .turbo
-          key: turbo-${{ runner.os }}-${{ github.sha }}
+          path: .eslintcache
+          key: eslint-${{ runner.os }}-${{ hashFiles('yarn.lock', 'eslint.config.js') }}-${{ github.sha }}
```
examlly-tech, line 121: **Minor — primary key never hits:** `${{ github.sha }}` in the primary key guarantees a miss every commit. The cache uploads each run but is only ever restored via `restore-keys`. Drop `-${{ github.sha }}`; the `hashFiles('yarn.lock','eslint.config.js')` prefix alone is right for a per-job cache that nothing else consumes.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,28 @@ permissions:
...
+# rules on `main`. Renaming or removing any of them requires a coordinated
+# protection-rule update in the same change — otherwise every subsequent PR
+# blocks on a check that no longer exists.
+#
+#   - format-check
+#   - lint
+#   - build
+#   - typecheck      (job name: "TypeScript Check")
+#   - test-unit      (job name: "Unit Tests")
+#   - audit-tokens   (job name: "Audit Tokens")
```
examlly-tech: **Minor — inconsistent annotation:** Two jobs get parenthetical name hints (`typecheck → "TypeScript Check"`, `test-unit → "Unit Tests"`); the other four don't, which implies their `id` matches their `name:` (false for `audit-tokens`/`test-coverage`/`build-storybook`). Annotate every entry or none — branch protection matches on `name:`.

.github/workflows/ci.yml
```diff
@@ -269,25 +287,33 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
-      - name: Regenerate @nexus/tailwind and check freshness
+      - name: Regenerate @nexus/tailwind
+        id: regen
+        run: yarn tokens:tailwind
```
examlly-tech: **Minor — modular pipeline freshness gap:** This job regen+diffs `packages/tailwind/` correctly, but the parallel `yarn tokens:modular` pipeline writes tracked files under `apps/playground/public/themes/**` and `apps/playground/src/styles/**` with no equivalent check. Same vacuous-check class of bug this PR fixes. Add a sibling step.

.github/workflows/ci.yml
```diff
@@ -35,19 +58,33 @@ jobs:
...
+            # the react test/build/bundle-size jobs. Revisit if an app starts
+            # importing from @nexus/react.
             react:
               - 'packages/react/**'
               - 'packages/tailwind/**'
               - 'packages/core/**'
               - 'packages/test-utils/**'
             packages:
               - 'packages/**'
+              - 'apps/**'
```
examlly-tech: **Minor — stub gating required Build check:** Adding `apps/**` to this filter (round 4) routes `apps/docs/**` into the required Build job, where `@nexus/docs` ships a stub `build: "echo Docs site not yet implemented"` that exits 0 vacuously. Either exclude `apps/docs/**` from the filter or drop the script so turbo skips it.

.github/workflows/ci.yml
```diff
@@ -225,11 +238,16 @@ jobs:
...
 
+  # Runs the `unit` vitest project under jsdom — a deliberate fast-fail signal
+  # before the slower `test-coverage` job (which runs both vitest projects:
+  # `unit` in jsdom and `storybook` under Playwright, plus coverage). The unit
+  # project therefore runs twice in CI; the duplication is the point — this
+  # job fails fast without the Playwright + coverage overhead.
   test-unit:
     name: Unit Tests
     runs-on: ubuntu-latest
     needs: [changes, build]
```
examlly-tech: **Minor — false build coupling:** `test-unit`/`test-coverage`/`build-storybook` `needs: build` + download `build-artifacts`, but vitest reads `src/` and storybook reads `src/**/*.stories.*`. Only `bundle-size` actually needs `dist/`. A failed build now blocks tests that don't structurally depend on it. Verify cross-workspace imports first.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:287-319` | `audit-tokens` now correctly regen+diffs the bundled `packages/tailwind/` output, but the parallel modular-token pipeline (`yarn tokens:modular` → `scripts/sync-playground-themes.js`) writes tracked files under `apps/playground/public/themes/**` and `apps/playground/src/styles/**` with no equivalent freshness check. Same class of vacuous-check bug this PR was opened to fix. | Add a sibling step running `yarn tokens:modular` then `git diff --exit-code -- apps/playground/public/themes apps/playground/src/styles`. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:69-72` + `apps/docs/package.json:6` | Adding `apps/**` to the `packages` filter (round 4) routes `apps/docs/**` changes into the Build required check, where `@nexus/docs` ships a stub `build: "echo Docs site not yet implemented"` that exits 0 — the required gate passes vacuously on docs edits. | Either exclude `apps/docs/**` from the `packages` filter until docs has a real build, or remove the `build` script from `@nexus/docs` so turbo skips it instead of echo-passing. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:241,275,317,360` | `test-unit`, `test-coverage`, `build-storybook` all `needs: build` + download `build-artifacts`, but they execute from source (vitest reads `src/`, storybook reads `src/**/*.stories.*`). Only `bundle-size` actually consumes `packages/react/dist/`. A failed `build` blocks tests that don't structurally depend on it. | Drop `needs: build` and the `download-artifact` steps from `test-unit`, `test-coverage`, `build-storybook`; keep them only on `bundle-size`. Verify locally first — cross-workspace `@nexus/*` imports may resolve to dist. |

### Architectural Concerns

**Modular-pipeline freshness gap mirrors the very bug this PR fixes** (`apps/playground/public/themes/**`, `apps/playground/src/styles/**`) — `audit-tokens` validates the bundled `packages/tailwind/` output correctly, but the parallel `yarn tokens:modular` pipeline has no equivalent CI freshness check. Token-primitive edits land silently broken in the playground. Same shape as the issue this PR was opened to solve.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -269,25 +287,33 @@ jobs:
...
-      - name: Regenerate @nexus/tailwind and check freshness
+      - name: Regenerate @nexus/tailwind
+        id: regen
+        run: yarn tokens:tailwind
+
+      - name: Check token CSS freshness
+        if: ${{ !cancelled() && steps.regen.conclusion == 'success' }}
         run: |
-          yarn workspace @nexus/core build:tailwind
           if ! git diff --exit-code -- packages/tailwind; then
```
examlly-tech: **Minor — drift blind to new files:** `git diff --exit-code` only inspects tracked files. If a future token category produces a new CSS filename in `packages/tailwind/`, the check passes vacuously despite the uncommitted output. Switch to `git status --porcelain -- packages/tailwind` (catches untracked) or assert exact file-set equality.

.github/workflows/ci.yml
```diff
@@ -315,56 +341,13 @@ jobs:
           restore-keys: |
             playwright-${{ runner.os }}-
```
examlly-tech: **Minor — cache bloat on Playwright bumps:** When `yarn.lock` changes the Playwright version, the prefix restore brings stale browsers; `install chromium` layers new ones, and both get saved under the new key. Either drop the prefix `restore-keys` line or clear `~/.cache/ms-playwright` on prefix-only hit.

examlly-tech (commented):
## SDE2 Review — Verdict: APPROVED

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:294-300` | `git diff --exit-code -- packages/tailwind` only inspects tracked files. If a future token category emits a brand-new `.css` filename, the diff passes vacuously despite uncommitted output. | Use `git status --porcelain -- packages/tailwind` instead (catches untracked), or pair the diff with an ls-files cardinality check. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:335-350` | Playwright cache uses `restore-keys: playwright-${{ runner.os }}-` (no version). When `yarn.lock` bumps Playwright, the prefix restore brings in stale browsers, then `install chromium` layers the new version on top — both get saved under the new key, growing the cache over time. | Drop the prefix `restore-keys` line (saved cache will re-create cleanly on lock change), or `rm -rf ~/.cache/ms-playwright` before the new install when only the prefix matched. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

examlly-tech (commented):

### Issues

| 1 | ❌ Blocking | `.github/workflows/ci.yml:45-78` | Required-check + conditional-job trap. A PR touching only repo-root `.md`, `apps/docs/**`, `reports/**`, or `.claude/**` matches no filter (`react`/`packages`/`ci`). Every required check evaluates `if:` to false and is skipped, but GitHub branch protection treats skipped required checks as not-met — the PR can't merge. | Add an always-running aggregator job (`needs:` every other job, `if: always()`) listed as the single required check in branch protection; the aggregator passes when no `needs.*.result == 'failure'`. Move branch protection from per-job names to this aggregator. |
| 2 | ❌ Blocking | `.github/workflows/ci.yml:271-316` | `audit-tokens` is gated on `react` only, but its freshness check now covers `apps/playground/public/themes/` and `apps/playground/src/styles/`. A playground-only PR matches `packages` (not `react`), so `audit-tokens` doesn't run — the only thing protecting those paths is skipped. | Add `apps/playground/**` to the `react` filter, or extract a `tokens` filter that mirrors the freshness `DIRS` list and gate `audit-tokens` on that. |
| 3 | ❌ Blocking | `.github/workflows/ci.yml:271-316` | `audit:class-refs` scans `apps/**` sources (it walks the whole tree for semantic-token usage), but the audit only runs when the `react` filter fires. An `apps/**` change introducing an invalid semantic class (e.g., `nx:bg-accent` instead of `nx:bg-background-hover`) bypasses validation. | Either decouple `audit:class-refs` into its own job gated on `packages`, or include `apps/**` in the `react` filter. |

### Architectural Concerns

**`test-coverage` is now the only Playwright signal in CI** (`.github/workflows/ci.yml:320-361`) — with `test-storybook` removed and `test-unit` jsdom-only, story-test regressions surface only via the slowest required job. The consolidation has no fast-fail path for browser-mode failures until #67 lands. Worth folding into #67's scope or opening a separate follow-up.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -35,19 +58,37 @@ jobs:
...
               - 'packages/tailwind/**'
               - 'packages/core/**'
               - 'packages/test-utils/**'
             packages:
               - 'packages/**'
+              - 'apps/**'
+              # apps/docs is a stub placeholder with no build/dev scripts;
+              # changes there shouldn't trigger Build / TypeScript Check.
+              # Drop the exclusion when @nexus/docs becomes a real package.
+              - '!apps/docs/**'
```
examlly-tech: **Blocking — required-check trap:** With `!apps/docs/**` excluded and no filter for repo-root `.md` / `.claude/**` / `reports/**`, a docs-only PR matches none of `packages`/`react`/`ci`. Every required check is skipped, and GitHub treats skipped required checks as not-met — branch protection will block the merge. Add an always-running aggregator job as the single required check.

.github/workflows/ci.yml
```diff
@@ -243,19 +265,14 @@ jobs:
...
       - name: Run unit tests
         run: yarn test:unit
 
   audit-tokens:
     name: Audit Tokens
     runs-on: ubuntu-latest
-    needs: [changes, build]
-    if: needs.changes.outputs.react == 'true' && needs.build.result == 'success'
+    needs: changes
+    if: needs.changes.outputs.react == 'true' || needs.changes.outputs.ci == 'true'
```
examlly-tech: **Blocking — gate/scope mismatch:** `audit-tokens` is gated on `react`, but its freshness check now covers `apps/playground/public/themes/` and `apps/playground/src/styles/`. A playground-only PR matches `packages` not `react`, so the freshness check never runs against the paths it's supposed to protect. Add `apps/playground/**` to the `react` filter or split a `tokens` filter.

.github/workflows/ci.yml
```diff
@@ -269,25 +286,42 @@ jobs:
...
             exit 1
           fi
 
       - name: APCA contrast audit
+        if: ${{ !cancelled() && steps.regen.conclusion == 'success' }}
         run: yarn workspace @nexus/core audit:contrast
 
       - name: Tailwind class-ref audit
+        if: ${{ !cancelled() && steps.regen.conclusion == 'success' }}
         run: yarn workspace @nexus/core audit:class-refs
```
examlly-tech, line 303: **Blocking — audit reach gap:** `audit:class-refs` scans `apps/**` sources, but the job only fires when `react` is true. An `apps/**` change introducing an invalid semantic class (e.g. `nx:bg-accent`) bypasses validation. Run class-refs whenever `packages` fires, or split it into its own job.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,29 @@ permissions:
...
+#   - TypeScript Check
+#   - Unit Tests
+#   - Audit Tokens
+#   - Test Coverage     (renamed from "Storybook Tests" when consolidated;
+#                        see also #67 for the coverage-instrumentation gap)
+#   - Build Storybook
+#   - Bundle Size
+#
+# If you add or rename a required check, update both this list AND the
+# protection rule.
```
examlly-tech: **Minor — keep-in-sync comment:** This block duplicates branch-protection rules and asks readers to manually sync two systems. `code-comments.md` bans this pattern ("if sync is required, enforce it at compile time"). Branch protection is the source of truth — trim to a one-line pointer or drop the block.

.github/workflows/ci.yml
```diff
@@ -35,19 +58,37 @@ jobs:
...
   # ============================================
+  #
+  # These two jobs run full-tree (no `--filter='...[origin/base]'`) while
+  # build, typecheck, and test run affected-only. The asymmetry is intentional:
+  # local `yarn lint` / `yarn format:check` run full-tree, so CI mirrors that
+  # exactly. A divergence here would let a stale rule miss a file locally that
+  # CI then flags (or vice versa). Affected-only lint is also error-prone — a
+  # config change in eslint.config.js must invalidate the whole tree, which
+  # `--filter` won't catch. Keep these jobs full-tree until wall-clock pain
+  # forces the move.
```
examlly-tech: **Minor — rationale block:** 11 lines explaining "why full-tree, not affected-only" is tradeoff narration that `code-comments.md` says belongs in the PR description, not the file. A one-liner ("runs full-tree to mirror local yarn lint / format:check") is enough.

.github/workflows/ci.yml
```diff
@@ -225,11 +242,16 @@ jobs:
   # React Package Jobs
   # ============================================
 
+  # Runs the `unit` vitest project under jsdom — a deliberate fast-fail signal
+  # before the slower `test-coverage` job (which runs both vitest projects:
+  # `unit` in jsdom and `storybook` under Playwright, plus coverage). The unit
+  # project therefore runs twice in CI; the duplication is the point — this
+  # job fails fast without the Playwright + coverage overhead.
```
examlly-tech: **Minor — rationale block:** The "duplication is the point" explanation is design narration. Cut to: `# Fast-fail unit subset; coverage runs separately in test-coverage.`

.github/workflows/ci.yml
```diff
@@ -269,25 +286,42 @@ jobs:
...
+        if: ${{ !cancelled() && steps.regen.conclusion == 'success' }}
         run: |
-          yarn workspace @nexus/core build:tailwind
-          if ! git diff --exit-code -- packages/tailwind; then
-            echo "::error::packages/tailwind/ is stale — run 'yarn workspace @nexus/core build:tailwind' and commit the result."
+          DIRS="packages/tailwind apps/playground/public/themes apps/playground/src/styles"
+          # Use git status --porcelain instead of git diff so new untracked
+          # files (e.g. a new CSS filename from a future token category) also
+          # trigger the failure.
+          DIRTY=$(git status --porcelain -- $DIRS)
```
examlly-tech: **Minor — shell robustness:** Unquoted `$DIRS` word-splits correctly only because the paths happen to have no spaces, and a `git status` failure would leave `DIRTY` empty and silently pass. Add `set -euo pipefail` at the top of the `run:` block (or pass paths as an array).

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `.github/workflows/ci.yml:17-38` | The 22-line required-checks header is a textbook keep-in-sync comment between this file and branch protection — `.claude/rules/code-comments.md` explicitly bans this pattern ("if sync is required, enforce it at compile time"). | Branch protection is the source of truth. Trim this to a one-line pointer (`# See branch protection settings on main for the required-check list.`) or drop the block entirely. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:80-91` | 11-line tradeoff rationale block describing why lint runs full-tree vs. affected-only. `.claude/rules/code-comments.md` says rationale belongs in PR description, not in the file. | Replace with a one-liner: `# Full-tree to mirror local 'yarn lint' / 'yarn format:check'.` Move the rest to the PR body. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:245-249` | 5-line "duplication is the point" rationale block above `test-unit:` — same anti-pattern. | One-liner: `# Fast-fail unit subset; coverage runs separately in test-coverage.` |
| 4 | ⚠️ Minor | `.github/workflows/ci.yml:296-307` | `git status --porcelain -- $DIRS` relies on word-splitting of an unquoted variable — works today only because the paths have no spaces. A `git status` failure would leave `DIRTY` empty and the step would pass silently. | Add `set -euo pipefail` at the top of the `run:` block (or use an array of paths) to fail loud on either issue. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -434,3 +383,36 @@ jobs:
...
+  # ============================================
+  # Runs unconditionally so docs-only / .claude/** / repo-root PRs still
+  # report a status that satisfies branch protection. Skipped upstream jobs
+  # (filtered out by `if:`) are treated as success; only `failure` and
+  # `cancelled` propagate.
+  ci-status:
+    name: CI Status
+    runs-on: ubuntu-latest
+    if: always()
+    needs:
```
examlly-tech, line 422: **Blocking — aggregator misses `changes`:** `needs:` lists every downstream job but not the `changes` job they all depend on. If `changes` fails (transient paths-filter error, checkout failure), every downstream `if:` evaluates against empty outputs → all skip → aggregator sees only `"skipped"` results, passes, and the PR merges with zero validation. Add `- changes` to `needs:`.

.github/workflows/ci.yml
```diff
@@ -434,3 +383,36 @@ jobs:
...
         run: yarn size-limit
+
+  # ============================================
+  # Aggregator — the single required status check
+  # ============================================
+  # Runs unconditionally so docs-only / .claude/** / repo-root PRs still
+  # report a status that satisfies branch protection. Skipped upstream jobs
+  # (filtered out by `if:`) are treated as success; only `failure` and
+  # `cancelled` propagate.
+  ci-status:
```
examlly-tech, line 418: **Minor — contract anchor:** Add a one-line comment here ("every required job must appear in `needs:` below") so the keep-in-sync rule between this list and any newly-added required job lives next to the list, not in the file header 380 lines up.

examlly-tech (changes requested):

### Issues

| 1 | ❌ Blocking | `.github/workflows/ci.yml:393-407` | `ci-status` aggregator omits `changes` from `needs:`. If the Detect Changes job fails (transient paths-filter error, checkout failure, network blip), every downstream job's `if:` evaluates to false (because the outputs are empty) → all downstream jobs skip → aggregator sees only `"skipped"` results → aggregator passes → PR merges with zero actual validation. | Add `- changes` to the `needs:` list. A failed Detect Changes will then propagate as `failure` through the aggregator. |

### Architectural Concerns

**Aggregator membership has no compile-time guard** (`.github/workflows/ci.yml:393-407`) — adding a future required job to ci.yml requires also appending it to `ci-status.needs:`; nothing fails fast if you forget. Acceptable trade-off vs. the per-job-required-check trap this replaces, but add a one-line comment on `ci-status` itself stating "every required job must appear in `needs:` below" so the contract lives next to the list, not in the file header 380 lines away.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

.github/workflows/ci.yml
```diff
@@ -35,19 +40,27 @@ jobs:
...
               - 'packages/tailwind/**'
               - 'packages/core/**'
               - 'packages/test-utils/**'
             packages:
               - 'packages/**'
+              - 'apps/**'
+              # apps/docs is a stub placeholder with no build/dev scripts;
+              # changes there shouldn't trigger Build / TypeScript Check.
+              # Drop the exclusion when @nexus/docs becomes a real package.
+              - '!apps/docs/**'
```
examlly-tech: **Blocking — negation is a no-op:** `dorny/paths-filter` defaults to `predicate-quantifier: 'some'`, so any file matching `apps/**` sets `packages: true` regardless of `!apps/docs/**`. The docs-only-skip behavior the comment claims doesn't actually hold — Build/Typecheck/Audit-Tokens still run on docs-only PRs. Split into separate filter outputs and OR them at consumers.

.github/workflows/ci.yml
```diff
@@ -434,3 +383,36 @@ jobs:
...
+      - test-coverage
+      - build-storybook
+      - bundle-size
+    steps:
+      - name: Aggregate required-check results
+        run: |
+          set -euo pipefail
+          results='${{ toJSON(needs) }}'
+          echo "$results"
+          if echo "$results" | grep -qE '"result"[[:space:]]*:[[:space:]]*"(failure|cancelled)"'; then
```
examlly-tech: **Minor — reinvents native idiom:** The canonical pattern is `if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}` on a one-line `exit 1` step. Drops the bash, the JSON-via-regex (false-positive prone on future `needs.*.outputs`), and the `set -euo pipefail`.

.github/workflows/ci.yml
```diff
@@ -434,3 +383,36 @@ jobs:
...
+      - typecheck
+      - test-unit
+      - audit-tokens
+      - test-coverage
+      - build-storybook
+      - bundle-size
+    steps:
+      - name: Aggregate required-check results
+        run: |
+          set -euo pipefail
```
examlly-tech: **Minor — `set -euo pipefail` is dead weight:** one operative command (the `if`-guarded grep), no piped commands, no variable expansion, and `-e` is suppressed inside an `if` test anyway. Drop the line, or collapse to `set -e` if you keep the grep.

examlly-tech (changes requested):
## SDE2 Review — Verdict: CHANGES REQUIRED

### Issues

| 1 | ❌ Blocking | `.github/workflows/ci.yml:51-58` | The `!apps/docs/**` negation in the `packages` filter is silently a no-op. `dorny/paths-filter` defaults to `predicate-quantifier: 'some'`, meaning the file matches the filter if it matches **any** pattern — so a file under `apps/docs/` still matches `apps/**` and trips `packages: true`. The docs-only-skip behavior the PR body and inline comment claim does not actually happen; Build, TypeScript Check, and Audit Tokens run on docs-only edits. | Split into two filter outputs (`packages-impl: 'packages/**'` and `apps-non-docs: 'apps/**' + '!apps/docs/**'` under `predicate-quantifier: 'every'`) and OR them at consumer `if:` lines. Verify by opening a docs-only test PR and confirming only `ci-status` reports. |
| 2 | ⚠️ Minor | `.github/workflows/ci.yml:407-417` | Aggregator regex-parses `toJSON(needs)` to detect failure/cancelled. Fragile (a future `needs.*.outputs.result` field would false-positive the grep) and reinvents the documented `contains(needs.*.result, …)` idiom. | Replace the whole `run:` with a step gated by `if: ${{ contains(needs.*.result, 'failure') \|\| contains(needs.*.result, 'cancelled') }}` whose body is just `exit 1`. Drops the bash, the regex, and the `set -euo pipefail`. |
| 3 | ⚠️ Minor | `.github/workflows/ci.yml:411` | `set -euo pipefail` in the aggregator's `run:` block has no operative effect — there's one command (the grep `if`), no piped commands needing `pipefail`, no variable expansion needing `-u`, and `-e` is suppressed inside an `if` test. | Drop the line, or collapse to `set -e` if the grep approach is kept after #2 is reconsidered. |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

examlly-tech (approved):
## Principal Architect Review — Verdict: APPROVED

The round-8 changes resolve the two open architectural items: the split paths-filter call with `predicate-quantifier: every` makes the `!apps/docs/**` negation operative, and adding `changes` to `ci-status.needs:` closes the silent-skip-everything gap on a Detect Changes failure. The required-check contract is now documented inline above `ci-status`, and every job ID in the file appears in its `needs:` list. The `react` filter excluding `apps/**` is justified by current consumption (no app imports `@nexus/react`) and tracked for revisit; the `packages` gate on `audit-tokens` correctly covers `audit:class-refs`'s `apps/**` scan surface. Outstanding architectural items are tracked in #66 (workspace graph), #67 (coverage threshold), and #68 (fast Storybook signal).

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

examlly-tech (approved):
## SDE2 Review — Verdict: APPROVED

Round-8 changes cleanly resolve the two blocking items. The split paths-filter call is the correct fix for the `!apps/docs/**` no-op under `predicate-quantifier: 'some'`. The aggregator rewrite drops the regex, `toJSON(needs)` parse, and inert `set -euo pipefail` together — step-level `if: contains(needs.*.result, 'failure') \|\| contains(needs.*.result, 'cancelled')` + bare `exit 1` is the idiomatic shape. Shell quoting, cache key structure, freshness check, and guard semantics all check out.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/64
- **GitHub Issue:** #42

Issue #63, opened 2026-05-21
Figma↔code parity: audit shadow primitives


aishvaryagupta:
## Summary

Sub-issue of #56 (Figma↔code parity audit, merged). Covers shadow primitive parity — the most structured category: 5 modes × 2 themes (light/dark) × 7-8 named shadow definitions × 5 sub-tokens each.

## Prerequisites

- [x] **#56 has merged** — `yarn audit:figma-parity --category shadow` walks theme-keyed shadow files correctly
- [x] #58 (focus shadow extracted to single source) — landed via PR #60

## Scope

- [ ] `yarn audit:figma-parity --category shadow` — walks all modes × themes × shadows × sub-tokens (offsetX, offsetY, blur, spread, color)
- [ ] Resolve drift per sub-token
- [ ] Verify focus shadow values (already pushed in #56 part 1) are still aligned
- [ ] Pay particular attention to alpha-bearing colors (e.g., `2xs` shadow color) — these bypass OKLCH conversion via the mechanical path and are the most likely to drift silently
- [ ] Re-run audit; reports zero drift across all 5 modes × 2 themes
- [ ] If shadow surfaces a canonical decision rule, document it in `.claude/rules/figma.md`

## Notes for implementing session

The script built in #56 may need a per-category sub-walker for shadow's depth. If `--category shadow` doesn't handle theme-keyed files (`shadow-{mode}-{light|dark}.json`), expand the script here.

## Out of scope

- Other primitive categories

aishvaryagupta:
## ⏸ Deferred (2026-05-21)

Active audit work is wrapping up; this issue is being shelved — **not closed**, just removed from the active work queue.

**Rationale:** Focus color values were just manually pushed to Figma (matching code: \`#737373\` / \`#b0b0b0\` / \`#dc2626\` / \`#fca5a5\`). The change was verified by hand in this session. Without this audit category wired up, the values aren't CI-protected, but they also rarely change after a deliberate retune.

**When to pick this back up:**
- Next time focus or shadow values are tuned (the catch-net would prevent silent drift back to bad contrast)
- Before a 1.0/public release
- If anyone reports focus-ring visibility regressions

No blocker; foundation in #56 is fully usable. Shadow category currently refuses with \`exit=2\` citing this issue.

aishvaryagupta:
Closing as part of retiring the Figma-parity track (milestone #9 · Phase 1). Reopen if the parity work is revived.

Issue #62, opened 2026-05-21
Figma↔code parity: audit typography primitives


aishvaryagupta:
## Summary

Sub-issue of #56 (Figma↔code parity audit, merged). Covers typography primitive parity across all 3 modes (`nova` / `vega` / `maia`). The earlier `lyra` / `mira` typography modes were byte-duplicates of `vega` and were removed.

Typography is more complex than dimensional categories because tokens are **composite** (font-family + font-size + font-weight + line-height + letter-spacing per token).

## Prerequisite

- [x] **#56 has merged** — `yarn audit:figma-parity --category typography` walks composite tokens correctly

## Scope

- [ ] `yarn audit:figma-parity --category typography` — emit per-sub-property drift
- [ ] Resolve drift per sub-property: font-family, font-size, font-weight, line-height, letter-spacing
- [ ] Pay particular attention to letter-spacing — Figma exports ship float-32 junk (`-0.800000011920929px`) which #45/PR #53 fixed in code. Verify Figma side has clean rounded values too; if not, push fixes to Figma.
- [ ] Re-run audit; reports zero drift across all 3 modes
- [ ] If typography surfaces a canonical decision rule, document it in `.claude/rules/figma.md`

## Notes for implementing session

If the script built in #56 doesn't handle composite tokens deeply enough, expand it here. The audit must compare each sub-property independently, not just the composite resolved value.

## Out of scope

- Other primitive categories

aishvaryagupta:
## ⏸ Deferred (2026-05-21)

Active audit work is wrapping up; this issue is being shelved — **not closed**, just removed from the active work queue.

**Rationale:** Typography values rarely change post-initial-tune. The letter-spacing float-32 issue (the most likely class of typography drift) was already fixed in code by #45/PR #53. The audit script refuses the `typography` category with `exit=2` citing this issue.

**When to pick this back up:**
- Next time someone re-imports a Figma typography export (where float-32 rounding can creep in)
- Before a 1.0/public release
- If a typography token gets re-tuned in either code or Figma

No blocker; foundation in #56 is fully usable.

aishvaryagupta:
Closing as part of retiring the Figma-parity track (milestone #9 · Phase 1). Reopen if the parity work is revived.

Issue #61, opened 2026-05-21
Figma↔code parity: audit dimensional primitives (radius, borderwidth)


aishvaryagupta:
## Summary

Sub-issue of #56 (Figma↔code parity audit, merged). Covers two dimensional primitive categories that share a common structure (per-mode JSON files mapping a named key to a single dimension value):

- **radius** (blunt / sharp / subtle / smooth / mellow)
- **borderwidth** (vega / lyra / maia / mira / nova)

The `size` primitive layer was deleted in #121 — there is no `--nx-size-*` namespace anymore; spacing now lives as direct-px per-mode semantic files and is audited under `--category spacing` (#128). This issue is scoped to the two remaining dimensional primitive axes only.

## Prerequisite

- [x] **#56 has merged** — `yarn audit:figma-parity --category <X>` works for at least the `color` category (proven there as PoC)

## Scope

- [ ] `yarn audit:figma-parity --category radius` — resolve all reported drift, sync canonical values
- [ ] `yarn audit:figma-parity --category borderwidth` — resolve all reported drift, sync canonical values
- [ ] Re-run each audit; both report zero drift
- [ ] If any category surfaces a new canonical decision rule (e.g., "designers own the radius scale, code follows"), document it in `.claude/rules/figma.md`

## Out of scope

- Color (audited under #56)
- Spacing (audited under #128)
- Typography (separate sub-issue, #62)
- Shadow (separate sub-issue, #63)

aishvaryagupta:
## ⏸ Deferred (2026-05-21)

Active audit work is wrapping up; this issue is being shelved — **not closed**, just removed from the active work queue.

**Rationale:** Dimensional primitives (size/radius/borderwidth) rarely change after initial tuning. The audit script in main today handles the color category and refuses unknown categories with a clear error citing this issue (`exit=2`). Manual spot-checks are sufficient until/unless real drift surfaces.

**When to pick this back up:**
- A dimensional value drifts between code and Figma and gets noticed manually
- Before a 1.0/public release where full audit coverage is wanted
- Someone proposes adding a new dimensional mode/key and wants automated parity guarantees

No blocker; the foundation in #56 is fully usable without this extension.

aishvaryagupta:
Closing as part of retiring the Figma-parity track (milestone #9 · Phase 1). Reopen if the parity work is revived.

Issue #56, opened 2026-05-21
Figma↔code parity foundation: push focus values + build audit:figma-parity script + audit color


aishvaryagupta:
## Summary

**Foundation issue** for the Figma↔code parity audit work. Originally filed with a broader scope ("audit all primitives"); split for reviewability. This issue now covers only the urgent focus-color drift + the audit tooling + color as a proof-of-concept category. The remaining primitive categories live in sub-issues:

- **This issue (#56)** — push focus values to Figma + build `audit:figma-parity` script + audit color
- **#61** — audit dimensional primitives (size + radius + borderwidth)
- **#62** — audit typography primitives
- **#63** — audit shadow primitives

All three sub-issues are blocked on this issue landing first (they need the script).

## Context

Discovered while reviewing PR #55 — the focus ring colors in Figma and `packages/core/tokens/primitives/` have diverged. Code was intentionally retuned in PR #55 to clear WCAG 2.2 SC 1.4.11 (Lc ≥ 45); the Figma values fail the new `audit:contrast` check at all 20 focus↔background pairs.

Known drift values:

| Token | Figma (drifted) | Code (canonical, post-#55) |
|---|---|---|
| `focus.default.color` (light) | `#e5e5e580` (near-white @ 50% α) | `#737373` |
| `focus.default.color` (dark)  | `#e5e5e533`                       | `#b0b0b0` |
| `focus.error.color` (light)   | `#fecaca80`                       | `#dc2626` |
| `focus.error.color` (dark)    | `#991b1b80`                       | `#f87171` |

Other primitive categories may have drifted too — no current process catches this, hence the audit tooling below.

### Part 1 — Push focus values back to Figma

- [ ] Update `focus/default.color` (light + dark) in Figma to match code: `#737373` / `#b0b0b0`
- [ ] Update `focus/error.color` (light + dark) in Figma to match code: `#dc2626` / `#f87171`
- [ ] Verify the focus tokens resolve correctly in Figma now that #58/PR #60 has extracted focus from shadow primitives (focus is theme-agnostic, single source)

### Part 2 — Build `audit:figma-parity` script

- [ ] Add `packages/core/scripts/audit-figma-parity.js`
- [ ] Reads Figma variables via the Figma REST API (read-only — no write operations through this script; pushing fixes back to Figma is manual or via separate tooling)
- [ ] Diffs against the JSON token sources in `packages/core/tokens/primitives/`
[...]

PR #65, opened 2026-05-21
feat(tokens): add audit:figma-parity script + color snapshot

aishvaryagupta:
## Summary

- New CLI `yarn audit:figma-parity --category <name>` diffs `packages/core/tokens/primitives/` against a checked-in Figma snapshot at `packages/core/tokens/figma-snapshot.json`. Pure JS, no external creds at audit time.
- 11 unit tests cover the pure `diffTokenTrees(codeSubtree, snapshotSubtree)` core: value-mismatch, type-mismatch, missing-in-figma, missing-in-code, hex case-insensitivity, alpha preservation, stable sort.
- Initial snapshot captures all 244 color variables from the `Foundation-Master` Figma file (key `9T3zwjS0oLunWaPC3rcDVJ`) via the figma MCP. Running `--category color` reports **0 drift findings** — color primitives are fully in sync today.
- `.claude/rules/figma.md` gains a "Code-vs-Figma Parity Audit" section documenting the canonical rule (**code wins**), exit codes, snapshot refresh workflow (figma MCP → `use_figma`), and drift resolution actions per finding kind.

### Scope decisions baked into this PR

- **Color category only.** `size`/`radius`/`borderwidth` (#61), `typography` (#62), `shadow` (#63) error out with `exit=2` citing their tracking issue. Sub-issues each wire one category.
- **Designer pushes focus values to Figma manually.** This PR ships the audit infrastructure; the issue's Part 1 (push `focus.default.color` / `focus.error.color` to Figma) stays open as designer work. The focus drift documented in the issue body lives in the `focus/` primitive category, not the `color/` palette audited here.
- **Snapshot refresh is manual via figma MCP.** Not auto-refreshed in CI. Documented.

### Architecture

- `diffTokenTrees(codeSubtree, snapshotSubtree)` is the pure core — parsed objects in, sorted finding list out. No fs, no env, no globals. CLI's `main()` is a thin wrapper that does file I/O.
- Exit codes: `0` no drift / `1` drift / `2` config error.
- Hex values normalized to lowercase before compare; alpha-bearing 8-digit hex preserved exactly.
- Snapshot format mirrors per-category code JSON shape — `snapshot.color.<palette>.<shade>` matches `color.json`'s `<palette>.<shade>` directly.

## GitHub Issue

Refs #56

(The issue stays open until the designer pushes the focus color values to Figma and a follow-up commit refreshes the snapshot showing zero drift in the focus category too. Color category itself is at zero drift today.)

## Test Plan

- [ ] `yarn test:unit` — 11 new tests pass, plus all existing
- [ ] `yarn audit:figma-parity --category color` — reports `✓ no drift`, exits 0
- [ ] `yarn audit:figma-parity` (no args) — fails with `--category required` and lists supported + pending categories, exits 2
- [ ] `yarn audit:figma-parity --category size` — fails citing #61, exits 2
- [ ] `yarn audit:figma-parity --category bogus` — fails with unknown category, exits 2
- [ ] Temporarily delete `packages/core/tokens/figma-snapshot.json`, re-run `--category color` — fails with snapshot-missing message pointing at `.claude/rules/figma.md`, exits 2
- [ ] Temporarily edit one shade in `color.json` (e.g., `blue.500`), re-run `--category color` — surfaces one `value-mismatch` finding with the canonical "push code value to Figma" guidance
- [ ] `yarn lint` and `yarn format:check` — clean on changed files


packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+import { extractTokens, readTokenFile } from './utils.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const TOKENS_DIR = path.resolve(__dirname, '..', 'tokens');
+const PRIMITIVES_DIR = path.join(TOKENS_DIR, 'primitives');
+const DEFAULT_SNAPSHOT = path.join(TOKENS_DIR, 'figma-snapshot.json');
+
+const CATEGORIES = {
+  color: {
+    kind: 'single',
```
examlly-tech: **Discuss — multi-mode shape:** `kind: 'single'` implies a future `'multi-mode'` variant but doesn't sketch it. Sub-issues #61/#62/#63 each face categories with N mode files (e.g., `size-vega`, `size-lyra`). Land the multi-mode shape decision here so each sub-issue inherits it.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+    file: path.join(PRIMITIVES_DIR, 'color.json'),
+  },
+};
+
+const PENDING_CATEGORIES = {
+  size: '#61',
+  radius: '#61',
+  borderwidth: '#61',
```
examlly-tech: **Discuss — flag grouping mismatch:** `borderwidth` routes to #61 as its own flag, but #61 covers size+radius+borderwidth together. Sub-issue PR has to invoke audit three times. Consider grouping under `--category dimensional` or documenting the 3-invocation requirement.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+      `--category ${args.category} is not yet supported. Tracked in ${PENDING_CATEGORIES[args.category]}.`
+    );
+  }
+
+  const config = CATEGORIES[args.category];
+  if (!config) {
+    fail(`unknown category "${args.category}"`);
+  }
+
+  if (!fs.existsSync(args.snapshot)) {
```
examlly-tech, line 223: **Concern — stale snapshot risk:** Manual snapshot + "code wins" rule means a stale snapshot silently passes audit while real drift exists. Add an `_meta.capturedAt` staleness check, or document this as a known limitation pending CI integration.

.claude/rules/figma.md
```diff
@@ -12,6 +12,55 @@
...
+
+Exit codes:
+
+| Code | Meaning                                                          |
+| ---- | ---------------------------------------------------------------- |
+| 0    | No drift                                                         |
+| 1    | Drift findings reported                                          |
+| 2    | Configuration error (snapshot missing, unknown/pending category) |
+
+Categories supported today: `color`. Categories pending — wire up alongside the linked issue: `size`/`radius`/`borderwidth` (#61), `typography` (#62), `shadow` (#63).
```
examlly-tech: **Minor — drift-issue closure semantics:** Doc says "zero drift is the merge bar for the issue that prompted the audit" — but with manual snapshot refresh, this is a snapshot of agreement at one moment, not an ongoing contract. Clarify the merge bar is "audit passes at PR open."

examlly-tech (commented):

### Issues

| 1 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:12-17` | `CATEGORIES.color` ships `kind: 'single'` but no `kind: 'multi-mode'` shape — every sub-issue will reinvent the snapshot key layout. | Sketch the multi-mode snapshot shape now (e.g., `snapshot.size.vega.<shade>`) and land a stub for #61. |
| 2 | ⚠️ Discuss | `.claude/rules/figma.md:39` | `borderwidth` is exposed as its own `--category` flag but #61 groups size+radius+borderwidth — sub-issue PR has to run audit 3 times. | Either group `--category dimensional` covering all three, or document the 3-invocation pattern in #61. |
| 3 | ⚠️ Minor | `packages/core/tokens/figma-snapshot.json:2-8` | `_meta` has one flat `variableCollection` — won't extend to multi-mode/themed categories where capture state varies per file. | Move `_meta` next to each category subtree, or omit until shape is known. |

### Architectural Concerns

**Stale-snapshot masks drift** (`packages/core/scripts/audit-figma-parity.js:148-152`) — Manual refresh via figma MCP plus "code wins" canonical rule means a stale snapshot makes the audit pass while real drift exists. Add a snapshot-freshness check (e.g., warn if `_meta.capturedAt` is older than N days) or wire to CI before depending on this for merge gating.

**Multi-mode data shape is the load-bearing decision** (`packages/core/tokens/figma-snapshot.json:9`) — The snapshot top-level mirrors `color.json` (palette → shade), but `size-vega.json` / `size-lyra.json` etc. don't have a single canonical shape. Choosing this once now (in #65) prevents three divergent shapes across #61/#62/#63.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+    const figma = figmaMap.get(tokenPath);
+    if (!figma) {
+      findings.push({ path: tokenPath, kind: 'missing-in-figma', code });
+      continue;
+    }
+    if (code.type !== figma.type) {
+      findings.push({ path: tokenPath, kind: 'type-mismatch', code, figma });
+      continue;
+    }
+    if (code.value !== figma.value) {
```
examlly-tech, line 121: **Minor — reference equality on `$value`:** Once `--category size`/`radius` (#61) wires dimension tokens (`{value, unit}` objects), every token reports `value-mismatch` because `!==` compares object identity. Normalize to a canonical string in `buildTokenMap` before comparison.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:89` | `code.value !== figma.value` is reference equality; once #61 wires `size`/`radius`/`borderwidth`, dimension tokens (`{value, unit}` objects) all report as `value-mismatch`. | Normalize to a canonical string in `buildTokenMap` (e.g. `${v.value}${v.unit ?? 'px'}` for dimensions) so the pure core works for non-string values. |
| 2 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:40` | `--category --snapshot foo` swallows `--snapshot` as the category value, then prints `unknown category "--snapshot"`. | Guard `argv[i+1]?.startsWith('--')` and treat as missing-value, or require `--key=value` form only. |
| 3 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:107` | `formatFinding` interpolates `.value` directly; once a non-string value reaches it (issue 1) the line becomes `code=[object Object]`. | Same root cause as 1 — one normalization fixes both. |
| 4 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:74` | JSDoc restates what the code does ("symmetric inputs in, symmetric findings out — no fs, no env, no globals"). Rationale prose per `code-comments.md`. | Drop the second sentence; keep param/return docs. |
| 5 | ⚠️ Minor | `packages/core/tokens/figma-snapshot.json:2` | `_meta` is safe today because `main()` reads only `snapshot[args.category]`, but `extractTokens` skips `$`-prefixed keys only — a whole-file walk would recurse into `_meta`. Also flat `_meta` won't survive multi-mode categories. | Rename `_meta` → `$meta` so DTCG convention covers it; move per-category once multi-mode lands. |

### Architectural Concerns

**Pure-core contract mismatch** (`packages/core/scripts/audit-figma-parity.js:74`) — `diffTokenTrees`' JSDoc bills it as the reusable core for all categories, but its value comparison only works for strings. Either fix normalization now (issue 1) or narrow the docstring to color and `TODO(#61)` the rest.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+function parseArgs(argv) {
+  const args = { category: null, snapshot: DEFAULT_SNAPSHOT };
+  for (let i = 0; i < argv.length; i++) {
+    const arg = argv[i];
+    if (!arg.startsWith('--')) continue;
+    const eq = arg.indexOf('=');
+    if (eq !== -1) {
+      args[arg.slice(2, eq)] = arg.slice(eq + 1);
+    } else {
+      args[arg.slice(2)] = argv[++i];
```
examlly-tech: **Minor — arg parser consumes next flag:** `--category --snapshot foo` sets `category = "--snapshot"` and never assigns `snapshot`. Guard `argv[i+1]?.startsWith('--')` or require `--key=value` form only.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  return findings.sort((a, b) => a.path.localeCompare(b.path));
+}
+
+function formatFinding(finding) {
+  const label = finding.path.padEnd(40);
+  switch (finding.kind) {
+    case 'value-mismatch':
+      return `  ✗ ${label} code=${finding.code.value}  figma=${finding.figma.value}  → push ${finding.code.value} to Figma`;
```
examlly-tech, line 139: **Minor — `[object Object]` once non-string values land:** Template interpolation of `finding.code.value` will stringify dimension objects as `[object Object]`. Same root cause as the equality issue at line 89 — one normalization fixes both.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Pure diff function: takes two parsed DTCG token subtrees and returns a sorted
+ * list of drift findings. Symmetric inputs in, symmetric findings out — no fs,
+ * no env, no globals. Unit-testable.
+ *
+ * @param {object} codeSubtree - Parsed JSON for the code side (e.g., color.json contents)
+ * @param {object} snapshotSubtree - Parsed JSON for the Figma snapshot side (same shape)
+ * @returns {{ path: string, kind: string, code?: object, figma?: object }[]}
+ */
+export function diffTokenTrees(codeSubtree, snapshotSubtree) {
```
examlly-tech, line 106: **Minor — rationale prose in JSDoc:** "Symmetric inputs in, symmetric findings out — no fs, no env, no globals. Unit-testable." restates the code. Per `code-comments.md` drop it; param/return docs already carry the contract.

packages/core/tokens/figma-snapshot.json
```diff
@@ -0,0 +1,299 @@
+{
+  "_meta": {
```
examlly-tech: **Architecture:** Flat `_meta` won't survive multi-mode categories where capture state differs per file — consider per-category metadata. **Code quality:** Rename `_meta` → `$meta` so `extractTokens` (which skips only `$`-prefixed keys) ignores it in any future whole-file walk.

packages/core/tokens/figma-snapshot.json
```diff
@@ -0,0 +1,299 @@
+{
+  "color": {
```
examlly-tech, line 2: **Discuss — snapshot shape:** `snapshot.color` works because color is single-mode, but size/radius/borderwidth/typography are multi-mode (`size-vega.json`, `size-lyra.json`…) and shadow is mode×theme. #61/#62/#63 need a slot for mode. Decide the shape now — nested (`snapshot.size.vega`) or `--mode` flag — so sub-issues don't retrofit both file and CLI.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,201 @@
...
+const PRIMITIVES_DIR = path.join(TOKENS_DIR, 'primitives');
+const DEFAULT_SNAPSHOT = path.join(TOKENS_DIR, 'figma-snapshot.json');
+
+const CATEGORIES = {
+  color: {
+    file: path.join(PRIMITIVES_DIR, 'color.json'),
+  },
+};
+
+const PENDING_CATEGORIES = {
```
examlly-tech: **Discuss — drop `PENDING_CATEGORIES`:** Duplicates the doc's pending-category table and adds a two-place edit (`delete from PENDING` + `add to CATEGORIES`) for every future PR. Let the unknown-category error point readers at `.claude/rules/figma.md` and keep the live list in one place.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,201 @@
...
+    );
+  }
+
+  const codeData = readTokenFile(config.file);
+  const findings = diffTokenTrees(codeData, snapshotSubtree);
+
+  const meta = snapshotSubtree.$meta;
+  const lines = [];
+  lines.push(`─── audit-figma-parity --category ${args.category} ───`);
+  if (meta?.capturedAt) {
```
examlly-tech, line 266: **Discuss — stale-snapshot guard:** Audit prints `capturedAt` but never compares it. A snapshot older than the last `color.json` edit silently masks drift because both sides agree on the stale value. Consider failing when snapshot mtime predates the audited file, or warning past an age threshold.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,201 @@
...
+  size: '#61',
+  radius: '#61',
+  borderwidth: '#61',
+  typography: '#62',
+  shadow: '#63',
+};
+
+const EXIT_OK = 0;
+const EXIT_DRIFT = 1;
+const EXIT_CONFIG = 2;
```
examlly-tech, line 36: **Minor — overloaded exit code:** `EXIT_CONFIG=2` covers missing flag, pending category, unknown category, missing snapshot, and missing subtree. CI can't distinguish them. Either accept stderr as the contract or split (e.g., `EXIT_PENDING=3`, `EXIT_SNAPSHOT_MISSING=4`).

examlly-tech (commented):

### Issues

| 1 | ⚠️ Discuss | `packages/core/tokens/figma-snapshot.json:1-9` | Snapshot key shape (`snapshot.<category>`) has no slot for the mode dimension. Code stores size/radius/borderwidth/typography as `{category}-{mode}.json` and shadow as `{category}-{mode}-{theme}.json` — #61/#62/#63 will need to pick a mode or extend the shape. | Decide upfront: nested by mode (`snapshot.size.vega`, `snapshot.shadow.vega.light`) or a `--mode` flag with flat keys. Pick one before #61 starts so sub-issues don't retrofit both the file and the CLI. |
| 2 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:18-24` | `PENDING_CATEGORIES` encodes unimplemented-feature metadata in source and duplicates the doc table. Each future category PR has to delete an entry here _and_ add one to `CATEGORIES`. | Drop `PENDING_CATEGORIES`. Let unknown-category error point at `.claude/rules/figma.md` for the live list. One source of truth. |
| 3 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:173-179` | Audit prints `capturedAt` but never compares it. A snapshot older than the most recent edit to `color.json` silently masks drift because both sides agree on the stale value. | Either fail when snapshot mtime predates the audited code file, or warn when `capturedAt` is older than N days. Doc-only acknowledgement isn't a guard. |
| 4 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:26-28` | `EXIT_CONFIG=2` is overloaded across five distinct failure modes (missing flag, pending category, unknown category, missing snapshot file, missing subtree). CI can't route them. | Split into typed exits (e.g., `EXIT_USAGE=2` / `EXIT_PENDING=3` / `EXIT_SNAPSHOT_MISSING=4`) or document stderr as the contract. |

### Architectural Concerns

**Multi-mode extensibility** (`figma-snapshot.json:1-9`) — Color is single-mode in code so the flat `snapshot.color` shape works today. Every other pending category is multi-mode; the shape needs an answer before #61 lands, otherwise #61 retrofits the snapshot file and the CLI surface at the same time.

**Source-of-truth duplication for pending categories** (`audit-figma-parity.js:18-24`, `.claude/rules/figma.md:39`) — The mapping `{ size: '#61', radius: '#61', ... }` lives in both code and docs; deletions must happen in two places per future PR. Per `code-quality.md` (fewer moving parts), keep it in the doc only.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,201 @@
...
+  if (
+    type === 'dimension' &&
+    typeof value === 'object' &&
+    value !== null &&
+    'value' in value
+  ) {
+    const rounded = Math.round(value.value * 10000) / 10000;
+    return `${rounded}${value.unit || 'px'}`;
+  }
+  return value;
```
examlly-tech, line 86: **Minor — composite fall-through:** `normalizeValue` returns the raw value for any non-color, non-dimension `$type`. For object-shaped composites (shadow, typography) the Map stores object refs, so `code.value !== figma.value` reports a spurious `value-mismatch` for every token. Either `JSON.stringify` the object or assert on supported types so the pending-category wiring fails loudly.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,201 @@
...
+    fail(`unknown category "${args.category}"`);
+  }
+
+  if (!fs.existsSync(args.snapshot)) {
+    fail(
+      `snapshot file not found: ${args.snapshot}\n  Refresh the snapshot via the figma MCP — see .claude/rules/figma.md.`
+    );
+  }
+
+  const snapshot = readTokenFile(args.snapshot);
```
examlly-tech: **Minor — unhandled parse error:** `readTokenFile` throws on malformed JSON. `main()` doesn't catch, so a corrupt snapshot prints a stack trace instead of the documented `EXIT_CONFIG=2`. Wrap both `readTokenFile` calls in try/catch and route the error through `fail()`.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,201 @@
...
+  switch (finding.kind) {
+    case 'value-mismatch':
+      return `  ✗ ${label} code=${finding.code.value}  figma=${finding.figma.value}  → push ${finding.code.value} to Figma`;
+    case 'type-mismatch':
+      return `  ✗ ${label} type drift: code=${finding.code.type}  figma=${finding.figma.type}`;
+    case 'missing-in-figma':
+      return `  ✗ ${label} missing in Figma (code value: ${finding.code.value}) → add to Figma`;
+    case 'missing-in-code':
+      return `  ✗ ${label} extra in Figma (figma value: ${finding.figma.value}) → remove from Figma`;
+    default:
```
examlly-tech, line 146: **Minor — dead default branch:** `diffTokenTrees` only emits four kinds. The `default` case is unreachable. Drop it — the four cases are exhaustive.

packages/core/scripts/__tests__/audit-figma-parity.test.js
```diff
@@ -0,0 +1,184 @@
...
+  it('skips DTCG metadata keys ($-prefixed) on either side', () => {
+    const code = {
+      blue: { 500: colorToken('#3b82f6') },
+    };
+    const figma = {
+      $meta: { capturedAt: '2026-05-21' },
+      blue: { 500: colorToken('#3b82f6') },
+    };
+    expect(diffTokenTrees(code, figma)).toEqual([]);
+  });
```
examlly-tech, line 167: **Minor — missing combined-kind fixture:** all current tests exercise one drift kind per fixture. Add a scenario with `type-mismatch` + `value-mismatch` + `missing-in-figma` in one tree to lock in sort stability and finding independence.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:62` | `normalizeValue` falls through for composite types (shadow, typography) and stores raw object refs in the map — every token would falsely diff as `value-mismatch` once those categories are unlocked. | Either stringify unknown object values (e.g., `JSON.stringify`) or assert on supported types so future category wiring fails loudly instead of silently. |
| 2 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:162,170` | `readTokenFile` (`JSON.parse`) throws on malformed JSON — `main()` doesn't catch, so a corrupt snapshot prints a stack trace instead of the documented `EXIT_CONFIG=2`. | Wrap the two `readTokenFile` calls in try/catch and route through `fail()`. |
| 3 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:122` | `formatFinding` default branch is unreachable — `diffTokenTrees` only emits four kinds (`value-mismatch`, `type-mismatch`, `missing-in-figma`, `missing-in-code`). | Drop the `default` case; the four cases are exhaustive. |
| 4 | ⚠️ Minor | `packages/core/scripts/__tests__/audit-figma-parity.test.js` | All current tests exercise one drift kind per fixture. Missing a combined-drift scenario (`type-mismatch` + `value-mismatch` + `missing-in-figma` in one tree) to lock in finding independence and sort stability. | Add one fixture mixing all kinds at once; assert path order + kinds. |

examlly-tech (commented):

### Issues

| 1 | ⚠️ Discuss | `.claude/rules/figma.md:66` | "Multi-mode CLI invocation (likely a `--mode <name>` flag)" hedges the contract this PR claims to lock. #61 can still pick walk-all-modes vs per-mode without breaking the doc, so the shape isn't actually committed end-to-end. | Pick one pattern (`--mode` required for multi-mode categories) and drop "likely". |
| 2 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:14-18` | `CATEGORIES` is shaped for one file per category (`{file: <path>}`). Multi-mode categories need either `{files: {mode: path}}` or `{dir, modePattern}`; the shape #61 will need isn't sketched here. | Document the multi-mode entry shape now (in code as a stub comment, or in figma.md) so #61 fills in paths, not redesigns the registry. |
| 3 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:148-164` | `gitLastCommitTs` returns `null` on any git failure (shallow clone, missing git, file not tracked yet), silently disabling the stale-snapshot guard. The guard is the merge bar for the audit but the failure mode is invisible. | Fail closed when both timestamps are null and the snapshot is tracked, or log a one-liner stating the guard was skipped. |
| 4 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:27` | `--snapshot` is parsed and tested but not documented in figma.md; users only see `--category`. Hidden CLI surface ages badly. | Either document it under "Running the audit" or rename to a test-only env var. |

### Architectural Concerns

**Snapshot shape vs code-file shape mismatch** (`.claude/rules/figma.md:58-62`) — Doc commits snapshot as `snapshot.size.<mode>.<keys>` (all modes nested) but code files are per-mode (`size-vega.json` is just `{0: ..., 1: ...}`). The diff core walks whatever subtree it gets, so #61 has to decide: snapshot-side slices via `--mode`, or code-side merges via a multi-file loader. Pick one in this PR, otherwise the "shape is committed" claim is half-true.

.claude/rules/figma.md
```diff
@@ -12,6 +12,82 @@
...
+
+| Category type | Code layout                                    | Snapshot key shape                      |
+| ------------- | ---------------------------------------------- | --------------------------------------- |
+| Single-mode   | `primitives/color.json`                        | `snapshot.color.<palette>.<shade>`      |
+| Multi-mode    | `primitives/{cat}/{cat}-{mode}.json`           | `snapshot.<cat>.<mode>.<keys>`          |
+| Mode × theme  | `primitives/shadow/shadow-{mode}-{theme}.json` | `snapshot.shadow.<mode>.<theme>.<keys>` |
+
+Each category subtree carries a `$meta` block — at minimum `capturedAt` (ISO date) and `figmaFileName`. The `$`-prefix is DTCG metadata and is skipped by the token walker, so it never appears as a drift finding.
+
+Multi-mode CLI invocation (likely a `--mode <name>` flag) is locked in by #61 alongside its category wiring; single-mode color today only needs `--category color`.
```
examlly-tech: **Discuss — multi-mode CLI not locked:** "likely a `--mode <name>` flag" hedges the contract this PR claims to commit. #61 can still pick walk-all-modes vs per-mode invocation. Pick one and drop the hedge so sub-issues only fill paths, not redesign the surface.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+const TOKENS_DIR = path.resolve(__dirname, '..', 'tokens');
+const PRIMITIVES_DIR = path.join(TOKENS_DIR, 'primitives');
+const REPO_ROOT = path.resolve(__dirname, '..', '..', '..');
+const DEFAULT_SNAPSHOT = path.join(TOKENS_DIR, 'figma-snapshot.json');
+
+const CATEGORIES = {
+  color: {
+    file: path.join(PRIMITIVES_DIR, 'color.json'),
+  },
+};
```
examlly-tech, line 30: **Discuss — registry shape:** `{file: <path>}` matches single-mode color only. Multi-mode (#61) needs `{files: {mode: path}}` or `{dir, modePattern}`. Document the multi-mode entry shape now so #61 fills in paths, not redesigns the registry.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+      {
+        encoding: 'utf8',
+        cwd: REPO_ROOT,
+        stdio: ['ignore', 'pipe', 'ignore'],
+      }
+    ).trim();
+    const ts = parseInt(output, 10);
+    return Number.isFinite(ts) ? ts : null;
+  } catch {
+    return null;
```
examlly-tech, line 177: **Discuss — guard silently disables:** `gitLastCommitTs` returns null on any git failure and the audit proceeds. Stale-snapshot is the merge bar but the failure mode is invisible. Fail closed if the snapshot is tracked but timestamp lookup failed, or log when guard is skipped.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+};
+
+const EXIT_OK = 0;
+const EXIT_DRIFT = 1;
+const EXIT_CONFIG = 2;
+
+const DOCS_HINT = 'see .claude/rules/figma.md → Code-vs-Figma Parity Audit';
+
+export function parseArgs(argv) {
+  const args = { category: null, snapshot: DEFAULT_SNAPSHOT };
```
examlly-tech, line 49: **Minor — undocumented CLI flag:** `--snapshot` is parsed and tested but not in figma.md. Users only see `--category`. Document it under "Running the audit" or rename to a test-only env var.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+function readCodeFileOrFail(filePath) {
+  try {
+    return readTokenFile(filePath);
+  } catch (err) {
+    fail(`failed to parse code file ${filePath}: ${err.message}`);
+  }
+}
```
examlly-tech, line 161: **Minor — DRY:** `readSnapshotOrFail` and `readCodeFileOrFail` differ only in the error label. Collapse to `readJsonOrFail(filePath, label)` and pass `'snapshot'` / `'code file'`. Two helpers earn nothing the parameter doesn't.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+    const arg = argv[i];
+    if (!arg.startsWith('--')) continue;
+    const eq = arg.indexOf('=');
+    if (eq !== -1) {
+      args[arg.slice(2, eq)] = arg.slice(eq + 1);
+      continue;
+    }
+    const next = argv[i + 1];
+    if (next && !next.startsWith('--')) {
+      args[arg.slice(2)] = next;
```
examlly-tech: **Minor — adversarial input:** `--` and `--=value` set `args[''] = next` and silently eat the next argv slot. Guard with `if (arg === '--' || eq === 2) continue;` so bare `--` falls through cleanly.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+  process.stderr.write(`audit-figma-parity: ${message}\n`);
+  process.exit(EXIT_CONFIG);
+}
+
+function readSnapshotOrFail(filePath) {
+  try {
+    return readTokenFile(filePath);
+  } catch (err) {
+    fail(`failed to parse snapshot ${filePath}: ${err.message}`);
+  }
```
examlly-tech: **Minor — control flow:** `fail()` calls `process.exit`, but if exit is stubbed (test harness) this helper returns `undefined` and `main` crashes accessing `snapshot[args.category]`. Add `return` after `fail()`, or have `fail` throw and let `main` catch.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+  switch (finding.kind) {
+    case 'value-mismatch':
+      return `  ✗ ${label} code=${finding.code.value}  figma=${finding.figma.value}  → push ${finding.code.value} to Figma`;
+    case 'type-mismatch':
+      return `  ✗ ${label} type drift: code=${finding.code.type}  figma=${finding.figma.type}`;
+    case 'missing-in-figma':
+      return `  ✗ ${label} missing in Figma (code value: ${finding.code.value}) → add to Figma`;
+    case 'missing-in-code':
+      return `  ✗ ${label} extra in Figma (figma value: ${finding.figma.value}) → remove from Figma`;
+  }
```
examlly-tech, line 148: **Minor — exhaustiveness:** `formatFinding` switch has no default; a future `kind` returns `undefined` and prints a literal `undefined` line. Add `default: throw new Error(\`unknown kind: ${finding.kind}\`)` to lock the contract.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,238 @@
...
+    return value.toLowerCase();
+  }
+  if (
+    type === 'dimension' &&
+    typeof value === 'object' &&
+    value !== null &&
+    'value' in value
+  ) {
+    const rounded = Math.round(value.value * 10000) / 10000;
+    return `${rounded}${value.unit || 'px'}`;
```
examlly-tech, line 78: **Minor — `||` vs `??`:** `value.unit || 'px'` coerces empty-string `unit: ''` to `'px'`. Use `??` so only `null`/`undefined` defaults; malformed empty units stay visible as value-mismatch findings.

packages/core/scripts/__tests__/audit-figma-parity.test.js
```diff
@@ -0,0 +1,220 @@
...
+
+  it('throws when an object value reaches an unsupported $type (locks contract before composite categories land)', () => {
+    const code = {
+      body: {
+        $value: { fontFamily: 'Inter', fontSize: '16px' },
+        $type: 'typography',
+      },
+    };
+    expect(() => diffTokenTrees(code, {})).toThrow(/normalizeValue/);
```
examlly-tech: **Minor — brittle assertion:** `/normalizeValue/` matches the function name. If the wording moves, the test breaks. Prefer `/object value for \$type/` — asserts the structural reason, not the function name.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:132-146` | `readSnapshotOrFail` and `readCodeFileOrFail` are near-identical; only the error label differs. Two helpers where one parameterized helper reads cleaner. | Collapse to `readJsonOrFail(filePath, label)`; call sites pass `'snapshot'` / `'code file'`. |
| 2 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:26-43` | `parseArgs` silently writes `args[''] = next` for bare `--` or `--=value`, and consumes the following argv slot. Adversarial input is swallowed, not flagged. | Skip when `arg === '--'` or `arg.slice(2, eq).length === 0`. |
| 3 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:127-146` | `fail()` calls `process.exit` but is untyped/uncommented as terminating. If `process.exit` is stubbed (tests, harness), `readSnapshotOrFail`/`readCodeFileOrFail` return `undefined` and `main` crashes on `snapshot[args.category]`. | Add explicit `return` after `fail()` in the try/catch helpers, or have `fail` throw and let `main` catch. |
| 4 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:113-125` | `formatFinding` has no default branch and no exhaustiveness guard. A future `kind` slips through and returns `undefined`, surfacing as a literal `undefined` line in the report. | Add `default: throw new Error(\`formatFinding: unknown kind "${finding.kind}"\`)` to lock the contract. |
| 5 | ⚠️ Minor | `packages/core/scripts/__tests__/audit-figma-parity.test.js:190-198` | The unsupported-type test asserts on `/normalizeValue/` substring of the error message — brittle if the wording moves. | Assert on the structural part instead (`/object value for \$type/`) or on a stable `err.name`/`err.code`. |
| 6 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:55-56` | `value.unit \|\| 'px'` uses `\|\|`, so `unit: ''` (empty string) falls through to `'px'`. For a typed dimension token an empty string is invalid input that should surface as drift, not silently coerce. | Use `??` so only `null`/`undefined` defaults; empty `unit` stays visible as `${value}` and shows up in value-mismatch findings. |

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,259 @@
...
+ *      with `--mode <name> --theme <variant>`, diff `snapshot[cat][mode][theme]` against the matching file.
+ *
+ * #61/#62/#63 add their categories by filling paths into this map. Snapshot/CLI shape is
+ * locked in .claude/rules/figma.md — Code-vs-Figma Parity Audit.
+ */
+const CATEGORIES = {
+  color: {
+    file: path.join(PRIMITIVES_DIR, 'color.json'),
+  },
+};
```
examlly-tech, line 30: **Discuss — locked contract not wired:** JSDoc declares three registry shapes (single/multi/mode×theme), but `main()` only consumes `config.file` and ignores `args.mode`/`args.theme`. #61/#62/#63 will still design routing, not just add a row. Either wire the resolver now or drop the 'locked' claim.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,259 @@
...
+    const snapDate = new Date(snapshotTs * 1000).toISOString().slice(0, 10);
+    fail(
+      `snapshot is stale: ${path.basename(config.file)} was last committed ${codeDate}, after snapshot's ${snapDate}. Refresh per .claude/rules/figma.md.`
+    );
+  }
+
+  const codeData = readJsonOrFail(config.file, 'code file');
+  const findings = diffTokenTrees(codeData, snapshotSubtree);
+
+  const meta = snapshotSubtree.$meta;
```
examlly-tech, line 263: **Discuss — `$meta` read assumes single-mode shape:** figma.md:73 puts multi-mode `$meta` at `snapshot.size.vega.$meta`, but this reads `snapshotSubtree.$meta` (category-level). Same gap as the routing — both need to flow through one resolver.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,259 @@
...
+  const snapshotSubtree = snapshot[args.category];
+  if (!snapshotSubtree) {
+    fail(
+      `snapshot has no "${args.category}" subtree. Refresh the snapshot or check the category name.`
+    );
+  }
+
+  const codeTs = gitLastCommitTs(config.file);
+  const snapshotTs = gitLastCommitTs(args.snapshot);
+  const staleGuardSkipped = codeTs === null || snapshotTs === null;
```
examlly-tech, line 239: **Discuss — skip is silent in CI:** shallow-clone CI hits `null` timestamps, prints `⚠ skipped` to stdout, exits 0. CI log filters can hide that line. Consider hard-failing (exit 2) when `process.env.CI` is set so the guard's CI promise actually holds.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:26-30, 175-218` | "Locked contract" for multi-mode/theme exists only in JSDoc + figma.md — `parseArgs` accepts `--mode`/`--theme` silently, `CATEGORIES` has no `dir`/`filePattern` handling, and `main()` never reads `args.mode`/`args.theme`. #61/#62/#63 will still touch routing, not just paths. | Either land a no-op multi-mode resolver now (return the file path from `dir + filePattern + args.mode[+args.theme]`, validate `args.mode` against an allowed list per entry) so #61 truly just adds one registry row, or downgrade the doc claim from "locked contract" to "intended shape — wired alongside #61." |
| 2 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:220-229` | `snapshotSubtree.$meta` read is hard-coded to category-level, but figma.md:73 declares multi-mode `$meta` lives at `snapshot.size.vega.$meta`. Same locked-contract gap as #1, visible in the print path. | When (1) is resolved, route the meta lookup through the same resolver so the printed `Snapshot:` line picks the right `$meta` for single- vs multi-mode. |
| 3 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:157-173, 208, 227-229` | Stale-guard prints `⚠ skipped` when `git log -1` returns empty — covers the shallow-clone CI signal, but the audit still exits 0 with drift undetectable as stale. CI on a shallow clone will green-light a stale snapshot. | Gate the skip on `process.env.CI` so CI fails closed (exit 2) and local stays soft. Status quo lets a stale snapshot ship under CI cover. |

### Architectural Concerns

**Locked-contract claim is aspirational, not enforced** (`packages/core/scripts/audit-figma-parity.js:14-25`). The JSDoc registry shape and figma.md "locked contract for #61/#62/#63" describe paths the code doesn't execute; sub-issue PRs will still design the multi-mode routing and the per-mode `$meta` read. Either build the resolver now or stop calling it locked.

**Stale-guard skip is silent-to-CI** (`packages/core/scripts/audit-figma-parity.js:208`). The on-screen `⚠ skipped` line is invisible in non-interactive CI logs that get filtered; the exit code is the contract CI reads. Without a CI-mode hard-fail on skip, the guard's promise ("blocks stale snapshots") doesn't hold in the one environment that matters.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,259 @@
...
+
+const DOCS_HINT = 'see .claude/rules/figma.md → Code-vs-Figma Parity Audit';
+
+class ConfigError extends Error {}
+
+export function parseArgs(argv) {
+  const args = { category: null, snapshot: DEFAULT_SNAPSHOT };
+  for (let i = 0; i < argv.length; i++) {
+    const arg = argv[i];
+    if (!arg.startsWith('--')) continue;
```
examlly-tech: **Minor — loose flag prefix check:** `arg.startsWith('--')` accepts `---foo`, which then writes `args['-foo'] = next` and silently swallows a real arg. Add a `/^--[A-Za-z]/` shape check before treating `arg` as a flag.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,259 @@
...
+    return value.toLowerCase();
+  }
+  if (
+    type === 'dimension' &&
+    typeof value === 'object' &&
+    value !== null &&
+    'value' in value
+  ) {
+    const rounded = Math.round(value.value * 10000) / 10000;
+    return `${rounded}${value.unit ?? 'px'}`;
```
examlly-tech: **Minor — `??` vs `||` drift:** `value.unit ?? 'px'` accepts empty-string unit; `utils.js:57` `formatTokenValue` uses `||` and would replace it. Align on one — preferably extract to a shared dimension formatter so both stay in sync.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  const config = CATEGORIES[args.category];
+  if (!config) {
+    fail(
+      `unknown category "${args.category}"\n  supported: ${Object.keys(CATEGORIES).join(', ')}\n  pending categories — ${DOCS_HINT}`
+    );
+  }
+
+  if (!fs.existsSync(args.snapshot)) {
```
examlly-tech, line 223: **Minor — empty-string snapshot:** `--snapshot=` parses to `args.snapshot = ''`, then `fs.existsSync('')` is false and the user sees `snapshot file not found:` with a blank path. Normalize empty to `undefined` before the existence check.

packages/core/scripts/__tests__/audit-figma-parity.test.js
```diff
@@ -0,0 +1,220 @@
...
+    const args = parseArgs(['--category', '--snapshot', 'foo.json']);
+    expect(args.category).toBe(null);
+    expect(args.snapshot).toBe('foo.json');
+  });
+
+  it('handles a trailing bare --flag without crashing', () => {
+    const args = parseArgs(['--category']);
+    expect(args.category).toBe(null);
+  });
+});
```
examlly-tech, line 248: **Minor — missing coverage for new guards:** the `eq === 2` and `arg === '--'` skips landed in round 3 but no test exercises them. Add `--=value`, `--`, and `---foo` cases so a future regression of the guard logic surfaces here.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:42-56` | `---foo bar` writes `args['-foo'] = 'bar'`; only `arg.startsWith('--')` is checked, not that the char after `--` is alphabetic. Junk flags silently captured. | Reject any arg matching `/^---/` or where the key portion starts with `-`. |
| 2 | ⚠️ Minor | `packages/core/scripts/__tests__/audit-figma-parity.test.js:201-220` | New `parseArgs` guards (`eq === 2` skip, `arg === '--'`) lack direct test coverage; a regression of either would slip through. | Add cases: `parseArgs(['--=value'])` produces no extraneous key, `parseArgs(['--', '--category', 'color'])` sets category, `parseArgs(['---foo', 'bar'])` ignored. |
| 3 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:192-196` | `--snapshot=` (empty value) parses to `args.snapshot = ''`, then `fs.existsSync('')` is false and the user sees `snapshot file not found:` with a blank path — confusing. | After `parseArgs`, treat empty-string `snapshot`/`category` as if absent (fall through to default / required error). |
| 4 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:61-81` | `normalizeValue` uses `value.unit ?? 'px'` while sibling `formatTokenValue` in `utils.js:57` uses `value.unit \|\| 'px'`. Empty-string `unit` would survive here but be replaced there — inconsistency between two dimension formatters. | Align on `\|\|` to match `utils.js`, or extract a shared dimension formatter so both stay in sync. |

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,267 @@
...
+  if (!snapshotSubtree) {
+    fail(
+      `snapshot has no "${args.category}" subtree. Refresh the snapshot or check the category name.`
+    );
+  }
+
+  const codeTs = gitLastCommitTs(config.file);
+  const snapshotTs = gitLastCommitTs(args.snapshot);
+  const staleGuardSkipped = codeTs === null || snapshotTs === null;
+  if (staleGuardSkipped && process.env.CI) {
```
examlly-tech: **Discuss — CI fetch-depth coupling:** Hard-fail depends on full git history. `actions/checkout@v4` defaults to `fetch-depth: 1`, so files whose last commit is outside the shallow window return null and trip the fail. Document `fetch-depth: 0` in figma.md, or detect via `git rev-parse --is-shallow-repository`.

examlly-tech: **Minor — CI truthiness:** `process.env.CI` is a string; `CI=0`, `CI=false`, and `CI=anything` all evaluate truthy, so opt-out values still hard-fail. Use allowlist `['true', '1'].includes(CI?.toLowerCase())` or exclude opt-out `CI && CI !== '0' && CI !== 'false'`.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,267 @@
...
+
+function readJsonOrFail(filePath, label) {
+  try {
+    return readTokenFile(filePath);
+  } catch (err) {
+    fail(`failed to parse ${label} ${filePath}: ${err.message}`);
+  }
+}
+
+function gitLastCommitTs(filePath) {
```
examlly-tech, line 163: **Minor — commit-timestamp gap:** Guard reflects committed state only. Uncommitted edits on either side bypass detection. Fine for CI; worth a one-liner in figma.md so local devs don't trust the guard for pre-commit verification.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:212` | CI hard-fail requires full git history; `actions/checkout@v4` defaults to shallow (`fetch-depth: 1`). Audited files outside the shallow window return null timestamps and spuriously trip the fail. Requirement undocumented. | Document `fetch-depth: 0` in figma.md, or detect shallow clones via `git rev-parse --is-shallow-repository` and emit a targeted message. |
| 2 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:212` | `process.env.CI` is a JS string: `'0'`, `'false'`, and any non-empty value evaluate truthy, so opt-out via `CI=0` still hard-fails. | Use an explicit allowlist (`['true', '1'].includes(process.env.CI?.toLowerCase())`) or exclude known opt-out values (`CI && CI !== '0' && CI !== 'false'`). |
| 3 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:160` | Stale-guard compares commit timestamps; uncommitted local edits on either side bypass detection silently. Acceptable for CI (sees committed state) but undocumented for local workflow. | Add a one-liner to figma.md noting the guard reflects committed state only. |

### Architectural Concerns

**Stale-guard couples script to CI checkout depth** (`packages/core/scripts/audit-figma-parity.js:160-216`) — The CI hard-fail's "CI must verify snapshot freshness" promise depends on full git history, but that dependency is silent. First contributor to wire `audit:figma-parity` into the `audit-tokens` job will hit spurious failures unless they set `fetch-depth: 0`; either document the requirement or detect shallow clones in `gitLastCommitTs` and emit a targeted message.

packages/core/scripts/__tests__/audit-figma-parity.test.js
```diff
@@ -0,0 +1,243 @@
...
+
+  it('skips `---foo` (triple-dash junk flag) without writing args["-foo"]', () => {
+    const args = parseArgs(['---foo', 'bar', '--category', 'color']);
+    expect(args.category).toBe('color');
+    expect(args['-foo']).toBeUndefined();
+  });
+
+  it('skips `--snapshot=` (empty value) so the default snapshot survives', () => {
+    const args = parseArgs(['--snapshot=', '--category', 'color']);
+    expect(args.snapshot).not.toBe('');
```
examlly-tech: **Minor — loose assertion:** `not.toBe('')` passes if `snapshot` is `null`, `undefined`, or any non-empty mutation. To enforce "default survives," export `DEFAULT_SNAPSHOT` and assert `toBe(DEFAULT_SNAPSHOT)`.

.claude/rules/figma.md
```diff
@@ -12,6 +12,91 @@
...
+```bash
+yarn audit:figma-parity --category color
+yarn audit:figma-parity --category color --snapshot path/to/alt-snapshot.json
+```
+
+Flags:
+
+| Flag                | Required?                              | Default                                    |
+| ------------------- | -------------------------------------- | ------------------------------------------ |
+| `--category <name>` | Always                                 | —                                          |
```
examlly-tech: **Minor — doc/code drift:** `--mode <name>` is documented as required for multi-mode categories, but no category accepts it today and `main()` doesn't read `args.mode`. Mark this row pending (`wired by #61`) to match the Categories table convention.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,267 @@
...
+    const arg = argv[i];
+    if (!arg.startsWith('--')) continue;
+    if (arg === '--') continue;
+    if (arg[2] === '-') continue;
+    const eq = arg.indexOf('=');
+    if (eq === 2) continue;
+    if (eq !== -1) {
+      const val = arg.slice(eq + 1);
+      if (val.length === 0) continue;
+      args[arg.slice(2, eq)] = val;
```
examlly-tech: **Minor — silent unknown flags:** `args[arg.slice(2,eq)] = val` accepts any `--foo` without a whitelist. A typo like `--catagory color` writes `args.catagory` and surfaces later as "--category required". Whitelist known keys (`category`, `snapshot`, `mode`, `theme`) and `fail()` on unknown.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/core/scripts/__tests__/audit-figma-parity.test.js:240` | `expect(args.snapshot).not.toBe('')` passes even if `snapshot` becomes `null`/`undefined`/any non-empty mutation — doesn't actually enforce "default survives". | Export `DEFAULT_SNAPSHOT` and assert `toBe(DEFAULT_SNAPSHOT)`, or at minimum `toMatch(/figma-snapshot\.json$/)`. |
| 2 | ⚠️ Minor | `.claude/rules/figma.md:36` | `--mode <name>` row in the Flags table is described as "Required for multi-mode categories", but no category accepts it today and `main()` doesn't read `args.mode`. Doc/code drift. | Annotate the row "pending — wired by #61" to match the Categories table convention, or remove until #61 wires it. |
| 3 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:40-62` | `parseArgs` accepts any `--anything value` with no whitelist. A typo like `--catagory color` writes `args.catagory` and surfaces later as the cryptic "--category required" error. | Whitelist known keys (`category`, `snapshot`, `mode`, `theme`) and `fail()` on unknown — small change, big debug-time win. |

### Architectural Concerns

**parseArgs guard chain readability** (`packages/core/scripts/audit-figma-parity.js:40-62`) — Five sequential `continue`s (`!startsWith('--')`, `=== '--'`, `arg[2] === '-'`, `eq === 2`, `val.length === 0`) are correct but obscure intent. A single regex like `^--([a-zA-Z][a-zA-Z0-9-]*)(=(.+))?$` capturing key+value would express "valid long flag" once and reject all five malformed shapes in one match.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,298 @@
...
+ * snapshot subtree read (`snapshot[cat][mode]` / `snapshot[cat][mode][theme]`).
+ * See .claude/rules/figma.md — Code-vs-Figma Parity Audit → Snapshot shape.
+ */
+const CATEGORIES = {
+  color: {
+    file: path.join(PRIMITIVES_DIR, 'color.json'),
+  },
+};
+
+const KNOWN_FLAGS = new Set(['category', 'snapshot', 'mode', 'theme']);
```
examlly-tech: **Discuss — accept-and-ignore flag:** `mode`/`theme` are whitelisted but `main()` never reads them, so `--mode bogus` silently exits 0. Drop them from `KNOWN_FLAGS` until #61 wires the multi-mode branch — the rejection path then surfaces `unknown flag: --mode` against the docs.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,298 @@
...
+  if (!snapshotSubtree) {
+    fail(
+      `snapshot has no "${args.category}" subtree. Refresh the snapshot or check the category name.`
+    );
+  }
+
+  const codeTs = gitLastCommitTs(config.file);
+  const snapshotTs = gitLastCommitTs(args.snapshot);
+  const staleGuardSkipped = codeTs === null || snapshotTs === null;
+  if (staleGuardSkipped && isCI()) {
```
examlly-tech, line 240: **Blocking — CI freshness promise unmet:** The `isCI()` hard-fail and `isShallowClone()` targeted error are dead code — no CI job invokes `audit:figma-parity`. Either wire `audit-tokens` to call it (with `fetch-depth: 0`), or remove the CI-specific logic. Today the documented `figma.md:50` contract is aspirational.

examlly-tech (changes requested):

### Issues

| 1 | ❌ Blocking | `.github/workflows/ci.yml:254-284` | `audit:figma-parity` is not invoked in CI. The `audit-tokens` job runs `audit:contrast` and `audit:class-refs` only, so the elaborate `isCI()` / `isShallowClone()` hard-fail branch added in this PR is dead code. The figma.md "CI must verify snapshot freshness" promise is unmet end-to-end. | Add a `yarn workspace @nexus/core audit:figma-parity --category color` step to the `audit-tokens` job. Set `fetch-depth: 0` on its checkout (matching the pattern at lines 61/99/149/200) so the stale-snapshot guard works. |
| 2 | ⚠️ Discuss | `packages/core/scripts/audit-figma-parity.js:32` | `mode`/`theme` are in `KNOWN_FLAGS` but `main()` never reads them — `--category color --mode bogus` exits 0 silently with no signal `--mode` was ignored. Inverts the safety the whitelist exists to provide. | Drop `mode`/`theme` from `KNOWN_FLAGS` until #61 wires the multi-mode branch (the current rejection path would then say "unknown flag: --mode"). Re-add when the branch lands. |

### Architectural Concerns

**CI integration gap** (`audit-figma-parity.js:236-247`) — Elaborate CI detection is added (`isCI`, `isShallowClone`, targeted `fetch-depth: 0` error), but no CI job invokes the script — the entire branch is unreachable in practice. Wire the audit into `audit-tokens` in this PR so the freshness guard is actually live.

**Forward-flag whitelist** (`audit-figma-parity.js:32`) — Whitelisting flags that aren't implemented inverts the safety the whitelist exists to provide; today `--mode bogus` silently exits 0 with "no drift." Either implement-or-reject; don't accept-and-ignore.

packages/core/scripts/audit-figma-parity.js
```diff
@@ -0,0 +1,298 @@
...
+
+function main() {
+  const args = parseArgs(process.argv.slice(2));
+
+  const unknownFlags = Object.keys(args).filter((k) => !KNOWN_FLAGS.has(k));
+  if (unknownFlags.length > 0) {
+    fail(
+      `unknown flag(s): ${unknownFlags.map((k) => `--${k}`).join(', ')}\n  known: ${[...KNOWN_FLAGS].map((k) => `--${k}`).join(', ')}\n  ${DOCS_HINT}`
+    );
+  }
```
examlly-tech, line 207: **Minor — untested validation path:** the round-5 `unknownFlags` rejection lives inside `main()`, which isn't exported, so the typo-guard has no unit-test coverage. A future regression that drops the check slips through silently. Export `KNOWN_FLAGS` (or a thin `findUnknownFlags(args)`) and add a test asserting `--catagory color` is surfaced.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/core/scripts/audit-figma-parity.js:198-203` | The new unknown-flag rejection lives in non-exported `main()`, so the round-5 typo-guard has no unit-test coverage. A future regression that drops the check slips through silently. | Export `KNOWN_FLAGS` (or a thin `findUnknownFlags(args)` helper); add a test asserting `--catagory color` surfaces in the unknown-flag list. |

examlly-tech (approved):

### Unresolved Issues

| 1 | `audit:figma-parity` not invoked in CI; freshness promise unmet | `.github/workflows/ci.yml:288-289` | ✅ Fixed | Step added; `fetch-depth: 0` set; `react` filter covers `packages/core/**` |
| 2 | `--mode`/`--theme` in whitelist but unread; silent exit 0 | `packages/core/scripts/audit-figma-parity.js:32` | ✅ Fixed | Whitelist trimmed to `{category, snapshot}`; regression test locks `--mode` rejection |

### New Issues

None.

examlly-tech (approved):

### Unresolved Issues

| 1 | Unknown-flag rejection in non-exported `main()`; no unit-test coverage | `packages/core/scripts/audit-figma-parity.js:198-203` | ✅ Fixed | `findUnknownFlags(args)` extracted, exported; 4 new tests lock `--catagory`/`--mode`/multi-key |

### New Issues

None.

PR #143, opened 2026-05-24
[5/6] chore(stories): script per-base Storybook variant generation

### LINKED ISSUE #89: [5/6] chore(stories): script per-base Storybook variant generation
## Summary

Write a Node script that generates per-base Storybook variant stories from a template. Currently each component has one set of stories rendered under the default base (slate). We want each component to also have stories rendered under all 5 bases (slate, neutral, gray, stone, zinc) × 2 themes (light, dark) = 10 visual permutations per component. With ~13 components, that's 130 visual checkpoints — _generated from one template per component_, not hand-written.

## Research basis

Nexus's claim to fame on the matrix is perceptual consistency across multiple bases (the `perceptual-grid` work). That claim is currently _unverifiable_ by anyone who isn't running the codebase locally — there's no place to see the same component rendered across 5 bases side by side.

The previous-session phase plan called for this work but estimated 65 hand-written stories. Scripting collapses that to one template per component.

## Current state

- Component stories live in `packages/react/src/components/ui/{component}.stories.tsx`
- All stories render under the default base (slate, per `packages/tailwind/nexus.css`)
- No mechanism to render the same component under a non-default base
- The 5 base CSS variants exist in `dist/` after `yarn tokens:modular` runs

## Proposed change

1. Add a `scripts/generate-base-variants.mjs` to `packages/react/`
2. For each component story file, generate a companion `{component}.base-variants.stories.tsx` that:
   - Imports the component
   - Imports each base's CSS variant as a side-effect import
   - Renders a key story (e.g., `AllVariants`) under each base × theme combination
3. Generated files live in `packages/react/src/components/__generated__/` (gitignored — regenerated on each Storybook build)
4. Wire into the Storybook startup so `yarn storybook` includes the generated variants automatically

## Implementation order

1. Survey existing component stories to identify the canonical "showcase" story per component (usually `AllVariants`)
2. Define the generator template
3. Write the generator script — reads a config (`scripts/base-variants.config.json`) listing components + showcase stories
4. Wire into `packages/react/package.json` as `prestorybook` (or similar)
5. Verify Storybook renders all variants correctly under both themes
6. Add the generator behaviour to `.claude/rules/storybook.md`

## Acceptance criteria

- [ ] Generator script exists at `packages/react/scripts/generate-base-variants.mjs`
- [ ] Config file lists every interactive component + its showcase story
- [ ] Running `yarn storybook` produces base-variant stories for every listed component
- [ ] Chromatic captures all variants (visual regression coverage for cross-base parity)
- [ ] `.claude/rules/storybook.md` documents the generator + how to opt a new component in

## Open questions

1. **Generated files in git or gitignored** — gitignored keeps the repo clean but breaks CI Chromatic if it expects committed stories. Default: gitignored + Chromatic configured to run after generation.
2. **Per-theme rendering inside a single story** — render light + dark side-by-side in one story, or have separate `*-light` and `*-dark` story variants? Side-by-side is denser; separate matches existing convention.

## References

- Research: [Nexus matrix](../reports/nexus-ds-intelligence.html#sec-tokens-color) — perceptual-grid is the lead claim that needs demonstrable proof
- Related: #6 (Figma library mirror) depends on this issue's variant naming convention
- Related rule: [`.claude/rules/storybook.md`](../.claude/rules/storybook.md)

Milestone: [Color foundation · Phase 2](https://github.com/nexuslabs-ai/nexus/milestone/3)

aishvaryagupta:
## Summary

Scripts per-base Storybook variant generation so Nexus's cross-base perceptual consistency is visible in one place — each component rendered across all 5 bases × 2 themes, generated from one template per component.

- `packages/react/scripts/generate-base-variants.mjs` reads `base-variants.config.json` and emits, into the gitignored `src/components/__generated__/`:
  - `base-variants.css` — each base's semantic tokens scoped to `[data-nexus-base]` (light) / `[data-nexus-base].dark` (dark), `--nx-color-*` prefixed, merging base + the `neutral` brand so cells are theme-hermetic
  - one `{Name}.base-variants.stories.tsx` per component — a 2×5 grid (light/dark × 5 bases) that reuses the component's existing showcase
- Config covers all 13 components (12× `AllVariants` + Avatar's `AllSizes`)
- Generation chained ahead of `storybook`/`build-storybook` and `test`/`test:storybook`
- Documented in `.claude/rules/storybook.md`

### Design notes
- **Token source is the committed semantic JSON** (not `dist/modular`), so there is no build-order dependency; output verified byte-identical to `dist/modular/base-slate.css` for base tokens.
- Utilities read **prefixed** vars (`nx:bg-background` → `var(--nx-color-background)` via `prefix(nx)`), so the generator emits `--nx-color-*` scoped to `[data-nexus-base]`.
- The grid wrapper uses **inline styles** (not `nx:` utilities) so no new Tailwind classes need scanning of the gitignored dir; the showcase brings its own (scanned) classes.
- **Component decorators** (e.g. Tooltip's `TooltipProvider`) are applied; the global preview decorator is skipped so each cell stays self-contained.
- Per decision, **`@storybook/addon-vitest`** is the visual-regression capture mechanism (no Chromatic SaaS).
- Overlay open/portal content and `chart-categorical` tokens are intentionally not per-cell-scoped (documented in the rule).

## GitHub Issue

Closes #89

## Test Plan

- [x] `yarn lint` (whole repo, `--max-warnings 0`) — clean
- [x] `yarn build-storybook` — all 13 generated stories compile; scoped `[data-nexus-base]` CSS ships in the bundle
- [x] `yarn test:storybook` — **230/230** story tests pass (incl. all 13 base-variant stories)
- [x] Generated CSS verified byte-identical to `dist/modular/base-slate.css` for base tokens (77/77 light + dark)


packages/react/scripts/base-variants.config.json
```diff
@@ -0,0 +1,72 @@
+{
+  "bases": ["slate", "neutral", "gray", "stone", "zinc"],
+  "brand": "neutral",
+  "themes": ["light", "dark"],
+  "components": [
+    {
+      "name": "Accordion",
+      "storiesFile": "Accordion.stories",
```
examlly-tech: **Minor — redundant field:** `storiesFile` is `{name}.stories` for all 13 entries — a second source of truth that can drift from `name`. Derive it from `name` in the generator and drop the key (`code-quality.md`: one value beats one config key).

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,244 @@
...
+
+function buildStory(component, config) {
+  const bases = JSON.stringify(config.bases);
+  const themes = JSON.stringify(config.themes);
+  return `/* AUTO-GENERATED by scripts/generate-base-variants.mjs — DO NOT EDIT. */
+/* Source: ../ui/${component.storiesFile} (${component.showcase}). */
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import type { ReactNode } from 'react';
+
+import componentMeta, { ${component.showcase} as showcase } from '../ui/${component.storiesFile}';
```
examlly-tech: **Minor — no existence check:** this templates an import from `../ui/${storiesFile}` without verifying the file or `${showcase}` export exists. A typo produces a valid `.tsx` that fails only at the Storybook/vitest build, not at generation. Add an `existsSync` guard mirroring `readJson`, failing fast with the component name.

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,244 @@
...
+  }
+  return JSON.parse(fs.readFileSync(filePath, 'utf8'));
+}
+
+/**
+ * Convert a DTCG reference (`{slate.200}`, `{white}`, `{slate.a200}`) to the
+ * prefixed primitive CSS variable the Tailwind utilities consume.
+ * Dots become dashes; everything else is preserved.
+ */
+function refToVar(value) {
```
examlly-tech, line 66: **Minor — unvalidated ref→var:** any `{…}` becomes `var(--nx-color-…)` with no check against real primitives, so a removed/typo'd ref ships a broken `var()` silently. Core's resolver warns and falls back. Validate against the primitive set or reuse core's resolver.

examlly-tech (commented):

### Issues

| #   | Severity | File                             | Issue                                                                                                                                                                                                                 | Suggestion                                                                                        |
| --- | -------- | -------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
| 1   | ⚠️ Minor | `package.json:25`                | `typecheck` (`turbo typecheck` → react `tsc` over `src/**`, incl. `__generated__`) never regenerates; the files are gitignored → absent on a fresh CI checkout, so generated stories are silently untypechecked there | Regenerate before `typecheck`, or exclude `__generated__` from tsc so the gap is explicit         |
| 2   | ⚠️ Minor | `base-variants.config.json:8`    | `storiesFile` is always `{name}.stories` (all 13 entries) — a redundant field that can drift from `name`                                                                                                              | Derive `storiesFile` from `name` in the generator; drop the config key                            |
| 3   | ⚠️ Minor | `generate-base-variants.mjs:108` | Generator templates `import … from '../ui/{storiesFile}'` without checking the file/export exists; a typo yields a valid `.tsx` that fails only at the Storybook/vitest build                                         | `existsSync` the stories file (mirror the `readJson` guard) and fail fast with the component name |
| 4   | ⚠️ Minor | `generate-base-variants.mjs:33`  | `refToVar` emits `var(--nx-color-…)` for any `{…}` with no primitive-set validation; a removed/typo'd ref silently ships a broken `var()` (core's resolver warns + falls back)                                        | Validate against the primitive set, or document the divergence from core's resolver               |

### Architectural Concerns

**Re-implemented ref→var resolver vs reusing core's** (`generate-base-variants.mjs:33`) — Output is verified byte-identical to `dist/modular/base-slate.css` today, but the generator hardcodes the dot→dash transform that `@nexus/core` derives from a primitive map, so the two can drift if a primitive's cssName ever stops being a pure path-join. Reusing core's resolver would collapse the duplication and inherit its warn-on-miss safety.

**CI typecheck coverage is zero for the generated `src/**`** — The `typecheck`job runs`turbo typecheck`(dependsOn`^build`only, no regeneration), and the generated`.tsx`is gitignored, so it is absent on a fresh CI checkout. Generated-story type errors surface only in`test-react`/`build-storybook`, not at the `typecheck` gate. Regenerate in that job, or make the exclusion explicit so the gap isn't mistaken for coverage.

---

_Review performed against `.claude/rules/` guidelines_

examlly-tech (commented):

### Issues

| #   | Severity | File                             | Issue                                                                                                                                                    | Suggestion                                                                                                  |
| --- | -------- | -------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------- |
| 1   | ⚠️ Minor | `package.json:15,18`             | `test:watch` / `test:storybook:watch` skip `generate:base-variants`; on a fresh checkout (output gitignored) watch mode starts with no generated stories | Prefix both with `yarn workspace @nexus/react generate:base-variants &&` to match `test` / `test:storybook` |
| 2   | ⚠️ Minor | `generate-base-variants.mjs:126` | `args` merge + `context.args` are dead plumbing — every configured showcase renders `(_args)` and neither decorator reads context                        | Keep as forward-compat, or trim to a bare `{}` context; flagging the unused wiring                          |
| 3   | ⚠️ Minor | `generate-base-variants.mjs:75`  | `Map` dedup is unnecessary — base/brand keys are fully disjoint (0 overlap)                                                                              | Harmless; a plain spread reads simpler, but `Map` is fine as collision insurance                            |

### Correctness Concern

**Static ids in reused showcases collide across cells** (`generate-base-variants.mjs:168`) — Switch's `AllVariants` embeds fixed `id="setting1"` etc.; rendering it 10× emits duplicate DOM ids, breaking each cell's `htmlFor`↔control pairing. Currently masked by `a11y: { test: 'off' }` and smoke-only assertions. The generator can't scope ids it doesn't own — document this as a showcase-authoring constraint in `storybook.md § Caveats` (alongside the portal caveat), or namespace the subtree.

---

_Review performed against `.claude/rules/` guidelines_

package.json
```diff
@@ -11,10 +11,10 @@
     "build": "turbo run build",
     "dev": "turbo run dev",
     "lint": "eslint . --max-warnings 0",
-    "test": "vitest run",
+    "test": "yarn workspace @nexus/react generate:base-variants && vitest run",
     "test:watch": "vitest",
```
examlly-tech: **Minor — lifecycle gap:** `test:watch` and `test:storybook:watch` don't run `generate:base-variants`, unlike `test` / `test:storybook`. On a fresh checkout (output gitignored) watch mode starts with no generated stories. Prefix both with the generate step.

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,244 @@
...
+
+function toDecorators(d: unknown): AnyDecorator[] {
+  if (!d) return [];
+  return (Array.isArray(d) ? d : [d]) as AnyDecorator[];
+}
+
+function renderShowcase(): ReactNode {
+  const render = showcase.render as unknown as AnyRender | undefined;
+  if (!render) return null;
+  const args = { ...componentMeta.args, ...showcase.args };
```
examlly-tech: **Minor — dead plumbing:** every configured showcase renders `(_args)` and neither Tooltip nor Input decorators read `context`, so the `args` merge and `context.args` never feed anything. Fine as forward-compat, but currently unused — worth a note or trimming to a bare `{}`.

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,244 @@
...
+          fontSize: '0.6875rem',
+          fontWeight: 600,
+          textTransform: 'uppercase',
+          letterSpacing: '0.04em',
+          color: 'var(--nx-color-muted-foreground)',
+        }}
+      >
+        {base}
+      </span>
+      {renderShowcase()}
```
examlly-tech, line 206: **Minor — duplicate ids across cells:** `renderShowcase()` reruns showcases that embed static ids (Switch `AllVariants` has `id="setting1"` etc.) once per cell → 10 duplicates, broken `htmlFor` pairing. Masked by `a11y: test off`. Document as a showcase constraint in `storybook.md § Caveats`.

.claude/rules/storybook.md
```diff
@@ -336,3 +336,62 @@ yarn test:storybook:ui     # Interactive UI
...
+### What it produces
+
+`packages/react/scripts/generate-base-variants.mjs` reads `base-variants.config.json` and writes to `packages/react/src/components/__generated__/` (gitignored, regenerated on every run):
+
+| Output                             | Contents                                                                                                     |
+| ---------------------------------- | ------------------------------------------------------------------------------------------------------------ |
+| `base-variants.css`                | Each base's semantic tokens scoped to `[data-nexus-base="…"]` (light) / `[data-nexus-base="…"].dark` (dark)  |
+| `{Name}.base-variants.stories.tsx` | One per component — a 2×5 grid (light/dark rows × 5 base columns), under the `Base Variants/*` sidebar group |
+
+The stories run as smoke/render tests under `@storybook/addon-vitest` — that is the visual-regression capture mechanism (no Chromatic).
```
examlly-tech: **Minor — overstated claim:** addon-vitest smoke tests assert the grid renders without error across 5 bases × 2 themes; they don't capture pixels, so they can't gate a perceptual-consistency regression — the thing this generator exists to demonstrate. Reword to "render/smoke coverage".

examlly-tech (commented):
## Principal Architect Review — Verdict: APPROVED

### Issues

| 1 | ⚠️ Minor | `.claude/rules/storybook.md:353` | "that is the visual-regression capture mechanism" overstates the addon-vitest smoke tests — they assert the matrix renders without throwing, not pixel output, so they cannot gate a perceptual-consistency regression (the thing this generator exists to demonstrate). | Reword to "render/smoke coverage". Pixel-diffing is a deliberate non-goal of this PR (smoke tests are the chosen mechanism), so only the wording needs correcting. |

### Architectural Concerns

**Generation wired below the turbo layer** — the gitignored `__generated__/` artifact sits outside turbo's input/output model, so `typecheck`'s cache key doesn't include it; a token-source change that alters generated output is caught only via the `^build` dependency, not the typecheck hash. Pre-existing pattern, sound while every consumer regenerates first — revisit only if a future task consumes the output without regenerating.

---

_Review performed against `.claude/rules/` guidelines_

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,289 @@
...
+const PRIMITIVE_NAMES = collectPrimitiveNames(readJson(PRIMITIVES_PATH));
+
+/**
+ * Convert a DTCG reference (`{slate.200}`, `{white}`, `{slate.a200}`) to the
+ * prefixed primitive CSS variable the Tailwind utilities consume.
+ * Dots become dashes; everything else is preserved.
+ */
+function refToVar(value) {
+  if (typeof value !== 'string' || !value.startsWith('{') || !value.endsWith('}')) {
+    console.warn(`  ⚠ non-reference semantic value passed through verbatim: ${value}`);
```
examlly-tech: **Minor — inconsistent + dead branch:** this non-reference path warns and returns the raw value (emitting a broken `--nx-color-*` decl), while the branch right below throws for an unresolvable ref. The sole caller only ever passes `{ref}` strings, so it never runs — drop it; let non-refs hit the throw.

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,289 @@
...
+ * Fail fast (before templating) if the canonical stories file or its named
+ * showcase export is missing — a typo otherwise yields a valid `.tsx` that
+ * only breaks at the Storybook/vitest build.
+ */
+function assertShowcaseExists(component) {
+  const file = path.join(UI_DIR, `${component.name}.stories.tsx`);
+  if (!fs.existsSync(file)) {
+    throw new Error(`Stories file missing for "${component.name}": ${file}`);
+  }
+  if (!new RegExp(`export const ${component.showcase}\\b`).test(fs.readFileSync(file, 'utf8'))) {
```
examlly-tech: **Minor — unescaped RegExp from config:** `new RegExp(`export const ${showcase}\b`)` interpolates the config value unescaped. Config-controlled, so low risk. If simplifying, escape it rather than dropping to `.includes` — the `\b` prevents a false match on `export const AllVariantsFoo`.

examlly-tech (commented):

### Issues

| 1 | ⚠️ Minor | `packages/react/scripts/generate-base-variants.mjs:61` | `refToVar`'s non-reference branch warns and returns the raw value (which would emit a broken `--nx-color-*` declaration), while the branch just below throws for an unresolvable ref — two fail modes for the same class of bad input. The sole caller (`flattenColorTokens`) only ever passes `{ref}` strings, so the warn branch is also unreachable. | Drop the warn-passthrough; let a non-`{…}` value fall through to the throw (fail fast). |
| 2 | ⚠️ Minor | `packages/react/scripts/generate-base-variants.mjs:258` | `assertShowcaseExists` interpolates the config `showcase` into a `RegExp` unescaped. Config-controlled, so no real risk today. | Low-priority nit. If touched, escape the value rather than switching to `.includes` — the `\b` guards against a false match on `export const AllVariantsFoo`. |

### Architectural Concerns

None.

---

_Review performed against `.claude/rules/` guidelines_

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+const PRIMITIVE_NAMES = collectPrimitiveNames(readJson(PRIMITIVES_PATH));
+
+/**
+ * Convert a DTCG reference (`{slate.200}`, `{white}`, `{slate.a200}`) to the
+ * prefixed primitive CSS variable the Tailwind utilities consume.
+ * Dots become dashes; everything else is preserved.
+ */
+function refToVar(value) {
```
examlly-tech, line 66: **Architectural — second copy of the token resolver.** `refToVar`/`flattenColorTokens` duplicate logic exported from `packages/core/scripts/utils.js`, but that file isn't in `@nexus/core`'s `exports` nor built to `dist/`, so the package API can't supply it today. Either export it from `@nexus/core` and import, or anchor this copy to the canonical functions in a comment.

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,289 @@
...
+ * Recursively flatten a DTCG token tree into [{ cssVar, value }].
+ * Nested groups join with `-`; `$`-prefixed metadata keys are skipped.
+ */
+function flattenColorTokens(node, prefix = []) {
+  const out = [];
+  for (const [key, val] of Object.entries(node)) {
+    if (key.startsWith('$') || val === null || typeof val !== 'object') {
+      continue;
+    }
+    if (val.$value !== undefined) {
```
examlly-tech: **Minor — diverges from the canonical flatten.** Core's `extractTokens` requires `$value` and `$type` and derives the `nx-color` prefix from `$type`; this checks `$value` only and hardcodes `--nx-color-`. Correct only because every semantic token is currently `color` — a future non-color semantic alias would mis-prefix silently. Gate on `$type`.

examlly-tech (commented):
## 🏛️ Principal Architect Review — #143

**Verdict: NEEDS DISCUSSION** — no blocking defects; two latent drift risks and one shipped constraint violation worth resolving before merge.

### Issues

| 1 | ⚠️ Minor | `Switch.stories.tsx` (`AllVariants`) | Embeds static ids (`setting1`, `label-right`, …) yet Switch is opted into base-variant generation — violating the constraint this PR documents. Across the 10 cells these become duplicate DOM ids, masked only by `a11y:{test:'off'}`. | Swap the static ids for `React.useId()` / `aria-label` in `AllVariants`. |
| 2 | ⚠️ Minor | `generate-base-variants.mjs:83` | `flattenColorTokens` gates on `$value` only and hardcodes `--nx-color-`; the canonical `extractTokens`/`processSemanticTokens` gate on `$value` AND `$type` and derive the prefix from `$type`. Correct only while every semantic token is `color`. | Mirror canonical: skip nodes without `$type`, derive the prefix from it. |

### Architectural Concerns

**Token resolver duplicated outside the canonical pipeline.** `refToVar` / `collectPrimitiveNames` / `flattenColorTokens` re-implement logic exported from `packages/core/scripts/utils.js`, but that file is not in `@nexus/core`'s `exports` and isn't built to `dist/`, so it cannot be imported through the package API today. Either export the resolver from `@nexus/core` and import it (one source of truth), or keep the copy and anchor it to the canonical functions in a comment.

**`chart-categorical` scope gap is acceptable but undeclared in config.** The 5 chart tokens are silently absent from per-cell CSS — fine while no charted component is opted in, but `base-variants.config.json` gives no signal that opting one in would render it under the wrong base. A `"//"` note in the config (not just `storybook.md`) keeps the constraint next to the data.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Review performed against `.claude/rules/` guidelines_

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,289 @@
...
+export const AllBases: Story = {};
+`;
+}
+
+/**
+ * Fail fast (before templating) if the canonical stories file or its named
+ * showcase export is missing — a typo otherwise yields a valid `.tsx` that
+ * only breaks at the Storybook/vitest build.
+ */
+function assertShowcaseExists(component) {
```
examlly-tech, line 270: **Minor — guard doesn't match the doc.** This checks the export name only. An args-only showcase passes here, then `renderShowcase()` returns `null` → silently empty cells, not fail-fast. The doc says a showcase 'must not be args-only' and that generation 'fails fast'. Assert the matched export block also contains `render:`, or soften the doc.

packages/react/scripts/generate-base-variants.mjs
examlly-tech: **Minor — regex false positive.** A name-only match can hit a commented or stringified occurrence of the showcase name before the real export. Anchor it to the start of a line (add a `^` prefix with the multiline flag) so only a top-level `export const` satisfies the guard.

packages/react/scripts/generate-base-variants.mjs
examlly-tech: **Minor — silent pass-through.** A non-reference value warns and passes through verbatim, emitting `--nx-color-x: <raw>` (invalid CSS). All semantic values are refs today, so this branch is unreachable, but it diverges from `flattenColorTokens` which requires refs. Throw here, like the unresolvable-ref branch just below.

examlly-tech (commented):
## 🔍 SDE2 Review — #143

**Verdict: MINOR CHANGES** — core generation paths verified correct (typecheck passes; `reduceRight` decorator nesting matches Storybook's `StoryFn` contract; alpha/`{white-a900}` leaves emit well-formed CSS; error paths exit non-zero). All findings below are doc-vs-behaviour gaps or a shipped showcase violating the PR's own authoring constraint — each fixable in this PR per `no-follow-up-deferral.md`.

### Issues

| 1 | ⚠️ Minor | `Switch.stories.tsx` (`AllVariants`) | Opted into base-variant generation but embeds static ids (`setting1`, `label-right`, `label-left`); 10 cells reuse its `render()` → duplicate DOM ids, breaking `htmlFor`↔control pairing. Exactly the case the PR's § Caveats forbids. | Replace the `id=`/`htmlFor=` pairing with `React.useId` or `aria-label`. |
| 2 | ⚠️ Minor | `generate-base-variants.mjs:253` | `assertShowcaseExists` checks only the export name. An args-only showcase passes, then `renderShowcase()` returns `null` → silently empty cells, not fail-fast. The doc says showcases 'must not be args-only' and 'fails fast'. | Assert the matched export block also contains `render:`, or soften the doc claim. |
| 3 | ⚠️ Minor | `generate-base-variants.mjs:258` | Name-only regex matches a commented or stringified occurrence before the real export — a false positive passes the guard. | Anchor to the start of a line (multiline-flag regex) so only a top-level `export const` passes. |
| 4 | ⚠️ Minor | `generate-base-variants.mjs:61` | `refToVar` warns and passes a non-reference value through verbatim while `flattenColorTokens` requires refs — emits invalid CSS (`--nx-color-x: <raw>`) if ever hit. Unreachable today, diverges from the sibling branch. | Throw, matching the unresolvable-ref branch directly below. |

### Architectural Concerns
None — architecture is covered in the Principal Architect review.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Review performed against `.claude/rules/` guidelines_

examlly-tech (commented):
## 🏛️ Principal Architect Follow-up — #143

**Verdict: APPROVED** — all previous architectural issues resolved; no new issues.

Verified against commit `94b305b`:

- **Switch static-ids → `React.useId()`:** Confirmed unique ids per cell under the generator's `render({}, {})` direct-call pattern (`_r_0_`…`_r_4_`); the hook attaches to each distinct `BaseCell` fiber. All 13 base-variant stories + `typecheck` pass.
- **`$type === 'color'` gate:** Safe — every `base-*` / `brands-*` semantic file is 100% color tokens (950/950), so it drops nothing today and only guards a future non-color alias from mis-prefixing. Now mirrors core's `extractTokens` / `processSemanticTokens` split.
- **Resolver duplication:** The provenance comment's "byte-identical to dist/modular" claim verified for slate-light and zinc-dark. Copy-not-import is the correct call — `core/scripts` isn't in `@nexus/core`'s exports, and the `$type` gate makes the copy *more* faithful, not less.
- **`chart-categorical` scope gap:** The `"//"` config note is inert (generator reads only `bases`/`themes`/`brand`/`components`) and is the right JSON-comment mechanism.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Follow-up review performed against `.claude/rules/` guidelines_

examlly-tech (commented):
## 🔍 SDE2 Follow-up — #143

**Verdict: MINOR CHANGES** — all four previous issues resolved; one contrived, non-blocking new edge case.

### Unresolved Issues

- `refToVar` non-reference branch now throws (`generate-base-variants.mjs:66`), matching the sibling unresolvable-ref throw; the message carries the offending value.
- `assertShowcaseExists` escapes + anchors the export regex (`:278`) and requires `render:` within the export block (`:291`). Verified against all 13 showcases and the args-only / same-line-render / prefix-collision / block-bleed edge cases.
- Switch `AllVariants` uses `React.useId()` (`Switch.stories.tsx:229`); 14 story tests pass, each cell is a distinct fiber so ids no longer collide, and there are no React hook warnings.

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `generate-base-variants.mjs:291` | `render:` inside a string literal in an args-only block passes the guard. Contrived — no real showcase triggers it. |

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Follow-up review performed against `.claude/rules/` guidelines_

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,323 @@
...
+    throw new Error(
+      `Showcase export "${component.showcase}" not found in ${component.name}.stories.tsx`
+    );
+  }
+  // Require a `render:` within the export's block (up to the next top-level
+  // export) so an args-only story fails here rather than rendering empty cells.
+  const rest = src.slice(exportMatch.index);
+  const nextExport = rest.slice(exportMatch[0].length).search(/^export /m);
+  const block = nextExport === -1 ? rest : rest.slice(0, exportMatch[0].length + nextExport);
+  if (!/\brender\s*:/.test(block)) {
```
examlly-tech: **Minor — string-literal false positive:** `/\brender\s*:/` also matches `render:` inside a string in an args-only block (`args:{label:'render: x'}`) → could slip through to empty cells. No real showcase hits this. A `^\s*render`-anchored fix would break single-line `{ render: … }` exports — leave as-is unless hardened to strip strings.

examlly-tech (commented):
## 🏛️ Principal Architect Follow-up — #143

**Verdict: APPROVED** — all previous issues resolved; no new issues.

### Unresolved Issues

Commit `4039bd1` is a localized regex hardening inside `assertShowcaseExists`; it touches no module boundary, import, token-pipeline code, or generation output. The static-source guard remains the right design — importing the `.tsx` to inspect `showcase.render` would pull a React/Storybook runtime into a pre-build script and invert the dependency direction. The `storybook.md` "render-based / fails fast" contract is unchanged. Prior APPROVED stands.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Follow-up review performed against `.claude/rules/` guidelines_

packages/react/scripts/generate-base-variants.mjs
```diff
@@ -0,0 +1,327 @@
...
+  if (!exportMatch) {
+    throw new Error(
+      `Showcase export "${component.showcase}" not found in ${component.name}.stories.tsx`
+    );
+  }
+  // Require a `render:` object key within the export's block (up to the next
+  // top-level export) so an args-only story fails here rather than rendering
+  // empty cells. Match `render:` either at a line start (the indented-key case,
+  // robust to intervening comments/keys) or right after a `{`/`,` (same-line
+  // case) — so a `render:` inside a string value (mid-line, after a quote)
```
examlly-tech: **Minor — comment completeness:** the comment correctly scopes the string exclusion to a `render:` that's `(mid-line, after a quote)`. The residual gap it doesn't mention: a line-leading `render:` inside a multiline template literal still matches `^…/m`. Contrived and smoke-test-caught — worth a half-line note, or leave as-is.

examlly-tech (commented):
## 🔍 SDE2 Follow-up — #143

**Verdict: MINOR CHANGES** — previous finding resolved; one comment-completeness nit.

### Unresolved Issues

_Previous finding resolved._ `/(?:^|[{,])\s*render\s*:/m` rejects the flagged mid-line string `render:` (`args: { label: 'render: x' }` → no match) while accepting every real showcase, including Switch's `AllVariants` (comment lines between `{` and the key — the `^` line-start alternative is what makes this pass; the intermediate `/[{,]\s*render\s*:/` false-negatived it). All 13 showcases generate; Switch base-variant + canonical tests pass.

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `generate-base-variants.mjs:290` | Comment scopes the string exclusion to "(mid-line, after a quote)" — accurate, but doesn't note that a line-leading `render:` inside a multiline template literal still satisfies `^…/m` (contrived; smoke-test-caught). |

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Follow-up review performed against `.claude/rules/` guidelines_

examlly-tech (commented):
## 🏛️ Principal Architect Review — #143

**Verdict: APPROVED** — no issues in my lane.

Module boundaries, the config contract, and the showcase `render()` reuse are sound; every entry point that consumes `__generated__/` regenerates first (the turbo `typecheck` cache-hit is harmless — generated output is a pure function of hashed tracked inputs, and `tsc` is skipped on the same hit). Per-cell emission of the full base + `neutral`-brand set (disjoint sets) wins over an ancestor `.dark` by CSS inheritance, so cells are hermetic; the resolver copy resolves identically to `dist/modular`.

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Review performed against `.claude/rules/` guidelines_

examlly-tech (approved):
## 🔍 SDE2 Review — #143

**Verdict: APPROVED** — no issues in my lane.

Generator runs clean across all 13 showcases; `typecheck` passes; `test:storybook` green (230/230, incl. all 13 generated stories and decorator-wrapped Tooltip/Input); lint clean (0 warnings). `assertShowcaseExists` render-guard verified against edge cases (args-only rejected, later-export `render:` leak blocked, last-export handled, `render:`-in-string rejected); `refToVar` / `flattenColorTokens` resolve all `{ref}` forms incl. alpha/top-level primitives and skip non-color leaves; Switch `useId` stable (one hook per cell); build wiring consistent (`test:unit` correctly excluded).

### Links
- **PR:** https://github.com/nexuslabs-ai/nexus/pull/143
- **GitHub Issue:** #89

---
_Review performed against `.claude/rules/` guidelines_

PR #179, opened 2026-05-25
[10/10] feat(responsive): Show/Hide primitives (experimental)

### LINKED ISSUE #103: [10/10] feat(responsive): Show/Hide primitives (experimental)
## Summary

Ship two small declarative responsive-visibility primitives — `<Show>` and `<Hide>` — in a new `packages/react/src/components/primitives/` folder. Atlassian-inspired but Nexus-extended: where Atlassian's `<Show above="md">` only handles viewport, Nexus's adds `containerAbove` / `containerBelow` axes for container queries. The API surfaces the container-over-viewport stance at the moment of use, not just in docs.

## Research basis

- Phase 2 §5 Decision E: [`reports/phase-2-token-decisions.html § 5`](../reports/phase-2-token-decisions.html#bp-decisions)
- Atlassian DS `@atlaskit/primitives` `<Show>` / `<Hide>` components (source: `dist/types/responsive/show.d.ts`) — declarative wrappers using `display: none` / `display: revert`. Children always render; only visibility toggles.

Today, hiding a sidebar below `lg` in Nexus consumer code requires a two-class invert (`nx:hidden nx:lg:block`) — small but real footgun. Declarative primitives are clearer:

```tsx
<Show above="lg"><aside>Secondary nav</aside></Show>
<Hide below="md"><span>Extra context only on tablet+</span></Hide>
```

The Nexus extension adds container-query axes, surfacing the stance from Decision C:

```tsx
<Show containerAbove="md">
  <span>Visible when parent container ≥ md, regardless of viewport</span>
</Show>
```

## Scope

New folder: `packages/react/src/components/primitives/`. Five files:

| File | What |
| --- | --- |
| `primitives/show.tsx` | ~30 lines — CSS-driven wrapper using `display: none` / `display: revert`. Children render regardless; only visibility toggles. Supports `above` / `below` / `containerAbove` / `containerBelow`. |
| `primitives/hide.tsx` | Inverse semantics of `show.tsx`; otherwise same shape and API. |
| `primitives/Show.stories.tsx` | Storybook stories covering each prop axis with play tests verifying visibility at different viewport / container widths. |
| `primitives/Hide.stories.tsx` | Same coverage as Show. |
| `src/index.ts` (edit) | Export `Show`, `Hide`, and their prop types. |

### Proposed API

```tsx
interface ResponsiveVisibilityProps {
  /** Viewport min-width breakpoint */
  above?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
  /** Viewport max-width breakpoint (mutually exclusive with above) */
  below?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
  /** Container min-width (Nexus extension over Atlassian) */
  containerAbove?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
  /** Container max-width (Nexus extension over Atlassian) */
  containerBelow?: 'sm' | 'md' | 'lg' | 'xl' | '2xl';
  /** HTML element to render (default 'span') */
  as?: keyof JSX.IntrinsicElements;
  children: React.ReactNode;
}
```

**Constraint**: exactly one of `above` / `below` / `containerAbove` / `containerBelow` must be provided. TypeScript discriminated union enforces this at compile time.

### Implementation sketch

```tsx
// Conceptual — final implementation may differ
function Show({ above, below, containerAbove, containerBelow, as: Tag = 'span', children, ...rest }: ResponsiveVisibilityProps) {
  const className = cn(
    above && `nx:hidden nx:${above}:block`,
    below && `nx:block nx:${below}:hidden`,
    containerAbove && `nx:hidden nx:@${containerAbove}:block`,
    containerBelow && `nx:block nx:@${containerBelow}:hidden`,
  );
  return <Tag data-slot="show" className={className} {...rest}>{children}</Tag>;
}
```

(Real implementation must also handle the `display: revert` semantic for children that have their own display rules, e.g., `display: flex` parents.)

### Documentation

Document in `responsive.md` (created in [8/10]) as the canonical way to express responsive visibility in JSX. Discourage raw `nx:hidden nx:lg:block` when a primitive is available.

## ⚠ Pre-implementation spike (do first, before writing components)

**Runtime spike — `display: revert` semantics with non-block children.**

The Atlassian source (`@atlaskit/primitives/dist/types/responsive/show.d.ts`) notes that `<Show>` uses `display: none` ↔ `display: revert` to toggle visibility. `display: revert` returns the property to its **user-agent default** (typically `inline` for `span`, `block` for `div`), NOT to whatever was set by an ancestor's CSS.

This breaks for children with `display: flex` / `display: grid` inherited from parent context. Example:

```tsx
<div className="nx:flex">
  <Show above="lg" as="div">
    Sidebar item that should be a flex child
  </Show>
</div>
```

When `Show` toggles back on, `display: revert` makes the child `block`, not `flex`. Its size/layout breaks until refresh.

**Spike steps** (~30 min):

1. Create a minimal repro: `<Show>` inside a flex parent
2. Toggle visibility at the breakpoint (resize window past `lg`)
3. Verify whether the child renders as flex item or as a block (broken)
4. If broken, decide between two approaches:
   - **`display: contents`** for show state — the element disappears from layout but children remain in flex flow (caveat: breaks accessibility for some assistive tech)
   - **`@container` + `display: none`** without revert — use the component's own `display` rules + a CSS class toggle instead of inline style toggle
5. Document the chosen approach in the component's comment header

This spike MUST be done before writing the components, not after. The wrong approach will be discovered late by a consumer hitting the flex-child bug.

## Acceptance criteria

- [ ] **Spike above completed**; chosen visibility-toggle approach documented in the component header comment
- [ ] `Show` and `Hide` components exist in `packages/react/src/components/primitives/`
- [ ] Both support all four prop axes (above / below / containerAbove / containerBelow)
- [ ] TypeScript discriminated union enforces "exactly one axis" at compile time
- [ ] Storybook stories cover each prop axis
- [ ] Play tests verify visibility at expected viewport / container widths
- [ ] **Play test verifies behavior inside a flex parent** — child remains flex item after toggle, not block
- [ ] Exports added to `src/index.ts`
- [ ] `containerAbove` / `containerBelow` work with Tailwind v4's `@container` parent context (verified via play test)
- [ ] **Container query naming**: confirmed that `nx:@sm:` / `nx:@md:` etc. are generated correctly under `prefix(nx)` config — verify in built CSS, not just via JSX
- [ ] `responsive.md` updated with usage examples

### v1 scope limitations — document explicitly in component header + Storybook docs

- [ ] **Mark as experimental** in JSDoc (`@experimental`) until at least one consumer adopts and provides feedback. No user research has validated these primitives are wanted; they're inferred from Atlassian's pattern. Track adoption via deferred validation issue (see references).
- [ ] **Named container queries are out of scope for v1**: `containerAbove` / `containerBelow` use the nearest unnamed parent container. If a consumer needs to query a specific named container (e.g., `@container/sidebar` in a nested layout), they fall back to writing raw `nx:@container/{name}/{size}:` utilities. Document this limitation in the component header. File a v2 follow-up issue if a consumer hits the nested-container case.
- [ ] **Documented edge cases**: include in JSDoc the constraints "exactly one axis must be provided" and "children always render (no unmount); only visibility toggles" — so LLM agents using these primitives have the constraints visible at the call site.

## Sequencing

Should ship **after** [6/10] (breakpoint tokens exist) and [8/10] (responsive.md exists). Can ship independent of [7/10] (Dialog migration), though they're conceptually paired.

## References

- Phase 2 artifact: [`reports/phase-2-token-decisions.html § 5 Decision E`](../reports/phase-2-token-decisions.html#bp-decisions)
- Atlassian `<Show>` source: https://atlassian.design/components/primitives/responsive/show
- Atlassian `<Hide>` source: https://atlassian.design/components/primitives/responsive/hide
- Depends-on: [6/10] for breakpoint tokens; [8/10] for the rule file that documents usage

Milestone: [Interaction & layering tokens · Phase X](https://github.com/nexuslabs-ai/nexus/milestone/4)

aishvaryagupta:
## Summary

Ships the experimental `<Show>` / `<Hide>` declarative responsive-visibility primitives in a new `packages/react/src/components/primitives/` folder. Atlassian-inspired, Nexus-extended with container-query axes.

- **Four axes**, exactly one required (a discriminated union enforces it at compile time): `above` / `below` (viewport) and `containerAbove` / `containerBelow` (nearest `@container` — the Nexus extension).
- **`display: contents` ↔ `display: none`** toggle — children always render (no mount/unmount); a `contents` wrapper makes no box, so inline / flex / grid children keep their natural layout.
- Static class-map strategy so the utilities ship in the built CSS; the resolver is shared between the two components (`Hide` = `Show` inverted).

### Pre-implementation spike (mandated by the issue) — done

Resolved empirically against the real Tailwind v4 toolchain:

1. **Mechanism:** `display: contents`, not `display: revert` / `block` — the latter break inline (default `as="span"`) and flex/grid children.
2. **Class strategy:** static literal maps — Tailwind won't emit template-literal (`nx:${bp}:contents`) classes, so they'd be missing from the shipped CSS.
3. **Container scale:** the container axes use Tailwind's native `@container` scale (`md` = 28rem), deliberately distinct from the viewport scale (`md` = 48rem). Documented in `responsive.md`.

## GitHub Issue

Closes #103

## Test Plan

- [x] `yarn workspace @nexus/react typecheck` — union rejects zero/two axes; no export collision
- [x] Lint clean on all new/edited files
- [x] **12/12 storybook play tests pass** in real headless Chrome (vitest browser mode), incl. axe a11y
- [x] Container-query + **flex-child spike** tests verify `display: contents` across a width toggle
- [x] Passing `@container` assertions confirm `nx:@md:contents` emits in the built CSS
- [x] `responsive.md` finalized; scoped Decision-E superseded note added to the phase-2 artifact

## Notes

- Marked `@experimental` (JSDoc) — inferred from Atlassian's pattern; no consumer has validated demand yet.
- Named-container queries (`@container/{name}`) are out of scope for v1.

### DIFF
--- a/packages/react/src/components/primitives/Hide.stories.tsx
+++ b/packages/react/src/components/primitives/Hide.stories.tsx
@@ -1,0 +1,221 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { expect, waitFor, within } from 'storybook/test';
+
+import { Hide } from './hide';
+
+const meta: Meta<typeof Hide> = {
+  title: 'Primitives/Hide',
+  component: Hide,
+  parameters: {
+    layout: 'padded',
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof Hide>;
+
+const box =
+  'nx:rounded-md nx:border nx:border-border-default nx:p-3 nx:text-foreground';
+const label = 'nx:text-muted-foreground nx:mb-1 nx:block';
+
+// ============================================
+// VISUAL
+// ============================================
+
+export const Default: Story = {
+  render: () => (
+    <Hide above="lg" as="div">
+      <div className={box}>
+        Hidden when the viewport is <strong>lg</strong> (≥ 64rem) or wider —
+        shown below it.
+      </div>
+    </Hide>
+  ),
+};
+
+export const AllAxes: Story = {
+  render: () => (
+    <div className="nx:flex nx:flex-col nx:gap-4">
+      <div>
+        <span className={label}>
+          {'<Hide above="lg">'} — hidden viewport ≥ lg
+        </span>
+        <Hide above="lg" as="div">
+          <div className={box}>Hidden on wide viewports.</div>
+        </Hide>
+      </div>
+      <div>
+        <span className={label}>
+          {'<Hide below="lg">'} — hidden viewport &lt; lg
+        </span>
+        <Hide below="lg" as="div">
+          <div className={box}>Hidden on narrow viewports.</div>
+        </Hide>
+      </div>
+      <div
+        className="nx:@container nx:rounded-md nx:border nx:border-border-default nx:p-2"
+        style={{ width: 600 }}
+      >
+        <span className={label}>
+          {'<Hide containerAbove="md">'} — hidden in this 600px container
+        </span>
+        <Hide containerAbove="md" as="div">
+          <div className={box}>Hidden because the container is wide.</div>
+        </Hide>
+      </div>
+      <div
+        className="nx:@container nx:rounded-md nx:border nx:border-border-default nx:p-2"
+        style={{ width: 240 }}
+      >
+        <span className={label}>
+          {'<Hide containerBelow="md">'} — hidden in this 240px container
+        </span>
+        <Hide containerBelow="md" as="div">
+          <div className={box}>Hidden because the container is narrow.</div>
+        </Hide>
+      </div>
+    </div>
+  ),
+};
+
+// ============================================
+// INTERACTION / PLAY TESTS  (inverse of <Show>)
+// ============================================
+
+export const ViewportAxis: Story = {
+  render: () => (
+    <div className="nx:flex nx:flex-col nx:gap-2">
+      <Hide above="lg" as="div" data-testid="above">
+        <div className={box}>above=&quot;lg&quot;</div>
+      </Hide>
+      <Hide below="lg" as="div" data-testid="below">
+        <div className={box}>below=&quot;lg&quot;</div>
+      </Hide>
+    </div>
+  ),
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const above = canvas.getByTestId('above');
+    const below = canvas.getByTestId('below');
+    const lg = window.innerWidth >= 1024;
+    // Inverse of <Show>: Hide above="lg" is hidden at ≥ lg.
+    await waitFor(() => {
+      expect(getComputedStyle(above).display).toBe(lg ? 'none' : 'contents');
+      expect(getComputedStyle(below).display).toBe(lg ? 'contents' : 'none');
+    });
+  },
+};
+
+export const ContainerAxis: Story = {
+  render: () => (
+    <div className="nx:flex nx:gap-4">
+      <div className="nx:@container" style={{ width: 600 }} data-testid="wide">
+        <Hide containerAbove="md" as="div" data-testid="wide-above">
+          <div className={box}>wide · containerAbove</div>
+        </Hide>
+        <Hide containerBelow="md" as="div" data-testid="wide-below">
+          <div className={box}>wide · containerBelow</div>
+        </Hide>
+      </div>
+      <div
+        className="nx:@container"
+        style={{ width: 240 }}
+        data-testid="narrow"
+      >
+        <Hide containerAbove="md" as="div" data-testid="narrow-above">
+          <div className={box}>narrow · containerAbove</div>
+        </Hide>
+        <Hide containerBelow="md" as="div" data-testid="narrow-below">
+          <div className={box}>narrow · containerBelow</div>
+        </Hide>
+      </div>
+    </div>
+  ),
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    // Inverse of <Show>: 600 ≥ md, 240 < md.
+    await waitFor(() => {
+      expect(getComputedStyle(canvas.getByTestId('wide-above')).display).toBe(
+        'none'
+      );
+      expect(getComputedStyle(canvas.getByTestId('wide-below')).display).toBe(
+        'contents'
+      );
+      expect(getComputedStyle(canvas.getByTestId('narrow-above')).display).toBe(
+        'contents'
+      );
+      expect(getComputedStyle(canvas.getByTestId('narrow-below')).display).toBe(
+        'none'
+      );
+    });
+  },
+};
+
+// Confirms the `display: contents` mechanism also holds for <Hide>'s shown
+// state: <Hide containerBelow="md"> is shown at ≥ md, so its child is a flex item.
+export const FlexParent: Story = {
+  render: () => (
+    <div
+      className="nx:@container"
+      style={{ width: 600 }}
+      data-testid="container"
+    >
+      <div className="nx:flex nx:gap-4">
+        <div
+          data-testid="item-a"
+          className="nx:rounded-md nx:border nx:border-border-default"
+          style={{ width: 48, height: 48 }}
+        />
+        <Hide containerBelow="md" data-testid="wrap">
+          <div
+            data-testid="item-b"
+            className="nx:rounded-md nx:border nx:border-border-primary"
+            style={{ width: 48, height: 48 }}
+          />
+        </Hide>
+        <div
+          data-testid="item-c"
+          className="nx:rounded-md nx:border nx:border-border-default"
+          style={{ width: 48, height: 48 }}
+        />
+      </div>
+    </div>
+  ),
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const container = canvas.getByTestId('container');
+    const wrap = canvas.getByTestId('wrap');
+    const rect = (id: string) => canvas.getByTestId(id).getBoundingClientRect();
+
+    // ≥ md: containerBelow is shown -> `contents`, item-b is a real flex sibling.
+    await waitFor(() =>
+      expect(getComputedStyle(wrap).display).toBe('contents')
+    );
+    expect(rect('item-b').top).toBe(rect('item-a').top);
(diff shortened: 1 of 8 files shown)


examlly-tech (commented):
## 🤖 Principal Architect Review — NEEDS DISCUSSION

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `responsive-visibility.ts:62` | `as?: ElementType` advertises polymorphism but type-rejects element-specific props (`as="a"` + `href`) and a custom component's required props |

### Architectural Concerns

**`as` contract under-delivers vs its type** — `as?: ElementType` + `HTMLAttributes<HTMLElement>` permits the tag swap but rejects any tag-specific attribute and erases a custom component's props contract. Defensible for a non-semantic `@experimental` wrapper; pick one — make it generic (`<C extends ElementType>` + `ComponentPropsWithoutRef<C>`), or narrow the type so it promises only tag-name swapping.

---
_Review against `.claude/rules/` — focus: module boundaries & type-contract design._

packages/react/src/components/primitives/responsive-visibility.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+export type ResponsiveVisibilityProps = HTMLAttributes<HTMLElement> & {
+  /**
+   * Element rendered as the display-managed wrapper. Keep it non-semantic:
+   * `display: contents` can drop a semantic element's box from the accessibility
+   * tree, so put semantics on a child *inside* `<Show>` / `<Hide>` instead.
+   * @default 'span'
+   */
+  as?: ElementType;
```
examlly-tech: **`as` typing (arch + code-quality):** `as?: ElementType` allows tag swapping but type-rejects element-specific props (`as="a"`+`href` won't compile) and a custom component's required props. Fine for this `@experimental` non-semantic wrapper — either go generic `<C extends ElementType>`, or narrow the type to tag-swap-only so it can't promise more than it delivers.

packages/react/src/components/primitives/Show.stories.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <div className={box}>below=&quot;lg&quot;</div>
+      </Show>
+    </div>
+  ),
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const above = canvas.getByTestId('above');
+    const below = canvas.getByTestId('below');
+    // Self-consistent against the runner's own viewport — no hard-coded width.
+    const lg = window.innerWidth >= 1024;
```
examlly-tech: **Tautological viewport test:** the runner's viewport is fixed (~1200px), so `lg` is always true and only the `≥lg` branch runs; the `none` / `<lg` branch is never asserted in CI. Split into two stories whose ternary collapses to a constant, or assert both branches via a forced width.

packages/react/src/components/primitives/Hide.stories.tsx
```diff
@@ -0,0 +1,221 @@
...
+      <Hide below="lg" as="div" data-testid="below">
+        <div className={box}>below=&quot;lg&quot;</div>
+      </Hide>
+    </div>
+  ),
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const above = canvas.getByTestId('above');
+    const below = canvas.getByTestId('below');
+    const lg = window.innerWidth >= 1024;
```
examlly-tech: **Same as Show ViewportAxis:** keyed to the fixed runner viewport, so only one branch executes — the `<Hide>` inverse case at `<lg` is never exercised. Mirror whatever fix lands on `Show.stories` ViewportAxis.

examlly-tech (commented):
## 🤖 SDE2 Review — MINOR CHANGES

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `Show.stories.tsx:96` | ViewportAxis play test keyed to the runner's fixed viewport — only the `≥lg` branch runs in CI; the `<lg` branch is never asserted |
| 2 | ⚠️ Minor | `Hide.stories.tsx:100` | Same tautological viewport assertion — the inverse branch never executes |

---
_Review against `.claude/rules/` — focus: code quality & correctness._

examlly-tech (approved):
## 🤖 Principal Architect Follow-up — APPROVED

### Unresolved Issues

| 1 | `as?: ElementType` over-promised polymorphism (rejected `as="a"`+`href` / custom-component props) | `responsive-visibility.ts` | ✅ Fixed | Narrowed to `'span' \| 'div'` — enforces the non-semantic `display:contents` constraint at every call site |

A full `<C extends ElementType>` generic would re-admit `as="nav"`/`as="button"` — the misuse `display:contents` can't support — so the literal-union narrowing is the right call, not just an acceptable one. `ElementType` import removed with no stale refs; all `as=` story usages are `span`/`div`. No new architectural issues in this change set.

---
_Follow-up review against `.claude/rules/` — focus: module boundaries & type-contract design._

examlly-tech (approved):
## 🤖 SDE2 Follow-up — APPROVED

### Unresolved Issues

| 1 | Tautological viewport play test (gated on `window.innerWidth`; only the ≥lg branch ran) | `Show.stories.tsx` | ✅ Fixed | Now asserts resolved classes via `toHaveClass`, both axes, deterministic |
| 2 | Same tautology, inverse case | `Hide.stories.tsx` | ✅ Fixed | Same fix; asserts the inverted `Hide`=`Show` mapping |

Verified `cn()` (tailwind-merge, `prefix:'nx'`) keeps both `nx:hidden` + `nx:lg:contents` (distinct variant scopes don't conflict), so the multi-arg `toHaveClass` presence checks are robust. The `as?: 'span' \| 'div'` narrowing is sound for JSX intrinsic rendering, `ElementType` is fully removed, and `waitFor` is still used by `ContainerAxis`/`FlexParent`. No new issues.

---
_Follow-up review against `.claude/rules/` — focus: code quality & correctness._

PR #213, opened 2026-05-25
chore(tooling): format-check .claude/*.md, ignore .claude/worktrees in lint/prettier

aishvaryagupta:
## Summary

Two `.claude/` blind spots in the tooling config:

- **Lint over-reached.** `eslint .` (and prettier) had no ignore for `.claude/worktrees/` — untracked full-repo worktree copies — so `yarn lint` failed locally on their pre-existing errors. Now ignored in eslint + prettier, and gitignored so they're never committable.
- **Prettier under-reached.** `format:check`'s `**/*.{…,md,…}` glob doesn't descend into dot-directories, so the entire `.claude/*.md` corpus (rules, skills, commands, agents) was never CI-format-checked. Now `.claude/**/*.md` is added to `format` + `format:check`.

No formatting churn — every existing `.claude/*.md` already passes prettier.

## Changes

- `.gitignore`: ignore `.claude/worktrees/`
- `eslint.config.js`: ignore `**/.claude/worktrees/**`
- `.prettierignore`: ignore `.claude/worktrees/`
- `package.json`: add `.claude/**/*.md` to `format` and `format:check`

## Test Plan

- [x] `yarn format:check` → green (now covers `.claude/**/*.md`)
- [x] `yarn lint` → green (no longer descends into `.claude/worktrees/`)


examlly-tech (commented):
## 🤖 Principal Architect Review — APPROVED

Config-only tooling change — no module-boundary, data-model, security-boundary, scalability, API-contract, or stub-route surface in scope.

**Worktree-ignore boundary (the one architectural angle):** ignoring `.claude/worktrees/` across git + eslint + prettier is correct, not over-reach. Worktree placement under `.claude/worktrees/` is imposed by external tooling — no in-repo config controls it — so the repo can't relocate them outside the tree; ignore-everywhere is the only available boundary. It also matches existing precedent (`dist`, `__generated__` are each ignored in these same three files). The `.prettierignore` entry is load-bearing (worktrees are full-repo clones the broad `**/*` glob would otherwise descend into), and the two prettier globs compose without gap.

examlly-tech (approved):
## 🤖 SDE2 Review — APPROVED

All in-lane claims verified empirically:
- `.claude/**/*.md` and the legacy `**/*.{…,md,…}` glob are disjoint for this PR's targets — prettier doesn't descend into nested dot-dir paths, and no top-level `.claude/*.md` files exist → zero double-processing.
- `.prettierignore` wins over the new glob: a malformatted probe placed under `.claude/worktrees/` reports clean even when prettier is pointed directly at it.
- eslint `**/.claude/worktrees/**` is the correct, more-defensive shape (matches root + any nested worktrees dir; over-match is in the safe direction).
- `format` / `format:check` carry byte-identical patterns (only `--write` vs `--check` differ). ✓

PR #222, opened 2026-05-26
fix(storybook): apply .dark to <html> so portaled components theme in dark mode

### LINKED ISSUE #137: fix(storybook): apply .dark to <html> so portaled components theme in dark mode
## Summary

Storybook's dark-mode toolbar toggle applies the `dark` class to the decorator's wrapper `<div>` (`.storybook/preview.tsx:71`). Radix portals (Dialog overlay/content, and future Dropdown/Select/Popover/Tooltip/Toast/Sheet) render to `document.body` — **outside** that wrapper — so portaled components render with **light-theme tokens in dark mode**.

## Current behavior

1. Open a portaled component's story (e.g. `Components/Dialog → Default`).
2. Switch the toolbar theme to **Dark**.
3. The overlay + content render light (white card, light 75% scrim) instead of dark.

Verified on Dialog: in dark mode the overlay resolves to the light `--color-overlay` value; forcing `document.documentElement.classList.add('dark')` makes both the scrim and the content card resolve correctly.

## Expected behavior

The dark-mode toggle themes the entire preview, including portaled subtrees.

## Root cause

The decorator scopes `.dark` to the story wrapper `<div>`. The `.dark { … }` CSS scope only cascades to that div's descendants; `createPortal(…, document.body)` escapes it. In a real app `.dark` lives on `<html>`/`<body>`, so portals inherit it — this is a Storybook-harness-only gap.

## Proposed fix

Sync `dark` onto `document.documentElement` from the decorator instead of the wrapper div:

```tsx
import { useEffect } from 'react';
// …
decorators: [
  (Story, context) => {
    const theme = context.globals.theme;
    const isDocs = context.viewMode === 'docs';
    useEffect(() => {
      document.documentElement.classList.toggle('dark', theme === 'dark');
    }, [theme]);
    return (
      <div className={`nx:flex nx:items-center nx:justify-center nx:bg-background nx:text-foreground ${isDocs ? 'nx:py-12' : 'nx:min-h-screen'}`}>
        <Story />
      </div>
    );
  },
],
```

(`useEffect` syncing the DOM outside React's tree is the correct use per `.claude/rules/useeffect-escape-hatch.md`.)

## Affected components

All components that portal to `document.body`: Dialog (today); Dropdown, Select, Popover, Tooltip, Toast, Sheet as they land.

## Acceptance criteria

- [ ] Dark toggle themes portaled components (Dialog overlay + content render dark)
- [ ] Light mode unchanged
- [ ] Docs mode (multiple stories per page) still themes correctly
- [ ] Wrapper `bg-background` / `text-foreground` still resolve in both themes
- [ ] No `exhaustive-deps` suppression

## Discovered

While verifying the overlay→alpha migration in #81 (PR #135). Pre-existing; orthogonal to that PR.

aishvaryagupta:
## Summary

- Move `.dark` from the Storybook decorator's wrapper `<div>` to `document.documentElement` via `useEffect`.
- Radix portals (Dialog, DropdownMenu, Select, Tooltip — current and future) render to `document.body`, outside the wrapper, so they previously ignored the toolbar dark toggle and resolved light tokens in dark mode. With `.dark` on `<html>`, portal subtrees inherit the theme through the normal cascade.
- Decorator extracted to a named `ThemeDecorator: Decorator` so `react-hooks/rules-of-hooks` (configured as `error` in `eslint.config.js`) recognizes the call site as a component.
- `useEffect` is the right escape hatch per `.claude/rules/useeffect-escape-hatch.md` — it syncs an external system (the DOM outside React's tree), not React state.

## GitHub Issue

Closes #137

## Why not `useLayoutEffect`

`useEffect` runs after first paint, so a returning user with `dark` selected will see one frame of light theme before the effect snaps `<html>` to dark on initial load. Acceptable trade-off: `useLayoutEffect` would prevent the flash but blocks paint and adds an SSR-warning footgun (irrelevant for Storybook today, but a maintenance trap). The toolbar-toggle case has no flash either way. If FOUC becomes a complaint, swap in a follow-up.

## What stays unchanged

- Wrapper `nx:bg-background` / `nx:text-foreground` still resolve in both themes — they read `var(--nx-color-*)` which cascades from `<html>`.
- Generated base-variants stories are unaffected: each cell sets `[data-nexus-base="X"]{…}` and `[data-nexus-base="X"].dark{…}` directly on the cell, so CSS variable closest-ancestor inheritance wins regardless of what `<html>` defines.
- `Shadow.Dark` (the only story with per-story `globals: { theme: 'dark' }`) renders dark as before; navigating to a light story flips `<html>` back via the next decorator render.

## Test Plan

- [x] `yarn lint` — clean (`eslint . --max-warnings 0` from root)
- [ ] Manual: `Components/Dialog → Default` → toolbar `Dark` → overlay + content render dark
- [ ] Manual: toolbar `Light` → unchanged
- [ ] Manual: docs page (autodocs) → theming still works across multiple stories

### DIFF
--- a/packages/react/.storybook/preview.tsx
+++ b/packages/react/.storybook/preview.tsx
@@ -1,6 +1,25 @@
-import type { Preview } from '@storybook/react-vite';
+import { useEffect } from 'react';
+
+import type { Decorator, Preview } from '@storybook/react-vite';
 
 import '../src/index.css';
+
+const ThemeDecorator: Decorator = (Story, context) => {
+  const theme = context.globals.theme;
+  const isDocs = context.viewMode === 'docs';
+
+  useEffect(() => {
+    document.documentElement.classList.toggle('dark', theme === 'dark');
+  }, [theme]);
+
+  return (
+    <div
+      className={`nx:flex nx:items-center nx:justify-center nx:bg-background nx:text-foreground ${isDocs ? 'nx:py-12' : 'nx:min-h-screen'}`}
+    >
+      <Story />
+    </div>
+  );
+};
 
 const preview: Preview = {
   // Enable autodocs for all stories globally
@@ -62,19 +81,7 @@ const preview: Preview = {
       },
     },
   },
-  decorators: [
-    (Story, context) => {
-      const theme = context.globals.theme;
-      const isDocs = context.viewMode === 'docs';
-      return (
-        <div
-          className={`nx:flex nx:items-center nx:justify-center nx:bg-background nx:text-foreground ${theme === 'dark' ? 'dark' : ''} ${isDocs ? 'nx:py-12' : 'nx:min-h-screen'}`}
-        >
-          <Story />
-        </div>
-      );
-    },
-  ],
+  decorators: [ThemeDecorator],
 };
 
 export default preview;


examlly-tech (approved):
**Principal Architect — Verdict: APPROVE**

Scope: architecture / seam / type system.

The seam is the right shape for the bug: Radix portals attach to `document.body`, so the only ancestor that reaches them is `<html>`. The decorator is the centralized place to write the class, and `useEffect` keyed on `[theme]` is the documented escape-hatch use per `.claude/rules/useeffect-escape-hatch.md`. `ThemeDecorator: Decorator` (from `@storybook/react-vite`) matches the framework contract. The mutation is scoped to Storybook's iframe document — nothing in shipped library code touches `documentElement`. Base-variants cell scoping (`[data-nexus-base="X"].dark`) still wins via closest-ancestor custom-property inheritance. Docs mode is safe — all decorator instances on a page see the same `context.globals.theme` and write idempotently.

examlly-tech (approved):
**SDE2 — Verdict: APPROVE**

Scope: type safety, library API usage, edge cases, readability.

Types: `Decorator` from `@storybook/react-vite` resolves to `(fn: PartialStoryFn, c: StoryContext) => ReactNode` — annotation is correct, `<Story />` JSX form remains valid. Library API: `classList.toggle(name, force)` with explicit boolean is the documented MDN form — idempotent under StrictMode double-effect. Hook lint: hoisting to PascalCase `ThemeDecorator` satisfies `react-hooks/rules-of-hooks` (configured as `error`); deps `[theme]` is exhaustive, no suppression. Edge cases: `context.globals.theme` falsy or any non-`'dark'` value resolves to light (safe fallback); no cleanup function needed since the next decorator render re-syncs and the iframe document is dedicated to Storybook. Import ordering matches `simple-import-sort` groups.

PR #280, opened 2026-06-01
feat(console): add core auth — login/signup/OTP/forgot + route guards

aishvaryagupta:
## Summary

Phase 1a of the Atlas build — **core auth**, MSW-backed (no real backend). Adds the authenticated/unauthenticated split and four auth screens.

- **Session** — Zustand `useSession` store, persisted to localStorage. The router's `beforeLoad` guards read it synchronously via `getState()` (persist uses sync storage, so it's hydrated before first paint).
- **Routing** — two pathless layout routes off the root: `_app` (the existing shell + an auth guard that bounces unauthenticated visitors to `/login`) and `_auth` (a shell-less split-panel layout + a mirror redirect-if-authed guard). The thin root route holds only the app-wide `<Toaster>`. URLs stay clean (`/login`, `/design/reference`); the pathless ids only surface in route-id `from`s (e.g. `/app/m/$module`).
- **Screens** — login (prefilled demo creds), signup, 2FA OTP verify (`InputOTP`, code `123456`), forgot-password (resolves to a "check your email" confirmation — the reset-completion screen is reached from the email link, so it's out of scope here). All reuse the rhf + zod + `useMutation` pattern.
- **Shell** — the sidebar Account footer is now a dropdown showing the signed-in user + a **Sign out** item (clears the session, navigates to `/login`).
- **MSW** — `/api/auth/{login,signup,verify-otp,forgot}`; login/signup accept any email + an 8+ char password and hand off to the OTP step; verify-otp authenticates on `123456`.

## Scope (tracked in `docs/plans/atlas-company-os.md`)

Phase 1 is split into **1a (this PR)** and **1b** — SSO/OAuth · magic-link · accept-invite · session-expired · reset-password completion **+** onboarding (workspace creation · invite team · choose plan · checklist · first-run empty states). The plan doc carries the split; this is not an untracked deferral.

The Design System showcase now sits behind the auth guard (a deliberate B2B-SaaS shape) — demo creds are prefilled and the session persists, so it's one-time friction.

## Test Plan

Walked end-to-end in-browser (light + dark):

- [x] Unauthed hit on `/design/reference` → bounced to `/login`
- [x] Login (demo creds) → `/verify?email=…` → OTP `123456` → `/design/reference`
- [x] Invalid OTP → inline error alert
- [x] Account dropdown → **Sign out** → session cleared → bounced to `/login`
- [x] Dark-mode auth screen renders correctly (base tokens apply outside the shell)
- [x] Forgot → confirmation state
- [x] Signup → `/verify` → OTP → app
- [x] `yarn workspace @nexus/console typecheck` + eslint clean

### DIFF
--- a/apps/console/src/app/root.tsx
+++ b/apps/console/src/app/root.tsx
@@ -1,0 +1,17 @@
+import { Toaster } from '@nexus/react';
+import { Outlet } from '@tanstack/react-router';
+
+/**
+ * The thin root rendered by the root route. It holds only app-wide chrome that
+ * must sit above BOTH the authenticated shell and the auth screens — currently
+ * just the toast portal — plus an <Outlet> for whichever pathless layout (the
+ * authenticated `_app` shell or the shell-less `_auth` screens) matches.
+ */
+export function Root() {
+  return (
+    <>
+      <Outlet />
+      <Toaster />
+    </>
+  );
+}
--- a/apps/console/src/mocks/handlers.ts
+++ b/apps/console/src/mocks/handlers.ts
@@ -7,7 +7,7 @@ const OTP_CODE = '123456';
 
 /** Derive a stable demo user from an email — the mock has no real user table. */
 function userFromEmail(email: string): User {
-  const localPart = email.split('@')[0] ?? 'user';
+  const [localPart = ''] = email.split('@');
   const name = localPart
     .split(/[._-]+/)
     .filter(Boolean)
@@ -26,7 +26,8 @@ type VerifyBody = { email?: string; code?: string };
  * what actually authenticates and returns the user.
  */
 export const handlers: RequestHandler[] = [
-  // Credentials check: any email + an 8+ character password passes → proceed to OTP.
+  // Credentials check — no real validation (any email + an 8+ char password
+  // passes); the OTP step below is the only actual auth gate. Proceeds to OTP.
   http.post('/api/auth/login', async ({ request }) => {
     const { email, password } = (await request.json()) as LoginBody;
     if (!email || !password || password.length < 8) {
--- a/apps/console/src/mocks/handlers.ts
+++ b/apps/console/src/mocks/handlers.ts
@@ -1,7 +1,66 @@
-import type { RequestHandler } from 'msw';
+import { http, HttpResponse, type RequestHandler } from 'msw';
+
+import type { User } from '../lib/auth-api';
+
+/** The fixed demo OTP — shown as a hint on the verify screen. */
+const OTP_CODE = '123456';
+
+/** Derive a stable demo user from an email — the mock has no real user table. */
+function userFromEmail(email: string): User {
+  const localPart = email.split('@')[0] ?? 'user';
+  const name = localPart
+    .split(/[._-]+/)
+    .filter(Boolean)
+    .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+    .join(' ');
+  return { id: email.toLowerCase(), name: name || 'Atlas User', email };
+}
+
+type LoginBody = { email?: string; password?: string };
+type SignupBody = { name?: string; email?: string; password?: string };
+type VerifyBody = { email?: string; code?: string };
 
 /**
- * MSW request handlers. Empty for the Phase 0 skeleton — each module appends
- * its own handlers here as it starts fetching mock data.
+ * MSW auth handlers — there is no real backend. The flow is two-step: a
+ * credentials check (login/signup) hands the email to the OTP step, which is
+ * what actually authenticates and returns the user.
  */
-export const handlers: RequestHandler[] = [];
+export const handlers: RequestHandler[] = [
+  // Credentials check: any email + an 8+ character password passes → proceed to OTP.
+  http.post('/api/auth/login', async ({ request }) => {
+    const { email, password } = (await request.json()) as LoginBody;
+    if (!email || !password || password.length < 8) {
+      return HttpResponse.json(
+        { message: 'Incorrect email or password.' },
+        { status: 401 }
+      );
+    }
+    return HttpResponse.json({ email });
+  }),
+
+  http.post('/api/auth/signup', async ({ request }) => {
+    const { name, email, password } = (await request.json()) as SignupBody;
+    if (!name || !email || !password || password.length < 8) {
+      return HttpResponse.json(
+        { message: 'Please complete every field (password 8+ characters).' },
+        { status: 400 }
+      );
+    }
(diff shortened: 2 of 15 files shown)


aishvaryagupta:
**Reviewer note — MSW is dev-only, and that's a conscious call here.**

This PR gates the whole app behind a login whose endpoints are served only by MSW, which starts only in dev (`if (!import.meta.env.DEV) return;` in `main.tsx`). I checked whether that's a trap: the console app is **never built-and-served outside `vite dev`** — no Vercel/Netlify/Pages/Docker config, and CI only `turbo build`s it as a compile/typecheck artifact (it's never uploaded or served). So there's no path today where login hits a real backend.

The one narrow footgun: `vite build && vite preview` locally would hit a dead login wall. When Atlas actually gets a deployed/preview target (its own change, with the deploy config), that's the point to start MSW in the prod bundle too — not speculatively now.

apps/console/src/app/root.tsx
```diff
@@ -0,0 +1,17 @@
+import { Toaster } from '@nexus/react';
+import { Outlet } from '@tanstack/react-router';
+
+/**
+ * The thin root rendered by the root route. It holds only app-wide chrome that
+ * must sit above BOTH the authenticated shell and the auth screens — currently
+ * just the toast portal — plus an <Outlet> for whichever pathless layout (the
+ * authenticated `_app` shell or the shell-less `_auth` screens) matches.
+ */
+export function Root() {
```
examlly-tech, line 10: **Minor — bare unmatched routes:** `Root` carries no base tokens now, so a 404 / typo URL (matching neither `_app` nor `_auth`) renders against browser-default white — the dark-mode regression `auth-layout.tsx` documents. Add `nx:bg-background nx:text-foreground` (e.g. a `defaultNotFoundComponent`).

apps/console/src/mocks/handlers.ts
```diff
@@ -1,7 +1,66 @@
...
- * MSW request handlers. Empty for the Phase 0 skeleton — each module appends
- * its own handlers here as it starts fetching mock data.
+ * MSW auth handlers — there is no real backend. The flow is two-step: a
+ * credentials check (login/signup) hands the email to the OTP step, which is
+ * what actually authenticates and returns the user.
  */
-export const handlers: RequestHandler[] = [];
+export const handlers: RequestHandler[] = [
+  // Credentials check: any email + an 8+ character password passes → proceed to OTP.
+  http.post('/api/auth/login', async ({ request }) => {
```
examlly-tech, line 31: **Minor — credentials step gates nothing:** any email + an 8-char password proceeds, then fixed `123456` authenticates, so `/login` binds nothing the OTP step relies on. Fine for a dev-only demo — worth one line noting OTP is the only real gate.

examlly-tech (commented):
## 🤖 Principal Architect Review — #280

**Verdict: NEEDS DISCUSSION** — minor, non-blocking.

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `apps/console/src/app/root.tsx` | Thinning the root from `RootLayout`→`Root` leaves unmatched URLs (404 / typos) rendering bare — they match neither pathless layout, so no base surface/dark tokens apply. |
| 2 | ⚠️ Minor | `apps/console/src/mocks/handlers.ts` | The credentials step gates nothing — any email + an 8-char password proceeds, and fixed `123456` authenticates. Acceptable for a dev-only demo. |

### Architectural Concerns

**Unmatched-route base-token regression** — Before this PR `rootRoute.component` was `RootLayout` (base tokens + shell), so an unmatched URL still rendered against the base surface; now it renders bare under the thin `Root`, reproducing the dark-mode regression `auth-layout.tsx`'s own comment warns about. Apply a document-level base (or a `notFoundComponent`); the `/auth` and `/app` subtrees already carry their own.

Topology itself is sound: the mirror guards can't loop, `persist` rehydrates synchronously before `RouterProvider` mounts (so `getState().user` in `beforeLoad` is never stale-null), and sign-out navigates explicitly.

---

apps/console/src/mocks/handlers.ts
```diff
@@ -1,7 +1,66 @@
...
+import { http, HttpResponse, type RequestHandler } from 'msw';
+
+import type { User } from '../lib/auth-api';
+
+/** The fixed demo OTP — shown as a hint on the verify screen. */
+const OTP_CODE = '123456';
+
+/** Derive a stable demo user from an email — the mock has no real user table. */
+function userFromEmail(email: string): User {
+  const localPart = email.split('@')[0] ?? 'user';
```
examlly-tech: **Minor — dead fallback:** `String.split` always returns at least one element, so `?? 'user'` is unreachable, and the empty-localPart case is already handled by `name || 'Atlas User'` on line 16. Drop the `?? 'user'`.

examlly-tech (approved):
## 🤖 SDE2 Review — #280

**Verdict: APPROVED** — one minor, non-blocking.

Type safety, library usage (rhf `setError('root')` + `noValidate`, zod-4 `z.email`/`.min`, `useMutation` onSuccess/onError/`isPending`, `getRouteApi('/auth/verify')`, Radix `onSelect`, controlled `InputOTP`), and the `post()` error path all check out. `typecheck` + ESLint pass.

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `apps/console/src/mocks/handlers.ts` | `email.split('@')[0] ?? 'user'` — `String.split` always returns ≥1 element, so `?? 'user'` is unreachable; `name \|\| 'Atlas User'` already covers the empty-localPart case. |

---

aishvaryagupta:
All three review findings addressed in `6bccc07` (every one fixed in-PR per `no-follow-up-deferral.md`):

| Finding | Resolution |
| --- | --- |
| **Architect — unmatched routes render bare** (`root.tsx`) | Added `defaultNotFoundComponent` → a themed 404 (`app/not-found.tsx`) that carries its own `nx:bg-background nx:text-foreground`. Browser-verified in dark mode: renders the dark surface, not bare. |
| **SDE2 — dead `?? 'user'` fallback** (`handlers.ts`) | Collapsed the two competing fallbacks into one. Note: the literal "drop `?? 'user'`" would break under `noUncheckedIndexedAccess` (`[0]` is `string \| undefined`), so used `const [localPart = ''] = …`, leaving `name \|\| 'Atlas User'` as the single real fallback. |
| **Architect — credentials step gates nothing** (`handlers.ts`) | Clarified the inline comment: the credentials step does no real validation; the OTP step is the only gate. |

`typecheck` + ESLint clean. Both review verdicts were non-blocking (Architect COMMENT, SDE2 APPROVE).

Issue #282, opened 2026-06-02
feat(react): add EmptyState component


aishvaryagupta:
## Gap

Surfaced by the Atlas console CRM (Phase 2a). List/table/record views need a consistent empty state: icon + title + description + optional primary action.

Meanwhile, a **one-off empty message** is used in the CRM Contacts view. This issue tracks a reusable `@nexus/react` **EmptyState** component.

## Scope
- `nx:`-prefixed, semantic tokens, `data-slot`, composition via `children`/slots, story-coverage gate per `.claude/rules/testing-react.md`.

_Tracked gap from the Atlas dogfooding effort._

Issue #281, opened 2026-06-02
feat(react): add Pagination component


aishvaryagupta:
## Gap

Surfaced by the Atlas console CRM (Phase 2a — the Contacts DataTable). List/table views need page navigation (prev/next, page numbers, page-size select, disabled edges, a11y labels).

Meanwhile, an **app-local pager** lives inside `apps/console/src/components/data-table.tsx` (sanctioned app-composite content — not a hand-rolled DS component). This issue tracks promoting it to a polished, reusable `@nexus/react` **Pagination** component.

## Scope
- `nx:`-prefixed, semantic tokens, `data-slot`, story-coverage gate per `.claude/rules/testing-react.md`.
- shadcn ships a `Pagination` primitive as a structural reference.

_Tracked gap from the Atlas dogfooding effort._

PR #283, opened 2026-06-02
feat(console): CRM contacts table — sortable/filterable DataTable (Phase 2a)

aishvaryagupta:
## Summary

Phase 2a of the CRM flagship — a **Contacts list** backed by MSW, built on a reusable app-level **DataTable** composite. First slice of Phase 2 (table → 2b detail · 2c create/edit · 2d kanban · 2e saved views).

- **DataTable** (`apps/console/src/components/data-table.tsx`) — headless `@tanstack/react-table` over the Nexus `Table`/`Button`/`Input` primitives: client-side **sort**, single-column **filter**, **row selection**, app-local **pager**. Domain-agnostic — Phase 3 modules reuse it with their own column defs. (Per the plan's folder shape + shadcn's canonical position, DataTable is an *app composite*, not a DS component — columns/actions are domain-specific.)
- **Contacts** — typed `Contact` model + 24 deterministic fixtures spanning every column type (text · status **Badge** · owner **Avatar** · currency · date) + `GET /api/crm/contacts` (MSW, 500ms delay so the loading **Skeleton** renders).
- **Routing** — static `/m/crm` renders the table and **outranks** the dynamic `/m/$module` placeholder; the sidebar highlights by pathname and links built modules to their real route (fixes a TanStack route-template-mismatch warning).
- `@tanstack/react-table` added (headless dep — no Storybook gate, and the bundle-size gate is on `@nexus/react`, not the app).

## Design-system gaps surfaced (tracked, not hand-rolled)

The Contacts table needs two DS components Nexus doesn't have yet. Per the agreed approach, app-local stand-ins ship now and the polished components are tracked:
- **Pagination** → #281 (app-local pager lives inside the DataTable composite)
- **EmptyState** → #282 (one-off empty message in the Contacts view)

Surfacing these as tracked issues is the Atlas dogfooding deliverable.

## Note for reviewers

Column defs use `cell`/`header` render callbacks (`({ row }) => <JSX>`). This *looks* like a composition-over-render-props violation, but it's the shape `@tanstack/react-table` mandates — `composition-over-render-props.md` explicitly exempts "third-party library APIs that mandate this shape."

## Test Plan

Walked end-to-end in-browser (`/m/crm`, authed):

- [x] Table renders — status badges, owner avatars, currency, formatted dates
- [x] **Sort** (Open value desc → \$215,000 top), **Filter** ("zhang" → 1 row), **Paginate** (→ Page 2 of 3), **Select** (→ "1 of 24 selected")
- [x] Static `/m/crm` wins over `/m/\$module`; `/m/dashboard` still shows "coming soon"
- [x] Sidebar highlights CRM; loading Skeleton shows on first load
- [x] 0 console errors / 0 warnings; `typecheck` + ESLint clean


apps/console/src/components/data-table.tsx
```diff
@@ -0,0 +1,167 @@
...
+              </TableRow>
+            )}
+          </TableBody>
+        </Table>
+      </div>
+
+      {/* App-local pager — the polished @nexus/react Pagination is tracked in #281. */}
+      <div className="nx:flex nx:items-center nx:justify-between nx:gap-4">
+        <p className="nx:text-muted-foreground nx:text-sm">
+          {selectedCount} of {totalCount} row(s) selected.
```
examlly-tech: **Minor — selection footer always renders:** the select column is domain-supplied (opt-in), but this summary isn't — so a Phase 3 table without a select column shows a permanent "0 of N row(s) selected." Gate it on selection being enabled; that's the composite's stated reusability contract.

apps/console/src/components/data-table.tsx
```diff
@@ -0,0 +1,167 @@
...
+  columns,
+  data,
+  filterColumn,
+  filterPlaceholder,
+}: DataTableProps<TData, TValue>) {
+  const [sorting, setSorting] = useState<SortingState>([]);
+  const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
+  const [rowSelection, setRowSelection] = useState({});
+
+  const table = useReactTable({
```
examlly-tech, line 58: **Minor — implicit page size:** no `initialState.pagination.pageSize`, so paging rides TanStack's default of 10, which the fixtures comment silently relies on. Make it explicit (or a `pageSize` prop) per explicit-over-implicit.

apps/console/src/lib/crm-api.ts
```diff
@@ -0,0 +1,31 @@
...
+  lastContacted: string;
+};
+
+export async function fetchContacts(): Promise<Contact[]> {
+  const res = await fetch('/api/crm/contacts');
+  if (!res.ok) {
+    throw new Error('Failed to load contacts. Please try again.');
+  }
+  const data = (await res.json()) as { contacts: Contact[] };
+  return data.contacts;
```
examlly-tech: **Minor — envelope discarded:** returning a bare `Contact[]` drops the `{ contacts }` wrapper, so a future server-side `totalCount` has nowhere to live without a type change. Return the envelope to keep the server-side swap contained.

examlly-tech (commented):
## 🤖 Principal Architect Review — #283

**Verdict: NEEDS DISCUSSION** — minor, non-blocking; all fix-now.

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `apps/console/src/components/data-table.tsx` | Selection footer renders unconditionally — a Phase 3 table with no select column shows a permanent "0 of N row(s) selected." |
| 2 | ⚠️ Minor | `apps/console/src/components/data-table.tsx` | Page size rides TanStack's implicit default of 10; the fixtures comment ("3 pages at 10/page") silently depends on it. |
| 3 | ⚠️ Minor | `apps/console/src/lib/crm-api.ts` | `fetchContacts` returns a bare `Contact[]`, discarding the `{ contacts }` envelope — a future server-side `totalCount` has no home in the return type. |

### Architectural Concerns

**DataTable generality vs. its docstring** — The composite claims to "stay data-agnostic so every workspace module reuses it," but the unconditional selection footer (#1) bakes a selection summary into every consumer. Worth resolving before Phase 3 modules inherit it — that's the composite's stated purpose.

The `lib/crm-api.ts` type + `mocks/crm-fixtures.ts` seed + handler split mirrors the auth-api/`User` precedent cleanly; the new `components/` layer is the right home; the static-route + `route` discriminant grows linearly per built module (correct — don't abstract into a registry). `owner: string` (display name, not id) is fine for the mock, but 2c create/edit will want an id.

---

examlly-tech (commented):
## 🤖 SDE2 Review — #283

**Verdict: MINOR CHANGES** — non-blocking; both fix-now.

`typecheck` clean. react-table v8 state wiring, `flexRender`, the Radix tri-state Checkbox, and react-query `isPending`/`isError` are all correct. Three seeded edge cases — selection persisting across pages (intended bulk-select), `Math.max(getPageCount(), 1)` when filtered to 0, and `ContactsEmpty` vs in-table "No results." — are handled, not defects.

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `apps/console/src/modules/crm/contacts-columns.tsx` | `toast.success` fires unconditionally — reports success even when `navigator.clipboard` is undefined (insecure context) or `writeText` rejects. |
| 2 | ⚠️ Minor | `apps/console/src/components/data-table.tsx` | `TValue` generic resolves to `unknown` at every call site — phantom param. `DataTable<TData>` + `ColumnDef<TData>[]` reads the same. |

---

apps/console/src/modules/crm/contacts-columns.tsx
```diff
@@ -0,0 +1,195 @@
...
+    id: 'actions',
+    cell: ({ row }) => <RowActions contact={row.original} />,
+    enableSorting: false,
+  },
+];
+
+function RowActions({ contact }: { contact: Contact }) {
+  const copy = (label: string, value: string) => {
+    void navigator.clipboard?.writeText(value);
+    toast.success(`${label} copied`);
```
examlly-tech: **Minor — success toast on failure:** `toast.success` fires unconditionally, but the write is a no-op when `navigator.clipboard` is undefined and the `void` swallows a rejected `writeText`. Toast inside `.then()` (and guard on clipboard availability) so failure doesn't report success.

apps/console/src/components/data-table.tsx
```diff
@@ -0,0 +1,167 @@
...
+ * `@tanstack/react-table` wired over the Nexus `Table` primitives, with
+ * client-side sorting, single-column filtering, row selection, and paging.
+ * Domain modules supply the column definitions (where the cell/header JSX,
+ * sort triggers, and selection checkboxes live); this stays data-agnostic so
+ * every workspace module reuses it.
+ *
+ * Sort/filter/select/paginate only — column reorder/visibility/virtualization
+ * stay out until a consumer actually needs them.
+ */
+interface DataTableProps<TData, TValue> {
```
examlly-tech: **Minor — phantom generic:** `TValue` resolves to `unknown` everywhere (call site passes `ColumnDef<Contact>[]`). `DataTable<TData>` with `ColumnDef<TData>[]` reads identically. Idiomatic shadcn signature, so droppable rather than wrong.

aishvaryagupta:
All five review findings addressed in `5d32ff2` (every one fixed in-PR per `no-follow-up-deferral.md`):

| Finding | Resolution |
| --- | --- |
| **Architect — selection footer always renders** (`data-table.tsx`) | Footer now gates on `columns.some(c => c.id === 'select')` — shows "X of Y selected" only when a select column is wired, else a plain "Y row(s)". A Phase-3 table without selection won't show a permanent "0 of N selected". |
| **Architect — implicit page size** (`data-table.tsx`) | Added an explicit `pageSize` prop (default 10) → `initialState.pagination.pageSize`. No longer rides TanStack's implicit default. |
| **Architect — envelope discarded** (`crm-api.ts`) | `fetchContacts` now returns the `{ contacts }` envelope (route reads `data?.contacts`), mirroring the auth-api wrappers; a future server-side `totalCount` is now additive. |
| **SDE2 — success toast on clipboard failure** (`contacts-columns.tsx`) | `copy()` guards `navigator.clipboard` and toasts on the write's `.then()`/`.catch()`, so an unavailable/failed copy reports an error, not success. |
| **SDE2 — phantom `TValue` generic** (`data-table.tsx`) | Dropped to `DataTable<TData>` + `ColumnDef<TData>[]`. |

Browser re-verified: table renders, footer "0 of 24 row(s) selected.", "Page 1 of 3" (explicit pageSize). `typecheck` + ESLint clean. Both verdicts were non-blocking (COMMENT).

PR #330, opened 2026-06-02
feat(console): CRM contact detail page (Phase 2b)

aishvaryagupta:
## Summary

Phase 2b of the CRM flagship — the **record detail page** at `/m/crm/$id`, the read-only slice after the Contacts table (#283). Next: 2c create/edit · 2d kanban · 2e saved views.

- **Detail page** (`modules/crm/contact-detail-route.tsx`) — an action header (back link · contact name · status badge · avatar) over **Overview** (a Details card) and **Activity** (a timeline) tabs, with loading **Skeleton** and **not-found** states.
- **Data** — `ContactDetail` (`Contact` + `activity[]`) + `fetchContact(id)`; MSW `GET /api/crm/contacts/:id` finds the contact and **synthesises a deterministic activity feed** (offsets from `lastContacted`), 404 when the id is unknown.
- **Shared presentation** (`modules/crm/contact-ui.tsx`) — `ContactStatusBadge`, `initials`, and the currency/date formatters extracted from the 2a columns and reused by the detail (de-dupes the columns file).
- **Table wiring** — the name cell and a "View details" action link to `/m/crm/$id` (a static detail route; the sidebar already highlights `/m/crm/` by pathname).
- **query-client `retry: false`** — the mock API is deterministic, so error states (unknown id) render instantly instead of after react-query's backoff.

## Test Plan

Walked end-to-end in-browser:

- [x] Table name (or "View details") → `/m/crm/$id`; detail renders (header, status badge, avatar = contact initials)
- [x] **Overview** tab: Email / Owner / Open value / Last contacted
- [x] **Activity** tab: 4-item synthesised timeline, newest first, formatted dates
- [x] Unknown id (`/m/crm/does-not-exist`) → instant "Contact not found" (single 404, no retry storm)
- [x] Sidebar stays highlighted on CRM at the detail route
- [x] 0 console errors/warnings on happy paths; `typecheck` + ESLint clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #334, opened 2026-06-02
feat(console): CRM kanban board view (Phase 2d)

aishvaryagupta:
## Summary

Phase 2d of the CRM flagship — a **kanban board** view of contacts, toggled from the table via a `?view=table|board` search param. Follows 2a–2c. Next: 2e saved views/filters.

- **`ContactsBoard`** — `@dnd-kit` `DndContext` (Pointer + Keyboard sensors) with three droppable status columns (Lead / Active / Churned) and draggable contact cards. Dropping a card in another column changes that contact's **status optimistically** (`onMutate` updates the cache so the card moves immediately; the PATCH + `invalidate` reconcile in the background, with rollback `onError`).
- **View switcher** — a Table | Board toggle in the header drives the `view` search param (`validateSearch` zod enum, default `table`); the body swaps `DataTable` ↔ board. Linkable (`/m/crm?view=board`).
- Adds `@dnd-kit/core` + `@dnd-kit/utilities` (app deps — the `@nexus/react` bundle-size gate is unaffected).

## Test Plan

Walked in-browser:

- [x] Board renders three columns with correct card distribution (Lead 8 / Active 12 / Churned 4 = 24)
- [x] Switcher toggles Board ↔ Table via the URL (`?view=board` / `?view=table`); the body swaps
- [x] Status-change PATCH path verified (the data path `onDragEnd` uses)
- [x] 0 console errors/warnings; `typecheck` + ESLint clean

**Note on drag verification:** the drag *gesture* can't be driven by the test harness's synthetic pointer events (same limitation as Radix triggers in this environment) — a real pointer or the keyboard sensor performs the drag. The board rendering, the `onDragEnd → updateContact` wiring, and the PATCH-status endpoint are all verified; the optimistic cache update is the standard react-query pattern.

### DIFF
--- a/apps/console/src/app/router.tsx
+++ b/apps/console/src/app/router.tsx
@@ -84,6 +84,11 @@ const appearanceRoute = createRoute({
 const crmContactsRoute = createRoute({
   getParentRoute: () => appRoute,
   path: '/m/crm',
+  // `.default` (not `.catch`) keeps the param optional for navigation, so the
+  // existing `<Link to="/m/crm">` call sites don't need to pass `search`.
+  validateSearch: z.object({
+    view: z.enum(['table', 'board']).default('table'),
+  }),
   component: ContactsRoute,
 });
 
--- a/apps/console/src/app/router.tsx
+++ b/apps/console/src/app/router.tsx
@@ -84,10 +84,11 @@ const appearanceRoute = createRoute({
 const crmContactsRoute = createRoute({
   getParentRoute: () => appRoute,
   path: '/m/crm',
-  // `.default` (not `.catch`) keeps the param optional for navigation, so the
-  // existing `<Link to="/m/crm">` call sites don't need to pass `search`.
+  // `.default` keeps the param optional for navigation (existing `<Link
+  // to="/m/crm">` calls don't pass `search`); `.catch` also recovers an invalid
+  // value (e.g. a stale `?view=foo`) to 'table' instead of throwing a search error.
   validateSearch: z.object({
-    view: z.enum(['table', 'board']).default('table'),
+    view: z.enum(['table', 'board']).default('table').catch('table'),
   }),
   component: ContactsRoute,
 });
--- a/apps/console/src/modules/crm/contacts-board.tsx
+++ b/apps/console/src/modules/crm/contacts-board.tsx
@@ -15,6 +15,7 @@ import { useMutation, useQueryClient } from '@tanstack/react-query';
 import {
   type Contact,
   type ContactStatus,
+  crmKeys,
   updateContact,
 } from '../../lib/crm-api';
 import { formatCurrency } from '../../lib/format';
@@ -28,7 +29,6 @@ const COLUMNS: { status: ContactStatus; label: string }[] = [
 ];
 
 type ContactsCache = { contacts: Contact[] };
-const CONTACTS_KEY = ['crm', 'contacts'];
 
 /**
  * Kanban view of the contacts: three status columns, draggable cards. Dropping a
@@ -59,9 +59,11 @@ export function ContactsBoard({ contacts }: { contacts: Contact[] }) {
         value: contact.value,
       }),
     onMutate: async ({ contact, status }) => {
-      await queryClient.cancelQueries({ queryKey: CONTACTS_KEY });
-      const previous = queryClient.getQueryData<ContactsCache>(CONTACTS_KEY);
-      queryClient.setQueryData<ContactsCache>(CONTACTS_KEY, (old) =>
+      await queryClient.cancelQueries({ queryKey: crmKeys.contacts });
+      const previous = queryClient.getQueryData<ContactsCache>(
+        crmKeys.contacts
+      );
+      queryClient.setQueryData<ContactsCache>(crmKeys.contacts, (old) =>
         old
           ? {
               contacts: old.contacts.map((c) =>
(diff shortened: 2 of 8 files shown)


apps/console/src/app/router.tsx
```diff
@@ -84,6 +84,11 @@ const appearanceRoute = createRoute({
 const crmContactsRoute = createRoute({
   getParentRoute: () => appRoute,
   path: '/m/crm',
+  // `.default` (not `.catch`) keeps the param optional for navigation, so the
+  // existing `<Link to="/m/crm">` call sites don't need to pass `search`.
+  validateSearch: z.object({
+    view: z.enum(['table', 'board']).default('table'),
```
examlly-tech: **Minor — invalid-value crash:** `.default('table')` recovers a missing param, but `?view=foo` fails the enum parse → TanStack throws `VALIDATE_SEARCH`, halting the page. `.default('table').catch('table')` defaults missing *and* recovers invalid, while keeping the param optional for `<Link to="/m/crm">`.

apps/console/src/modules/crm/contacts-board.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { initials } from './contact-ui';
+
+const COLUMNS: { status: ContactStatus; label: string }[] = [
+  { status: 'lead', label: 'Lead' },
+  { status: 'active', label: 'Active' },
+  { status: 'churned', label: 'Churned' },
+];
+
+type ContactsCache = { contacts: Contact[] };
+const CONTACTS_KEY = ['crm', 'contacts'];
```
examlly-tech: **Minor — duplicated query-key contract:** `CONTACTS_KEY` here and the route's inline `['crm','contacts']` (contacts-route.tsx:24) are declared independently; the optimistic write breaks silently if they drift. Export one shared key (e.g. `crmKeys` in `crm-api.ts`) and consume it in both.

examlly-tech (commented):
## 🤖 Principal Architect Review — #334

**Verdict: NEEDS DISCUSSION** — minor, non-blocking.

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `apps/console/src/app/router.tsx` | `.default('table')` recovers a *missing* `view` but throws `VALIDATE_SEARCH` on an *invalid* one (`?view=foo`), halting the page. |
| 2 | ⚠️ Minor | `apps/console/src/modules/crm/contacts-board.tsx` | The contacts query key is declared independently in the board (`CONTACTS_KEY`) and inline in the route's `useQuery` — they can drift, and the optimistic write depends on them matching. |

### Architectural Concerns

**Flat CRM routing stays correct** — Table and Board are the same route (`?view=` swaps the body inside `ContactsRoute`); the switcher is chrome within one component and the detail route doesn't use it, so the cross-sibling-chrome condition from #330 is still unmet. No layout parent needed.

**Optimistic strategy is sound** — cancel + snapshot + rollback + blanket `['crm']` invalidate is the correct v5 pattern and flicker-free (`cancelQueries` aborts the in-flight refetch; the full-field PATCH matches the optimistic write, so `onSettled` returns identical data).

---

apps/console/src/modules/crm/contacts-board.tsx
```diff
@@ -0,0 +1,178 @@
...
+          : old
+      );
+      return { previous };
+    },
+    onError: (_error, _vars, context) => {
+      if (context?.previous) {
+        queryClient.setQueryData(CONTACTS_KEY, context.previous);
+      }
+    },
+    onSettled: () => queryClient.invalidateQueries({ queryKey: ['crm'] }),
```
examlly-tech: **Minor — third inline key:** `onSettled` invalidates a raw `['crm']` while `onMutate`/`onError` two lines up use the named `CONTACTS_KEY`. Inconsistent within one mutation; derive both from the shared key set.

examlly-tech (commented):
## 🤖 SDE2 Review — #334

**Verdict: MINOR CHANGES** — non-blocking.

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `apps/console/src/modules/crm/contacts-board.tsx` | `onSettled` invalidates a third inline `['crm']` literal while `onMutate`/`onError` use the named `CONTACTS_KEY` — inconsistent within one mutation; derive both from a shared key. |

Verified sound: the `UniqueIdentifier` (active/over.id) vs `ContactStatus`/id comparisons, the optimistic-update mechanics (cancel → snapshot → updater handles `undefined` → rollback → onSettled invalidate), @dnd-kit usage (sensors, `CSS.Translate`, listeners/attributes spread), and all drag-end guards (drop-outside, unknown column, missing contact, same-column no-op). `typecheck` clean.

---

examlly-tech (approved):
## ✅ Follow-up — both findings addressed (`953a703`)

**Invalid `?view` (Architect): FIXED.** `validateSearch` is now `z.enum(['table','board']).default('table').catch('table')` — `.default` keeps the param optional for `<Link to="/m/crm">`, `.catch` recovers an invalid value. Verified: `?view=foo` normalises to the table (no crash), board + table both render.

**Duplicated query keys (Architect + SDE2): FIXED.** Exported a `crmKeys` set (`all` / `contacts` / `contact(id)`) from `crm-api.ts` and consumed it in the table, board (including the `onSettled` invalidate), detail, and the create/edit mutation — no inline `['crm'…]` literals remain. Consolidated all four CRM files (not just the two this PR touched) to avoid a half-migrated key set.

`typecheck` + ESLint clean.

PR #338, opened 2026-06-02
feat(console): CRM saved views + status filter (Phase 2e)

aishvaryagupta:
## Summary

Phase 2e — the **Phase-2 capstone**: faceted **filtering** + **saved views** on the Contacts list. Completes the CRM flagship (table → detail → create/edit → kanban → views/filters).

- **Status facet filter** (table view) — a multi-select **Status** dropdown drives a `status` search param. The view/search contract now lives in `contacts-search.ts`: a single zod schema (`contactsSearchSchema`) that the router validates against and that `ContactsView` derives from via `z.infer`, with `CONTACT_STATUSES` as the one source for the status values. Table rows are pre-filtered; the facet is table-only since the board is already organised by status (switching to board clears the filter so it never persists invisibly).
- **Saved views** — a **Views** popover lists user-saved view+filter combos (Zustand, persisted to localStorage), applies one by navigating to its search params, and deletes per row. "Save current view…" names the live view in a dialog. The view/status state lives entirely in the URL, so views are shareable.
- Extracted **`ContactsToolbar`** (view switcher + status filter + views popover).

## Review fixes (b2ec2c8)

Addresses both bot reviews:

- **a11y blocker (SDE2)** — the per-row delete was a `<button>` nested in a `DropdownMenuItem`, unreachable by Radix roving focus and tripping axe `nested-interactive`. Rebuilt `SavedViewsMenu` as a **Popover** whose rows carry sibling apply/delete buttons, both Tab-reachable.
- **source of truth (Architect)** — `ContactsView` hand-duplicated the route's zod shape; now derived via `z.infer` from the shared `contactsSearchSchema`, with `CONTACT_STATUSES` single-sourcing the values (route facet + form enum both reference it).
- **stale board filter (SDE2)** — `setSearch` clears `status` when `view==='board'`.
- _Declined with rationale:_ persist `version`/`migrate` (`project-stage.md` forbids migration theater; `.catch([])` degrades a stale value gracefully) and pushing the filter to the query layer (would add latency + fork the shared `crmKeys.contacts` cache the board's optimistic writes depend on; an inline `select` re-runs each render too; 24-row scale).

## Test Plan (browser-verified)

- [x] `?status=["active"]` → table filtered to the 12 active contacts (Page 1 of 2); the **Status** button shows a count badge
- [x] `?status=foo` (invalid) → recovers to no filter, no crash
- [x] **Save** a view → persisted to localStorage as `{view, status}`; dialog closes
- [x] **Apply** a saved view → navigates to its search params (verified via keyboard)
- [x] **Delete** a saved view is keyboard-reachable — Tab lands on the delete control, Enter removes it (the a11y fix)
- [x] Switching to **Board** clears the status filter (`status=[]`)
- [x] `typecheck`, `eslint --max-warnings 0`, `prettier --check` all clean

### DIFF
--- a/apps/console/src/modules/crm/contacts-route.tsx
+++ b/apps/console/src/modules/crm/contacts-route.tsx
@@ -11,8 +11,8 @@ import { crmKeys, fetchContacts } from '../../lib/crm-api';
 import { ContactFormSheet } from './contact-form-sheet';
 import { ContactsBoard } from './contacts-board';
 import { contactColumns } from './contacts-columns';
+import type { ContactsView } from './contacts-search';
 import { ContactsToolbar } from './contacts-toolbar';
-import type { ContactsView } from './saved-views';
 
 const crmRoute = getRouteApi('/app/m/crm');
 
@@ -27,7 +27,15 @@ export function ContactsRoute() {
   const [createOpen, setCreateOpen] = useState(false);
 
   const setSearch = (patch: Partial<ContactsView>) =>
-    navigate({ search: (prev) => ({ ...prev, ...patch }) });
+    navigate({
+      search: (prev) => {
+        const next = { ...prev, ...patch };
+        // The status facet is table-only — the board is organised by status, so
+        // never carry a hidden status filter into board view (incl. when applying
+        // a saved board view).
+        return next.view === 'board' ? { ...next, status: [] } : next;
+      },
+    });
 
   return (
     <div className="nx:space-y-6 nx:p-6">
--- a/apps/console/src/modules/crm/contacts-route.tsx
+++ b/apps/console/src/modules/crm/contacts-route.tsx
@@ -1,12 +1,7 @@
 import { useState } from 'react';
 
 import { Button, Skeleton } from '@nexus/react';
-import {
-  IconLayoutKanban,
-  IconPlus,
-  IconTable,
-  IconUsers,
-} from '@tabler/icons-react';
+import { IconPlus, IconUsers } from '@tabler/icons-react';
 import { useQuery } from '@tanstack/react-query';
 import { getRouteApi } from '@tanstack/react-router';
 
@@ -16,6 +11,8 @@ import { crmKeys, fetchContacts } from '../../lib/crm-api';
 import { ContactFormSheet } from './contact-form-sheet';
 import { ContactsBoard } from './contacts-board';
 import { contactColumns } from './contacts-columns';
+import { ContactsToolbar } from './contacts-toolbar';
+import type { ContactsView } from './saved-views';
 
 const crmRoute = getRouteApi('/app/m/crm');
 
@@ -25,9 +22,12 @@ export function ContactsRoute() {
     queryFn: fetchContacts,
   });
   const contacts = data?.contacts;
-  const { view } = crmRoute.useSearch();
+  const { view, status } = crmRoute.useSearch();
   const navigate = crmRoute.useNavigate();
   const [createOpen, setCreateOpen] = useState(false);
+
+  const setSearch = (patch: Partial<ContactsView>) =>
+    navigate({ search: (prev) => ({ ...prev, ...patch }) });
 
   return (
     <div className="nx:space-y-6 nx:p-6">
@@ -46,24 +46,7 @@ export function ContactsRoute() {
         </Button>
       </header>
 
-      <div className="nx:inline-flex nx:gap-1">
-        <Button
-          variant={view === 'table' ? 'default' : 'ghost'}
-          size="sm"
-          onClick={() => navigate({ search: { view: 'table' } })}
-        >
-          <IconTable />
-          Table
-        </Button>
-        <Button
-          variant={view === 'board' ? 'default' : 'ghost'}
-          size="sm"
-          onClick={() => navigate({ search: { view: 'board' } })}
-        >
-          <IconLayoutKanban />
-          Board
-        </Button>
-      </div>
+      <ContactsToolbar view={view} status={status} setSearch={setSearch} />
 
       {isPending && <ContactsSkeleton />}
       {isError && (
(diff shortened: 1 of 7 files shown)


examlly-tech (commented):
## 🤖 Principal Architect Review — #338 (Phase 2e)

**Verdict: NEEDS DISCUSSION**

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Design | `saved-views.ts` / `router.tsx` | `ContactsView` hand-duplicates the router's zod search shape — two sources of truth for the same `view`+`status` contract |
| 2 | ⚠️ Minor | `saved-views.ts` | `persist` declares no `version`/`migrate`; persisted views hold stale enum values across a future schema change |
| 3 | ⚠️ Minor | `contacts-route.tsx` | Status filter runs client-side in the render path |

### Architectural Concerns

**Type source-of-truth.** The router's zod search schema and the `ContactsView`/`SavedView` types independently declare the same `view` union + `status[]`. Export the schema from a shared module and derive `ContactsView` via `z.infer<>` so the two cannot drift — `code-comments.md` requires sync enforced at compile time, not by the "mirrors the route's search params" comment. This also resolves the `setSearch(patch: Partial<ContactsView>)` boundary: the patch type then references the inferred search type rather than a store type.

---

apps/console/src/modules/crm/saved-views.ts
```diff
@@ -0,0 +1,42 @@
+import { create } from 'zustand';
+import { persist } from 'zustand/middleware';
+
+import type { ContactStatus } from '../../lib/crm-api';
+
+/** The shareable Contacts view state — mirrors the route's search params. */
+export type ContactsView = {
```
examlly-tech: **Dual source of truth.** `ContactsView` hand-duplicates the router schema's `view`/`status` shape. Export the zod schema (shared module) and derive this type with `z.infer<>`; `code-comments.md` bans keep-in-sync-by-comment and requires compile-time enforcement. Also resolves the `Partial<ContactsView>` patch-type boundary in `contacts-route.tsx`.

apps/console/src/modules/crm/saved-views.ts
```diff
@@ -0,0 +1,42 @@
...
+  save: (name: string, view: ContactsView) => void;
+  remove: (id: string) => void;
+};
+
+/**
+ * User-saved Contacts views (a named view + status-filter combo), persisted to
+ * localStorage. Applying one navigates the table to its search params; this
+ * store only owns the saved list, not the live view (which lives in the URL).
+ */
+export const useSavedViews = create<SavedViewsState>()(
```
examlly-tech, line 23: **No migration guard.** `persist` declares no `version`. A future status-enum change leaves stale localStorage values that `.catch([])` silently drops. Consider `version: 1` + a `migrate` that strips unknown values — weigh against the pre-production stage (`project-stage.md`).

apps/console/src/modules/crm/contacts-route.tsx
```diff
@@ -79,7 +62,11 @@ export function ContactsRoute() {
         ) : (
           <DataTable
             columns={contactColumns}
-            data={contacts}
+            data={
+              status.length > 0
+                ? contacts.filter((c) => status.includes(c.status))
```
examlly-tech, line 75: **Filter in render path.** `contacts.filter((c) => status.includes(c.status))` recomputes each render. At 24 rows this is negligible, but consider whether filtered data belongs at the query layer (`select`) — weigh against the shared `crmKeys.contacts` cache the board's optimistic writes depend on.

apps/console/src/modules/crm/contacts-toolbar.tsx
```diff
@@ -0,0 +1,236 @@
...
+            <DropdownMenuItem disabled>No saved views yet</DropdownMenuItem>
+          ) : (
+            views.map((saved) => (
+              <DropdownMenuItem
+                key={saved.id}
+                onSelect={() => setSearch(saved.view)}
+                className="nx:flex nx:items-center nx:justify-between nx:gap-2"
+              >
+                <span className="nx:truncate">{saved.name}</span>
+                <Button
```
examlly-tech: **Blocking — keyboard-unreachable delete.** Radix roving focus moves between `role="menuitem"` elements only; this nested `<button>` isn't one, so keyboard users can apply a view but never delete one (Tab closes the menu), and it trips axe `nested-interactive`. The mouse path (`stopPropagation` on `onClick`+`onPointerDown`) is correct — the gap is purely keyboard. Move delete out of the menu-item (e.g. a Popover panel with sibling buttons).

apps/console/src/modules/crm/contacts-toolbar.tsx
```diff
@@ -0,0 +1,236 @@
...
+        </DropdownMenuTrigger>
+        <DropdownMenuContent align="end" className="nx:w-56">
+          <DropdownMenuLabel>Saved views</DropdownMenuLabel>
+          {views.length === 0 ? (
+            <DropdownMenuItem disabled>No saved views yet</DropdownMenuItem>
+          ) : (
+            views.map((saved) => (
+              <DropdownMenuItem
+                key={saved.id}
+                onSelect={() => setSearch(saved.view)}
```
examlly-tech: **Minor — stale filter in board view.** Applying a saved view with `view:'board'` while `status` is set leaves the filter live in the URL, but `StatusFilter` isn't rendered in board view and `ContactsBoard` ignores `status` — so it's silent and unclearable. Clear `status` when entering board view.

examlly-tech (changes requested):
## 🤖 SDE2 Review — #338 (Phase 2e)

**Verdict: CHANGES REQUIRED**

### Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `contacts-toolbar.tsx` | Per-row delete `<Button>` nested in a `DropdownMenuItem` is keyboard-unreachable and trips axe `nested-interactive` |
| 2 | ⚠️ Minor | `contacts-toolbar.tsx` | Applying a `view:'board'` saved view leaves `status` live in the URL but unclearable — the filter UI isn't rendered in board view |

---

PR #348, opened 2026-06-02
feat(console): Inbox module — conversation list + thread + reply (Phase 3b)

aishvaryagupta:
## Summary

Phase 3b — the second **breadth** module: **Inbox / Support**, a two-pane mail-client layout (conversation list ⟷ thread) backed by MSW mocks. Where Projects reused the CRM table→detail→form record-loop, Inbox dogfoods a **new interaction shape**: a master/detail split driven by a search param, a reply composer, and in-place status changes — exercising `ScrollArea`, `DropdownMenu` (radio group), `Textarea`, `Avatar`, `Card`, and `Badge` together.

- **`lib/inbox-api.ts`** — typed client with a deliberate **list/detail payload split**: `Conversation` (lean preview row — `preview` + `lastMessageAt` derived from the last message) vs `ConversationDetail` (full `messages[]` thread). Single-sourced `CONVERSATION_STATUSES` (`as const` → union type + the status menu); `inboxKeys` query-key object.
- **Two-pane route at `/m/inbox`** — the open conversation rides in **`?c`** (no nested Outlet). Bounded to `calc(100svh − topbar)` so each pane scrolls internally. **Stacks on mobile** (full-width list ⟷ full-width thread) via `lg:` viewport prefixes — Atlas's first genuinely mobile-stacking route.
- **Reply** (plain `useState` composer + mutation; ⌘↵ to send) and **status change** (radio `DropdownMenu`) both **invalidate the thread *and* the list** — a reply changes the list's preview/time/unread/order; a status change updates both badges.
- **`key={c}`** remounts the thread on conversation switch so a drafted-but-unsent reply can't bleed into the next conversation's composer (cached data still renders immediately — no skeleton flash).
- **8 deterministic seed conversations** spanning every status (open/pending/closed) and both unread states; dated into the recent past (newest ~Jun 1, matching the CRM/Projects ceiling) so a reply sent now bumps its thread to the top.
- **MSW handlers** — list (projects to the preview shape, sorted newest-first), detail (404 on miss), reply (append an agent message + clear `unread`), status `PATCH`.
- **`format.ts`** — `formatDateTime` helper (UTC-pinned, like `formatDate`) for message timestamps.
- `route` added to the Inbox module-registry entry → Inbox is no longer "coming soon".

**Scope (advisor-validated):** assignee reassignment, canned replies, search/filter, and bulk actions are deferred — they add product surface, not new component coverage. Reply is **invalidate-on-success**, not optimistic (a 300 ms mock round-trip doesn't warrant the rollback complexity).

## Test Plan (browser-verified — desktop + mobile, light + dark)

- [x] Sidebar **Inbox** links to `/m/inbox` and renders the two-pane, not the "coming soon" placeholder
- [x] List: 8 conversations, newest-first; unread dot on exactly the unread rows; bold name when unread; subject/preview truncate; status badge per row
- [x] Click a row → `?c=<id>`; thread shows header (subject · customer · email), status badge + **Status** menu, message bubbles (customer left / agent right, primary-subtle), composer
- [x] **Reply** → message appends, composer clears, unread clears, list preview/time update, and the thread **bumps to the top** of the list
- [x] **Draft-bleed guard** → typing a draft then switching to an already-cached conversation shows an **empty** composer (the `key={c}` fix)
- [x] **Status change** → header badge + list-row badge both update (dual invalidation); toast shows the human label ("Marked as Closed")
- [x] **Mobile (390px)** → list and thread stack full-width (no horizontal overflow after the `min-w-0` fix); **dark mode** adapts via semantic tokens
- [x] `typecheck`, `eslint --max-warnings 0`, `prettier --check` clean; console clean through all interactions

### DIFF
--- a/apps/console/src/app/router.tsx
+++ b/apps/console/src/app/router.tsx
@@ -112,7 +112,8 @@ const projectsIssueDetailRoute = createRoute({
 });
 
 // Inbox (Phase 3b). Static `/m/inbox` outranks `/m/$module`. The open
-// conversation rides in `?c`; an absent/unknown id falls back to the empty pane.
+// conversation rides in `?c` — absent shows the empty pane; an unknown id
+// surfaces the thread's error state.
 const inboxRoute = createRoute({
   getParentRoute: () => appRoute,
   path: '/m/inbox',
--- a/apps/console/src/app/router.tsx
+++ b/apps/console/src/app/router.tsx
@@ -17,6 +17,7 @@ import { contactsSearchSchema } from '../modules/crm/contacts-search';
 import { AppearanceRoute } from '../modules/design-system/appearance-route';
 import { ReferenceRoute } from '../modules/design-system/reference-route';
 import { ScenesRoute } from '../modules/design-system/scenes-route';
+import { InboxRoute } from '../modules/inbox/inbox-route';
 import { IssueDetailRoute } from '../modules/projects/issue-detail-route';
 import { IssuesRoute } from '../modules/projects/issues-route';
 
@@ -110,6 +111,15 @@ const projectsIssueDetailRoute = createRoute({
   component: IssueDetailRoute,
 });
 
+// Inbox (Phase 3b). Static `/m/inbox` outranks `/m/$module`. The open
+// conversation rides in `?c`; an absent/unknown id falls back to the empty pane.
+const inboxRoute = createRoute({
+  getParentRoute: () => appRoute,
+  path: '/m/inbox',
+  validateSearch: z.object({ c: z.string().optional() }),
+  component: InboxRoute,
+});
+
 // One shared placeholder backs every not-yet-built module.
 const moduleRoute = createRoute({
   getParentRoute: () => appRoute,
@@ -162,6 +172,7 @@ const routeTree = rootRoute.addChildren([
     crmContactDetailRoute,
     projectsIssuesRoute,
     projectsIssueDetailRoute,
+    inboxRoute,
     moduleRoute,
   ]),
   authRoute.addChildren([loginRoute, signupRoute, verifyRoute, forgotRoute]),
--- a/apps/console/src/lib/inbox-api.ts
+++ b/apps/console/src/lib/inbox-api.ts
@@ -1,0 +1,119 @@
+/**
+ * Inbox API client. Thin typed wrappers over the mock endpoints served by MSW
+ * (`src/mocks/handlers.ts`) — there is no real backend. Mirrors the CRM/Projects
+ * shape: a single query-key object, fetch/mutate fns, and an `as const` enum that
+ * single-sources both the union type and the status badge.
+ *
+ * The list and detail payloads deliberately differ: the list row carries a lean
+ * `preview` + `lastMessageAt`, while the thread carries the full `messages` array
+ * — so the conversation list never ships every message body.
+ */
+
+/**
+ * Conversation lifecycle statuses — the single source for the
+ * {@link ConversationStatus} union and the thread's status menu.
+ */
+export const CONVERSATION_STATUSES = ['open', 'pending', 'closed'] as const;
+
+/** Where a conversation sits in its lifecycle — rendered as a status badge. */
+export type ConversationStatus = (typeof CONVERSATION_STATUSES)[number];
+
+/** A single message in a conversation thread. */
+export type Message = {
+  id: string;
+  /** Who wrote it — drives bubble alignment and the avatar. */
+  author: 'customer' | 'agent';
+  authorName: string;
+  body: string;
+  /** ISO timestamp (date + time) — rendered with {@link formatDateTime}. */
+  at: string;
+};
+
+/** Fields shared by the list row and the thread header. */
+type ConversationBase = {
+  id: string;
+  customer: string;
+  customerEmail: string;
+  subject: string;
+  status: ConversationStatus;
+  /** Team member handling the conversation — also signs agent replies. */
+  assignee: string;
+  /** Whether the latest customer message is still unhandled — shown as a dot. */
+  unread: boolean;
+};
+
+/** List-row shape: the lean preview, without the message bodies. */
+export type Conversation = ConversationBase & {
+  /** The most recent message's body — the list preview (CSS-truncated). */
+  preview: string;
+  /** ISO timestamp of the most recent message — the list sorts + labels by it. */
+  lastMessageAt: string;
+};
+
+/** Thread shape: the full message history, without the list-only preview fields. */
+export type ConversationDetail = ConversationBase & {
+  messages: Message[];
+};
+
+/**
+ * TanStack Query keys for the Inbox module — declared once so the list, thread,
+ * and mutations can't drift apart (a mismatched key silently breaks cache
+ * invalidation).
+ */
+export const inboxKeys = {
+  all: ['inbox'] as const,
+  conversations: ['inbox', 'conversations'] as const,
+  conversation: (id: string) => ['inbox', 'conversation', id] as const,
+};
+
+export async function fetchConversations(): Promise<{
+  conversations: Conversation[];
+}> {
+  const res = await fetch('/api/inbox/conversations');
+  if (!res.ok) {
+    throw new Error('Failed to load conversations. Please try again.');
+  }
+  return (await res.json()) as { conversations: Conversation[] };
+}
+
+export async function fetchConversation(
+  id: string
+): Promise<{ conversation: ConversationDetail }> {
+  const res = await fetch(`/api/inbox/conversations/${id}`);
+  if (!res.ok) {
+    throw new Error('Failed to load conversation.');
+  }
+  return (await res.json()) as { conversation: ConversationDetail };
+}
+
+/** Post an agent reply — appends a message and returns the updated thread. */
+export async function replyToConversation(
+  id: string,
+  body: string
+): Promise<{ conversation: ConversationDetail }> {
+  const res = await fetch(`/api/inbox/conversations/${id}/reply`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ body }),
+  });
+  if (!res.ok) {
+    throw new Error('Failed to send reply. Please try again.');
+  }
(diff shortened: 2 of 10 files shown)


apps/console/src/modules/inbox/inbox-route.tsx
```diff
@@ -0,0 +1,97 @@
...
+          Customer conversations across every channel. Pick one to read and
+          reply.
+        </p>
+      </header>
+
+      <div className="nx:border-border-default nx:flex nx:min-h-0 nx:flex-1 nx:overflow-hidden nx:rounded-lg nx:border">
+        {/* List pane — full-width on mobile, fixed on desktop; hidden on mobile
+            once a thread is open so the thread gets the whole screen. */}
+        <aside
+          className={`nx:border-border-default nx:w-full nx:min-h-0 nx:min-w-0 nx:flex-col nx:lg:flex nx:lg:w-80 nx:lg:border-r ${c ? 'nx:hidden' : 'nx:flex'}`}
```
examlly-tech, line 39: **Blocking — no mobile back-affordance:** the `aside` hides when `c` is set, but `ThreadContent` has no control to clear `?c`. On Narrow (co-first-class) the user is stranded. Add a `<Link to="/m/inbox" search={{}}>` back button in the thread header, visible only below `nx:lg:`.

apps/console/src/app/router.tsx
```diff
@@ -110,6 +111,15 @@ const projectsIssueDetailRoute = createRoute({
   component: IssueDetailRoute,
 });
 
+// Inbox (Phase 3b). Static `/m/inbox` outranks `/m/$module`. The open
+// conversation rides in `?c`; an absent/unknown id falls back to the empty pane.
```
examlly-tech: **Minor — comment inaccurate:** says an unknown id falls back to the empty pane, but it renders the thread's not-found (404 path). Update to match actual behaviour.

apps/console/src/lib/inbox-api.ts
```diff
@@ -0,0 +1,119 @@
...
+  /** Who wrote it — drives bubble alignment and the avatar. */
+  author: 'customer' | 'agent';
+  authorName: string;
+  body: string;
+  /** ISO timestamp (date + time) — rendered with {@link formatDateTime}. */
+  at: string;
+};
+
+/** Fields shared by the list row and the thread header. */
+type ConversationBase = {
```
examlly-tech, line 33: **Minor — base type carries list-unused fields:** `customerEmail` and `assignee` live on `ConversationBase` but the list row renders neither. The split's goal is a lean list — move them to `ConversationDetail` only.

examlly-tech (commented):
## Principal Architect Review — CHANGES REQUIRED

### Issues

| #   | Severity    | File                                     | Issue                                                                                                                                                                                 |
| --- | ----------- | ---------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | ❌ Blocking | `modules/inbox/inbox-route.tsx:39`       | Mobile back-navigation stranded: the list `aside` hides when `?c` is set, but nothing in the thread clears it — on Narrow (co-first-class) the user can only return via browser-back. |
| 2   | ⚠️ Minor    | `modules/inbox/conversation-list.tsx:54` | Row `Link` should set `resetScroll={false}` so the list's scroll position survives a row click.                                                                                       |
| 3   | ⚠️ Minor    | `lib/inbox-api.ts:33`                    | `ConversationBase` carries `customerEmail` + `assignee`, which the list row renders neither — undermines the lean-list projection.                                                    |

### Architectural Concerns

**Mobile stranded-navigation** (`inbox-route.tsx:39–55`) — below `lg`, selecting a row hides the list with no affordance to return. Also `router.tsx:115` mis-states the fallback: an unknown `?c` renders the thread's not-found, not the empty pane.

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/348

apps/console/src/modules/inbox/conversation-thread.tsx
```diff
@@ -0,0 +1,248 @@
...
+    onError: (error: Error) => toast.error(error.message),
+  });
+
+  // Keep the latest message in view when the thread opens or grows — a sent
+  // reply lands below the fold on mobile otherwise.
+  useEffect(() => {
+    endRef.current?.scrollIntoView({ block: 'nearest' });
+  }, [conversation.messages.length]);
+
+  const handleSend = () => {
```
examlly-tech, line 88: **Minor — double-submit on ⌘↵:** `handleComposerKeyDown` calls `handleSend` directly, bypassing the Button's `disabled`/`loading`. Two ⌘↵ within the 300 ms round-trip submit twice. Add `if (replyMutation.isPending) return;` at the top of `handleSend`.

apps/console/src/modules/inbox/conversation-thread.tsx
```diff
@@ -0,0 +1,248 @@
...
+      <DropdownMenuTrigger asChild>
+        <Button variant="outline" size="sm" disabled={disabled}>
+          Status
+          <IconChevronDown />
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent align="end">
+        <DropdownMenuRadioGroup
+          value={status}
+          onValueChange={(value) => onChange(value as ConversationStatus)}
```
examlly-tech, line 197: **Minor — unsound cast:** Radix types `onValueChange` as `(value: string) => void`, so `as ConversationStatus` is required but unchecked. Runtime-safe here since items come from `CONVERSATION_STATUSES`; a short inline note makes the intent explicit.

apps/console/src/modules/inbox/conversation-list.tsx
```diff
@@ -0,0 +1,91 @@
...
+  conversations,
+  activeId,
+}: ConversationListProps) {
+  return (
+    <div className="nx:flex nx:min-h-0 nx:flex-1 nx:flex-col">
+      <div className="nx:border-border-default nx:text-muted-foreground nx:border-b nx:px-4 nx:py-3 nx:text-sm">
+        {conversations.length} conversation
+        {conversations.length === 1 ? '' : 's'}
+      </div>
+      {/* Plain overflow container, not ScrollArea: the rows truncate (nowrap),
```
examlly-tech: **Minor — inline tradeoff block:** per `code-comments.md`, why-not rationale belongs in the PR description, not a 5-line code comment. Trim to the load-bearing constraint.

apps/console/src/modules/inbox/conversation-thread.tsx
```diff
@@ -0,0 +1,248 @@
...
+        {Array.from({ length: 4 }, (_, i) => (
+          <Skeleton key={i} className="nx:h-16 nx:w-3/4" />
+        ))}
+      </div>
+    </div>
+  );
+}
+
+// App-local not-found — the polished @nexus/react EmptyState is tracked in #282.
+function ThreadNotFound() {
```
examlly-tech: **Minor — misleading name:** `ThreadNotFound` renders for any query error, not only 404. Rename to `ThreadError` or widen the copy to "Couldn't load this conversation".

examlly-tech (changes requested):
## SDE2 Review — MINOR CHANGES

### Issues

| #   | Severity | File                                        | Issue                                                                                                                                                                                      |
| --- | -------- | ------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1   | ⚠️ Minor | `modules/inbox/conversation-thread.tsx:87`  | `handleSend` has no `isPending` guard; ⌘↵ fires it directly from the Textarea, bypassing the Button's `disabled`/`loading` — two quick presses double-submit within the 300 ms round-trip. |
| 2   | ⚠️ Minor | `modules/inbox/conversation-thread.tsx:177` | `value as ConversationStatus` casts an unvalidated `string`; runtime-safe (items come from `CONVERSATION_STATUSES`) but the cast hides the gap.                                            |
| 3   | ⚠️ Minor | `modules/inbox/conversation-list.tsx:24`    | Multi-line "why not ScrollArea" block is tradeoff rationale — `code-comments.md` routes that to the PR description, not inline.                                                            |
| 4   | ⚠️ Minor | `modules/inbox/conversation-thread.tsx:237` | `ThreadNotFound` renders for any fetch error, not only 404; the name over-claims.                                                                                                          |

### Links

- **PR:** https://github.com/nexuslabs-ai/nexus/pull/348

aishvaryagupta:
**Note on the conversation-list scroll container (why a plain `overflow-y-auto` div, not `ScrollArea`)** — moved here from an inline comment per `code-comments.md` (tradeoff rationale → PR, not code).

The list rows truncate (`nowrap` + `truncate`). Radix `ScrollArea`'s viewport wraps content in a `display:table` element, which lets `nowrap` children expand the row to their natural width instead of clipping — so the rows overflowed horizontally and the viewport scrolled sideways, hiding the customer name/subject. A plain block `overflow-y-auto` container gives the rows a definite width, so `min-w-0` + `truncate` clip correctly.

The thread pane keeps `ScrollArea` because its bubbles **wrap** (`whitespace-pre-wrap` + `max-w`) rather than truncate, so the table-layout quirk doesn't bite there — and its message list has no focusable children, so it benefits from `ScrollArea`'s built-in `tabIndex={0}` viewport (axe `scrollable-region-focusable`). The list rows are `<Link>`s (already focusable), so the plain div stays keyboard-reachable without a manual `tabIndex`.

examlly-tech (approved):
## Principal Architect Follow-up — APPROVED

### Unresolved Issues

All previous issues resolved — mobile back-link added, `customerEmail`/`assignee` moved to `ConversationDetail` and projected out of the list endpoint, `resetScroll={false}` added, router `?c` comment corrected.

### New Issues

None.

examlly-tech (approved):
## SDE2 Follow-up — APPROVED

### Unresolved Issues

All previous issues resolved — `isPending` guard added to `handleSend`, status-cast soundness documented, ScrollArea comment trimmed (full rationale in a PR comment), `ThreadNotFound` → `ThreadError` with error-agnostic copy.

### New Issues

None blocking. The thread's `nx:lg:hidden` back-link uses a viewport prefix, but it mirrors the route's page-shell list-pane hide at `lg` and is documented inline — the responsive.md viewport-exception applies. No change needed.

PR #371, opened 2026-06-03
feat(console): resizable inbox split-pane on desktop

aishvaryagupta:
## Summary

Makes the inbox **list ↔ thread** split drag-resizable on desktop, using the Nexus `Resizable*` components. Adds `react-resizable-panels@^4.11.2` as a console dependency (it's an optional peer of `@nexus/react`, so consumers install it themselves; the lockfile already had the resolution so `--frozen-lockfile` is unaffected).

### Design decisions (made explicit)

- **Desktop-only affordance.** Resize applies at ≥`lg` (64rem). Below `lg` the inbox keeps its existing single-pane toggle (list **or** thread, driven by the selected conversation). The two layouts are **conditionally rendered** via a small `useMediaQuery` hook — a pure-CSS dual render would mount the thread/composer twice.
- **Shared panes.** The list/thread panes are extracted to shared JSX so neither branch duplicates them; only the active branch mounts (no double-fetch / double-composer).
- **Ephemeral by design.** `react-resizable-panels` v4 has no `autoSaveId`, and the default **28 / 72** split is a good resting state, so the split resets to it on reload rather than carrying manual `localStorage` persistence.
- New `apps/console/src/hooks/use-media-query.ts` — `useSyncExternalStore`-based, reads the match synchronously (no first-render flash).

## GitHub Issue

No issue to close — the Resizable **component** (#294) already shipped in the component batch; this wires it into the console (the one Tier-1+Resizable item the user greenlit that needs a new peer dep).

## Test Plan

- [x] `@nexus/console` typecheck clean (validated against the real v4 `orientation` API)
- [x] `eslint . --max-warnings 0` clean
- [x] `prettier --check` clean
- [x] `@nexus/console` vite build clean — `react-resizable-panels` bundles into the console output
- [x] `yarn install --frozen-lockfile` clean (lockfile already in sync; no yarn.lock change)
- [x] `audit:class-refs` clean
- [ ] CI: format-check, lint, build, typecheck, audit-tokens green (react-gated jobs skip)

### DIFF
--- a/apps/console/package.json
+++ b/apps/console/package.json
@@ -23,6 +23,7 @@
     "react": "^19.2.3",
     "react-dom": "^19.2.3",
     "react-hook-form": "^7.77.0",
+    "react-resizable-panels": "^4.11.2",
     "recharts": "^3.8.1",
     "sonner": "^2.0.7",
     "tw-animate-css": "^1.4.0",
--- a/apps/console/src/hooks/use-media-query.ts
+++ b/apps/console/src/hooks/use-media-query.ts
@@ -1,0 +1,30 @@
+import { useCallback, useSyncExternalStore } from 'react';
+
+/**
+ * useMediaQuery
+ *
+ * Subscribes to a CSS media query and re-renders when it starts/stops matching.
+ * Reads the current match synchronously via `useSyncExternalStore`, so there is
+ * no first-render flash — important when layout branches on the result.
+ *
+ * @example
+ * ```tsx
+ * const isDesktop = useMediaQuery('(min-width: 64rem)');
+ * ```
+ */
+export function useMediaQuery(query: string): boolean {
+  const subscribe = useCallback(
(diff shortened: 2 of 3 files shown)


aishvaryagupta:
**Runtime-verified** (dev server + Playwright, since the react/axe CI jobs skip for console-only PRs):

- Desktop (1280px): split renders **28 / 72**, the list pane fills and scrolls internally, the thread pane fills with the composer, and the handle is a focusable `role=separator`.
- Mobile (390px): the resizable group is gone — the single-pane toggle (list **or** thread, driven by `c`) renders, no drag handle.

The check caught a real bug that build/typecheck could not: react-resizable-panels **v4 treats bare-number sizes as pixels**, so `maxSize={45}` had clamped the list pane to ~35px. Fixed by using percentage strings (`defaultSize="28%"` / `minSize="20%"` / `maxSize="45%"`).

Issue #332, opened 2026-06-02
refactor(react): reframe responsive code for mobile-first co-primary


aishvaryagupta:
## Context

The responsive philosophy was reframed to **mobile-first + desktop co-primary** (2026-06-02) across `responsive.md`, `breakpoints.json`, the docs Responsive page, and `components.md`. A few code sites still carry **desktop-first naming/comments**. Their **behavior is already correct** for co-primary (narrow → drawer, ≥`lg` → docked) — this issue is naming/comment hygiene only, deferred from the spec pass to keep the diff out of active Atlas work.

## Scope (behavior-preserving)

- **`packages/react/src/hooks/use-mobile.ts`** — rename `useIsMobile` → `useIsNarrow` (device-agnostic); drop the "below the Standard layout floor" desktop-centric comments. **Keep** the `(max-width: 63.99rem)` threshold and behavior. Update all import sites.
- **`packages/react/src/components/ui/sidebar.tsx`** — update the call site to the renamed hook; reframe comments (drawer below `lg` / docked at-or-above `lg` stays). The `nx:after:-inset-2` hit-area pattern is now cited by `responsive.md` § Touch targets — keep it.
- **`apps/console/src/app/auth-layout.tsx`** — reframe the mental model in comments (the form is the base; the brand sidebar is the ≥`lg` enhancement). `nx:lg:flex` class logic unchanged.

## Out of scope

No threshold/breakpoint-value changes; no behavior changes; no component retrofit for tap-targets.

_Deferred from the mobile-first co-primary spec pass (`responsive.md`). Milestone: slot into responsive-foundations or component Phase 2._

PR #374, opened 2026-06-03
feat(console): ⌘K command palette (Phase 4a)

aishvaryagupta:
## Summary

Phase 4a — the first **cross-cutting** slice (after the five Phase 3 breadth modules): a **⌘K command palette** that spans the whole console. Where each Phase 3 module owns its own search/filter, this dogfoods one global surface over all of them — a single `CommandDialog` that is a **launcher** when empty and a **cross-module search** once you type.

- **`shell/command-palette.tsx`** — a `CommandDialog` (cmdk via `@nexus/react`) with three concerns:
  - **Go to** (always) — every workspace module, rendered from the same `MODULE_ITEMS` registry the sidebar uses, with the identical `'route' in item` discrimination (built modules → static route; the rest → the `/m/$module` placeholder). The palette's nav and the sidebar's nav can't drift.
  - **Search** (only when the query is non-empty) — **contacts, issues, conversations, and people** in one list, each group fed by that module's **own TanStack Query key** (`crmKeys.contacts`, `projectKeys.issues`, …) so opening the palette after visiting a module is a cache hit, not a refetch. Queries are `enabled: open` so all four warm the moment the palette opens — no empty-state flash before the first keystroke.
  - **Actions** (always) — toggle theme (reuses the topbar's `useThemeContext`), jump to Appearance settings.
- **`root-layout.tsx`** — lifts the open state, adds a global **⌘K / Ctrl+K** keydown (a browser-API subscription, so an effect), and mounts the palette once in the shell.
- **`topbar.tsx`** — the ⌘K button was a disabled stub; it now opens the palette (interactive bg + hover + focus ring).

**Item-value scheme (non-obvious):** cmdk dedupes items by `value` **globally** and folds `value` into its fuzzy match. Record ids collide across modules — a contact and a conversation are **both `c-01`** — so a raw `value={id}` would silently break selection and let one record's id match another. Values are therefore **type-prefixed** (`contact-${id}`, `conversation-${id}`, …) for global uniqueness, with the human-searchable text in `keywords`. The inbox record is the one navigation odd-one-out: it routes via `search: { c: id }` (`/m/inbox?c=…`), not a path param.

## Scope & known (advisor-validated)

- **Fuzzy matching is permissive** — a short query (e.g. "Marcus") can surface loosely-scored items across groups. That's cmdk's **default scorer, shipped by the `@nexus/react` `Command` component's filter** — a design-system-layer decision, not console-tunable here. Longer queries tighten as expected.
- **Search-button tap target** keeps the topbar's `py-1.5` density rather than the ~44px touch floor — bumping only this control would make it stand taller than its neighbours (`SidebarTrigger`, theme toggle) and break the `h-14` bar's alignment. The whole-topbar mobile pass is tracked in **#332** (mobile-first reframe).
- **Deferred to later Phase 4 slices:** notifications center, the empty/error/404/403 state pass, and the responsive sweep are their own slices per the plan — not this PR.

## Test Plan (browser-verified — light + dark, console clean throughout)

- [x] **⌘K** (and **Ctrl+K**) toggles the palette open/closed from anywhere; the **topbar Search button** opens it too
- [x] **Empty state** → "Go to" (all 8 modules, mirroring the sidebar) + "Actions" (Toggle theme · Appearance); no record groups
- [x] **Typing** surfaces the record groups; "Go to"/"Actions" hide when nothing in them matches
- [x] **Contact** select → `/m/crm/c-01` (path param)
- [x] **Issue** select → `/m/projects/i-02` (path param)
- [x] **Member** select → `/m/people/m-02` (path param)
- [x] **Conversation** select → `/m/inbox?c=c-04` (search param, the odd-one-out) — and the thread actually loads
- [x] **Toggle theme** action flips `<html>.dark` and closes the palette
- [x] **Escape** closes; **reopen** after a selection → query is **reset to blank** (the palette stays mounted, so the controlled query is cleared on every close path)
- [x] Type-prefixed values confirmed unique in the DOM (`contact-c-01`, `conversation-c-04`, …) — no cross-module collision
- [x] Dark mode renders correctly (dark popover, legible text)
- [x] `typecheck` + `eslint --max-warnings 0` clean

### DIFF
--- a/apps/console/src/app/root-layout.tsx
+++ b/apps/console/src/app/root-layout.tsx
@@ -1,7 +1,10 @@
+import { useEffect, useState } from 'react';
+
 import { SidebarInset, SidebarProvider } from '@nexus/react';
 import { Outlet } from '@tanstack/react-router';
 
 import { AppSidebar } from '../shell/app-sidebar';
+import { CommandPalette } from '../shell/command-palette';
 import { Topbar } from '../shell/topbar';
 
 /**
@@ -11,11 +14,26 @@ import { Topbar } from '../shell/topbar';
  * bar persist across navigations.
  */
 export function RootLayout() {
+  const [commandOpen, setCommandOpen] = useState(false);
+
+  // ⌘K / Ctrl+K toggles the command palette from anywhere — a global keyboard
+  // subscription, which is exactly what an effect is for.
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'k' && (e.metaKey || e.ctrlKey)) {
+        e.preventDefault();
+        setCommandOpen((prev) => !prev);
+      }
+    };
+    document.addEventListener('keydown', onKeyDown);
+    return () => document.removeEventListener('keydown', onKeyDown);
+  }, []);
+
   return (
     <SidebarProvider>
       <AppSidebar />
       <SidebarInset>
-        <Topbar />
+        <Topbar onSearchClick={() => setCommandOpen(true)} />
         {/* nx:text-foreground re-establishes the adaptive base text color for
             module content — without it, reused content that relies on inherited
             foreground (e.g. the typography showcase) renders black in dark mode. */}
@@ -23,6 +41,7 @@ export function RootLayout() {
           <Outlet />
         </main>
       </SidebarInset>
+      <CommandPalette open={commandOpen} onOpenChange={setCommandOpen} />
     </SidebarProvider>
   );
 }
--- a/apps/console/src/shell/command-palette.tsx
+++ b/apps/console/src/shell/command-palette.tsx
@@ -1,0 +1,248 @@
+import { useState } from 'react';
+
+import {
+  CommandDialog,
+  CommandEmpty,
+  CommandGroup,
+  CommandInput,
+  CommandItem,
+  CommandList,
+} from '@nexus/react';
+import {
+  IconAddressBook,
+  IconBriefcase,
+  IconInbox,
+  IconMoon,
+  IconPalette,
+  IconSun,
+  IconUsers,
+} from '@tabler/icons-react';
+import { useQuery } from '@tanstack/react-query';
+import { useNavigate } from '@tanstack/react-router';
+
+import { useThemeContext } from '../app/theme-provider';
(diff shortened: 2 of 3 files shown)


examlly-tech (approved):
## 🤖 Principal Architect Review — #374

**Verdict: APPROVED**

### Architectural Concerns

| Severity | Area | Note |
| --- | --- | --- |
| ℹ️ Info | Client-side materialization | On open, all four record sets load in full and cmdk filters them in-memory (the four `useQuery` fire on `enabled: open`). This is an app-wide property — every module list route already loads-all-and-filters, and the palette reuses those exact query keys for a cache hit — **not a palette defect**. Server-side search is a data-layer decision outside this PR. |

Verified: cache-sharing is sound (palette reuses each module's `queryKey`+`queryFn`, so `enabled: open` warms/reuses the same entry — no double-fetch); dependency direction is clean (shell → lib, shell → sibling; modules never import shell); the four nav contracts (3 path-param + inbox `search:{c}`) all match the route definitions.

---

examlly-tech (approved):
## 🤖 SDE2 Review — #374 (full + follow-up)

**Verdict: APPROVED** — 2 findings, both fixed in-PR (`5967cbb`)

### Issues (resolved)

| # | Severity | File | Issue | Resolution |
| 1 | ⚠️ Minor | `command-palette.tsx` | ⌘K-close bypassed the query reset: `onOpenChange` only fires on Radix's own gestures, but ⌘K flips the controlled `open` prop in root-layout — so reopen showed the stale query. | Reset on the open→closed prop edge (adjust-state-during-render), covering Escape / outside / select / ⌘K alike; dropped the now-dead `handleOpenChange` wrapper. Browser-verified: type → ⌘K-close → reopen is blank. |
| 2 | ⚠️ Minor | `command-palette.tsx` | The 4 record queries had no error branch — a fetch failure read as "No results found.", out of step with the modules' `isError` handling. | `CommandEmpty` now shows a load-error message when any query errors. |

Follow-up re-review: both genuinely resolved; ripple clean (no dangling `handleOpenChange` refs); the during-render guard terminates with no flash. Type safety, TanStack Query/Router APIs, the effect deps + `preventDefault` chord, and inline-handler compliance all verified clean.

---

PR #377, opened 2026-06-03
feat(console): notifications center (Phase 4b)

aishvaryagupta:
## Summary

Phase 4b — the second **cross-cutting** slice: a **notifications center** in the topbar. A bell with an unread-count badge opens a `Popover` panel over the app-activity feed (mentions, assignments, comments, system, billing); tapping a notification — or "Mark all read" — marks it read.

- **`lib/notifications-api.ts`** — `Notification` type, `NOTIFICATION_KINDS` (`as const` → the union + the kind→icon map), `notificationKeys`, and `fetchNotifications` / `markNotificationRead` / `markAllNotificationsRead`.
- **`shell/notifications-menu.tsx`** — the self-contained bell + panel (Popover manages its own open state, mirroring the CRM saved-views menu — no root-layout lift):
  - Bell `Button` with an unread-count `Badge` overlay. The badge is `aria-hidden` because the count is already in the button's `aria-label` (`"Notifications, 3 unread"`), so a screen reader doesn't double-read it.
  - **Optimistic mark-read** (the CRM kanban's `onMutate` → `setQueryData` → rollback-on-error → reconcile-on-settle shape): the dot + bell count clear on tap, not after the ~800ms PATCH-then-refetch. Both mutations guard against double-fire while pending.
  - Per-kind leading icon from a single `Record<NotificationKind, …>` (no inline switch); unread rows get a bold title + a trailing dot; `isPending` shows a skeleton.
  - List is a plain `overflow-y-auto` div (not `ScrollArea`) because the bodies truncate — the documented ScrollArea-`display:table` gotcha.
- **`mocks`** — 7 deterministic seeds (3 unread, referencing real fixture records/teammates) + a mutable store + `GET /api/notifications`, `PATCH /:id/read`, `POST /read-all`.
- **`shell/topbar.tsx`** — mounts `<NotificationsMenu />` between the spacer and the theme toggle.

## Scope & rationale

- **Clicking marks read; it does not navigate.** The ⌘K command palette (Phase 4a) already owns cross-module jumps — re-treading that here would duplicate it and re-add the union-`to` typing for no new coverage. This slice earns its keep on the **read/unread lifecycle + the optimistic bell badge**.
- **Distinct from the Inbox module's `unread`.** This is the app-**activity** feed (mentions/assignments/system); Inbox's `unread` tracks customer-support conversation threads. They don't share a store.
- **No `EmptyState`.** The only mutations mark-read — nothing removes items — so the 7-seed feed never empties; an empty branch would be dead UI. A dismiss/clear action (and its empty state) is a future addition.
- **The red count `Badge` (`variant="error"`)** follows the universal notification-count convention — it signals unread volume, not an error.

## Test Plan (browser-verified — light + dark, console clean throughout)

- [x] Bell renders with the unread-count badge ("3"); `aria-label="Notifications, 3 unread"`, badge `aria-hidden`
- [x] Opening the panel shows the header ("Notifications" + "Mark all read"), a separator, and 7 rows — 3 unread (bold title + dot + kind icon) and 4 read
- [x] **Mark one** (tap an unread row) → that row's dot clears and the bell drops 3 → 2 **instantly** (optimistic), row stays in the list
- [x] **Mark all read** → every dot clears, the bell badge disappears, the button disables (`aria-label` back to "Notifications")
- [x] Escape closes the panel
- [x] Dark mode renders (dark popover surface, legible rows + red badge)
- [x] MSW endpoints exercised (GET 7 notifications, PATCH read, POST read-all); `typecheck` + `eslint --max-warnings 0` clean

### DIFF
--- a/apps/console/src/lib/notifications-api.ts
+++ b/apps/console/src/lib/notifications-api.ts
@@ -1,0 +1,79 @@
+/**
+ * Notifications API client. Thin typed wrappers over the mock endpoints served
+ * by MSW (`src/mocks/handlers.ts`) — there is no real backend. Drives the topbar
+ * bell: a flat activity feed with a read/unread lifecycle (mark one, mark all).
+ *
+ * This is the app-activity feed (mentions, assignments, system) — distinct from
+ * the Inbox module's per-conversation `unread`, which tracks customer support
+ * threads. The two don't share a store.
+ */
+
+/**
+ * Notification kinds — the single source for the {@link NotificationKind} union
+ * and the kind→icon map in the menu. Each ties loosely to a module's activity.
+ */
+export const NOTIFICATION_KINDS = [
+  'mention',
+  'assignment',
+  'comment',
+  'system',
+  'billing',
+] as const;
+
+/** What a notification is about — drives its leading icon. */
+export type NotificationKind = (typeof NOTIFICATION_KINDS)[number];
+
+export type Notification = {
+  id: string;
+  kind: NotificationKind;
+  /** One-line headline, e.g. "Grace Hopper mentioned you". */
+  title: string;
+  /** Supporting detail, e.g. "in ATL-102 — can you take a look?". */
+  body: string;
+  /** ISO timestamp (date + time) — rendered with {@link formatDateTime}. */
+  at: string;
+  /** Whether the recipient has seen it — drives the unread dot + bell count. */
+  read: boolean;
+};
+
+/**
+ * TanStack Query keys for notifications — declared once so the menu query and
+ * the mark-read mutations can't drift apart (a mismatched key silently breaks
+ * the optimistic cache writes).
+ */
+export const notificationKeys = {
+  all: ['notifications'] as const,
+  list: ['notifications', 'list'] as const,
+};
+
+export async function fetchNotifications(): Promise<{
+  notifications: Notification[];
+}> {
+  const res = await fetch('/api/notifications');
+  if (!res.ok) {
+    throw new Error('Failed to load notifications. Please try again.');
+  }
+  return (await res.json()) as { notifications: Notification[] };
+}
+
+/** Mark a single notification read — returns the updated notification. */
+export async function markNotificationRead(
+  id: string
+): Promise<{ notification: Notification }> {
+  const res = await fetch(`/api/notifications/${id}/read`, { method: 'PATCH' });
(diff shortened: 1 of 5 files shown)


examlly-tech (approved):
## 🤖 Principal Architect Review — #377

**Verdict: APPROVED**

### Architectural Concerns

| Severity | Area | Note |
| --- | --- | --- |
| ℹ️ Info | Unread count via full-list fetch | The topbar eagerly fetches the whole feed on mount to derive `unreadCount` client-side. Fine for 7 seeds and consistent with the 4a materialization tradeoff; against a real backend it doesn't survive pagination — split a count endpoint + a panel-open lazy list at that point. |

Verified: placement is correct (a shell feature — no route, no `modules/` dir — belongs in shell/ + lib/ + mocks/, matching `command-palette.tsx`); `notificationKeys` {all, list} matches the sibling convention (`all` is an invalidation prefix, partial-matching covers `list`); the 3 REST endpoints mirror inbox's PATCH-status + mutable-store lifecycle; the store is intentionally distinct from Inbox's per-conversation `unread`.

---

examlly-tech (approved):
## 🤖 SDE2 Review — #377 (full + follow-up)

**Verdict: APPROVED** — 3 findings, all fixed in-PR (`7365d67`, `19d0b0c`)

### Issues (resolved)

| # | Severity | Issue | Resolution |
| --- | --- | --- | --- |
| 1 | ⚠️ Minor | Fetch error rendered as empty-success (no `isError` branch), unlike every sibling | Added a three-way `isPending`/`isError`/`data` render (`7365d67`) |
| 2 | ⚠️ Minor | `markOne.isPending` serialized mark-read across rows → a tap on a 2nd row during the 1st's in-flight PATCH was silently dropped | Dropped the guard (mark-read is idempotent; `n.read` + the optimistic write prevent meaningful re-fires). Browser-verified: two rapid different-row taps both land (`7365d67`) |
| 3 | ⚠️ Minor | TanStack v5 retains `data` on a refetch error, so the new `isError` branch could stack above the list | Gated the error on `!data` — a refetch error keeps the stale list (`19d0b0c`) |

Verified clean: the optimistic mark-read idiom (cancel → snapshot → write → rollback → invalidate), mutation context typing, the `Record<NotificationKind>` map, double-fire handling, the `aria-hidden` count badge, and the `9+` cap.

---

Issue #412, opened 2026-06-04
a11y: configurable CardTitle / EmptyStateTitle heading semantics


aishvaryagupta:
Surfaced by the Atlas console **app-level axe sweep** (Phase 5d, #410-series). Two `@nexus/react` component gaps that every consuming app inherits:

## 1. `Sidebar` is not a navigation landmark — `region` (moderate, ×15 per page)

`Sidebar` renders `<div data-slot="sidebar">`, so all sidebar content (header, `SidebarGroupLabel`s, nav items) is flagged by axe's `region` rule as "not contained by landmarks." Every page in a sidebar app fails this.

**Fix:** make `Sidebar` (or `SidebarContent`) a navigation landmark — `<nav>` or `role="navigation"` with an `aria-label` (e.g. "Main"). shadcn's upstream has the same gap; Nexus should diverge here.

## 2. `CardTitle` heading level trips `heading-order` (moderate)

`CardTitle` renders a fixed `<h3>`. Used as a section heading under a page `<h1>` (no intervening `<h2>`), axe flags the level skip (seen on Analytics chart cards + Billing cards).

**Fix:** make `CardTitle`'s element configurable (an `as`/`level` prop, or polymorphic), so a consumer can render `<h2>` where the card is a top-level section. Default stays `<h3>`.

Both are non-blocking (moderate); the console sweep fixed all console-level a11y (duplicate `<main>`, empty table headers, missing page `<h1>`) and excludes axe `color-contrast` per the APCA policy.

aishvaryagupta:
Also surfaced by the Phase 5d sweep (#413 review): **`EmptyStateTitle` heading flexibility.** It renders a fixed `<div>` with no `asChild`, so a page whose primary heading is an EmptyState (e.g. the ComingSoon placeholder) can't make it the `<h1>` — the console works around it with a separate `sr-only <h1>`, duplicating the label in the a11y tree. Consider an `asChild` (Slot) on `EmptyStateTitle` (and `EmptyStateDescription`) so consumers can render the correct heading element. Same family as the CardTitle level issue above.

aishvaryagupta:
[...]

Issue #409, opened 2026-06-04
Audit primary.background as a chart mark (brand × base × theme APCA)


aishvaryagupta:
## Context

Phase 5b (#408) brands the single-series hero charts (Revenue area, Sessions line) with `var(--nx-color-primary-background)` so they re-colour when the brand flips (the topbar palette from 5a exposes all 6 brands).

## Concern (from #408 architect review)

`primary.background` is APCA-gated in `packages/core/scripts/audit-contrast.js` **only** as a button fill paired with `primary.foreground` text (`BRAND_PAIRS`) — never as a **bare chart mark on the chart canvas**. The categorical token it replaces *is* audited for exactly that role (`CHART_SURFACES` pairs `chart-categorical-*` against `background`/`container`).

So the brand mark must clear **APCA Lc 60 (UI tier)** as a 2px stroke / 0.3-opacity area fill on `background` **and** `container`, across all **6 brands × light/dark** — currently unverified.

## Action (Phase 5d — a11y/contrast sweep)

- Add `primary.background ↔ { background, container }` to the contrast audit (UI tier, Lc 60).
- Confirm all 6 brands × 2 themes pass; if any fail, adjust (e.g. fall back to a categorical token for that mark, or tune the shade).

Low actual risk (brand primaries are vivid accents designed to be legible on the canvas; the stroke carries the mark), but it should be formally gated. Owned by Phase 5d.

aishvaryagupta:
**Phase 5d verification.** The brand-chart-mark (`primary.background` as a single-series chart stroke/fill) is contrast-safe by reasoning: `primary.background` is the brand accent already APCA-gated for buttons (`primary.foreground ↔ primary.background`, UI tier Lc 60) and used for `border-primary` + link colors on the same `background`/`container` surfaces — so as a solid stroke on those surfaces it is equally legible across all 6 brands × 2 themes. Visually confirmed for blue + purple (5b screenshots).

The **formal audit gate** — adding `primary.background ↔ { background, container }` to `audit-contrast.js` — needs a brand×base combinatorial loop (today the audit iterates `BASE_PALETTES` and `BRAND_PALETTES` separately, and `primary.background` is per-brand while `background`/`container` are per-base). That's a focused `@nexus/core` change, disproportionate to bundle into the console a11y PR — so this stays open for it. Not a safety blocker; the marks are confirmed legible meanwhile.

aishvaryagupta:
[...]

PR #413, opened 2026-06-04
fix(console): app-level a11y sweep (Phase 5d)

aishvaryagupta:
## Summary

Phase 5d (finale). Ran an **axe-core sweep across every console route** — the first app-level a11y pass (components are axe-checked in Storybook, but the assembled app wasn't). Fixed every console-level finding:

| Finding | Fix |
| --- | --- |
| `landmark-no-duplicate-main` / `-main-is-top-level` / `landmark-unique` (every route) | `root-layout`'s content wrapper is now a `<div>` — `SidebarInset` already renders the `<main>` landmark, so there were two. |
| `empty-table-header` (CRM/Projects/People) | The Actions column gets a `sr-only` "Actions" header. |
| `page-has-heading-one` (unbuilt modules) | `ComingSoon` gets a `sr-only <h1>` (its visible title is an `EmptyStateTitle`, not a heading). |

Verified cleared by re-running axe after each fix.

## Not fixed — by policy

`color-contrast` (serious) flags e.g. the success badge (white-on-green, WCAG 3.35). **Nexus gates contrast via APCA, not WCAG ratios** — Storybook disables this same axe rule for exactly this reason (`testing-react.md`), and the token pairs pass `@nexus/core`'s `audit:contrast`. Excluded, not "fixed."

## Tracked separately

- **#412** — two `@nexus/react` component gaps surfaced by the sweep: `Sidebar` isn't a `nav` landmark (`region`, every route) and `CardTitle` is a fixed `<h3>` (`heading-order`). DS-level, not console-fixable.
- **#409** — the brand-as-chart-mark APCA gate (from 5b) needs a brand×base combinatorial addition to `audit-contrast.js`; verified safe by reasoning + visual, tracked for a focused core change.

## Test Plan

- [x] `typecheck` + `eslint` — clean
- [x] axe-core swept across crm/projects/analytics/billing/inbox/people/settings before & after; landmark / empty-table-header / page-has-heading-one all cleared; only the two DS-component findings (#412) remain (color-contrast excluded per APCA policy)

### DIFF
--- a/apps/console/src/app/root-layout.tsx
+++ b/apps/console/src/app/root-layout.tsx
@@ -34,12 +34,14 @@ export function RootLayout() {
       <AppSidebar />
       <SidebarInset>
         <Topbar onSearchClick={() => setCommandOpen(true)} />
-        {/* nx:text-foreground re-establishes the adaptive base text color for
+        {/* SidebarInset already provides the <main> landmark, so this is a
+            plain <div> (a second <main> trips axe landmark rules).
+            nx:text-foreground re-establishes the adaptive base text color for
             module content — without it, reused content that relies on inherited
             foreground (e.g. the typography showcase) renders black in dark mode. */}
-        <main className="nx:text-foreground nx:min-w-0 nx:flex-1">
+        <div className="nx:text-foreground nx:min-w-0 nx:flex-1">
           <Outlet />
-        </main>
+        </div>
       </SidebarInset>
       <CommandPalette open={commandOpen} onOpenChange={setCommandOpen} />
     </SidebarProvider>
--- a/apps/console/src/modules/coming-soon.tsx
+++ b/apps/console/src/modules/coming-soon.tsx
@@ -24,6 +24,8 @@ export function ComingSoon() {
 
   return (
     <EmptyState className="nx:min-h-[60svh]">
+      {/* The page's h1 — the visible title is EmptyStateTitle, not a heading. */}
+      <h1 className="nx:sr-only">{label}</h1>
       <EmptyStateHeader>
         {Icon && (
           <EmptyStateMedia variant="icon">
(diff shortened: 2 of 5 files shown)


examlly-tech (approved):
## 🤖 Principal Architect Review — Verdict: APPROVED

App-level a11y sweep, verified against source: `SidebarInset` is the `<main>` landmark (the inner one was a redundant second), the new `<div>` keeps `flex-1 min-w-0 text-foreground` verbatim (layout unchanged), `page-header` emits a real `<h1>` so `ComingSoon` was the only console page-heading gap, all three `*-columns.tsx` are the complete table set (8/8 headers), and the `color-contrast` exclusion matches Storybook's APCA policy. #412 (DS gaps) + #409 (chart-mark gate) are correctly scoped — both non-console-fixable.

### Discussion notes (non-blocking)

- **sr-only h1 duplicates the title in the a11y tree** (`coming-soon.tsx`) — a SR hears `label` twice (the h1 + the visible `EmptyStateTitle`). `EmptyStateTitle` is a plain `<div>` with no `asChild`, so `<EmptyStateTitle asChild><h1>` isn't available, and defaulting it to a heading would break `heading-order` in its empty-state uses. The `sr-only` h1 is the correct console-level call; `EmptyStateTitle asChild` is folded into #412.
- **Topbar sits inside the single `<main>`** (`root-layout.tsx`) — pre-existing (`SidebarInset` always wrapped both Topbar + content), axe passes, not introduced here.

---
_Review against `.claude/rules/` guidelines._

examlly-tech (approved):
## 🤖 SDE2 Review — Verdict: APPROVED

Verified by a fresh console build that `.nx\:sr-only` IS emitted (`clip-path:inset(50%); width:1px; height:1px; position:absolute; overflow:hidden`) — the h1 / Actions headers are hidden-but-accessible, not visible (the load-bearing risk). `typecheck` clean (`header: () => ReactNode` is valid `ColumnDef` typing). The inner-`<main>`→`<div>` swap preserves the landmark (`SidebarInset`'s `<main>`), the `flex-1 min-w-0 text-foreground` classes, and the slot — no console code depended on the old element. `EmptyStateTitle` renders a `<div>`, so the sr-only h1 is the only heading (no double-heading). All three column files carry the identical fix.

---
_Review against `.claude/rules/` guidelines._

PR #417, opened 2026-06-04
feat(console): persisted, configurable sidebar collapse (icon-rail / offcanvas)

aishvaryagupta:
## Summary

The Atlas console sidebar's collapse behaviour is now **persisted and user-configurable**.

- **Persist** — collapse/expand state reset on every reload: the Nexus `SidebarProvider` writes a `sidebar_state` cookie only for _server-side_ restore and never reads it back, and the console is a client-only SPA. A Zustand `persist` store (`apps/console/src/app/sidebar-store.ts`, localStorage key `nexus-console-sidebar`) now drives the provider's controlled `open`/`onOpenChange`; the rail toggle and `Cmd/Ctrl+B` both route through it (synchronous hydration → no first-paint flash).
- **Configurable mode** — a "Sidebar" toggle in the Appearance panel (`/design/appearance`) chooses how the sidebar collapses, persisted as `mode` in the same store:
  - **Icon rail** (`collapsible="icon"`) — collapses to icons. Section labels are dropped (their `-mt-8`/`opacity-0` collapse fade reflows the rail) and a `SidebarSeparator` keeps the two groups distinct.
  - **Full collapse** (`collapsible="offcanvas"`) — hides entirely; keeps the "Design System" / "Workspace" section headers.
- **Review fixes** — store relocated `shell/` → `app/` (alongside `session.ts`); JSDoc rationale trimmed per `code-comments.md`.

## GitHub Issue

No tracked issue — fixes the seeded demo-backlog item in `apps/console/src/mocks/projects-fixtures.ts` (_"Sidebar collapse state not persisted"_) and extends it into a user preference.

## Test Plan

Verified at runtime (Vite dev + Playwright):

- [x] Persistence: collapse/expand → reload → state survives (both `open` and `mode`); the toggle reflects the saved mode
- [x] Mode × open 2×2 — icon+expanded (separator, no headers) · icon+collapsed (rail + separator) · offcanvas+expanded (headers, no separator) · offcanvas+collapsed (fully hidden, width 0)
- [x] Switching mode **while collapsed** (icon ↔ offcanvas) transitions cleanly — no stale-prop glitch
- [x] `yarn workspace @nexus/console typecheck` + `eslint` clean

## Notes / open decisions

- **Desktop-only:** mobile renders the Sheet drawer regardless of `mode`, so the toggle has no effect on narrow viewports.
[...]

PR #421, opened 2026-06-04
fix(react): give Sidebar a navigation landmark

### LINKED ISSUE #418: a11y(sidebar): Sidebar exposes no navigation landmark
## Summary

The `Sidebar` component (`packages/react/src/components/ui/sidebar.tsx`) renders **no navigation landmark**. Its root is a plain `<div data-slot="sidebar">` (see `SidebarProvider`/`Sidebar`), and the only `aria-label` in the component is on the toggle button. So a consumer's sidebar nav is not exposed as a `navigation` landmark to assistive tech or to landmark-navigation (rotor / NVDA `D` / etc.).

## Impact

- Screen-reader users can't jump to the sidebar via landmark navigation.
- Grouping inside the sidebar has no landmark to anchor to. (Atlas console PR #417 worked around this at the app level by putting `role="group"` + `aria-label` on each `SidebarGroup` in icon mode — a per-app patch that ideally the component should make unnecessary.)

## Proposal

Expose the sidebar's nav region as a landmark — render the inner nav container as `<nav>` (or `role="navigation"`), and allow consumers to pass an `aria-label` (e.g. "Main") so multiple navs stay distinguishable. Decide whether the landmark sits on `Sidebar` itself or on `SidebarContent`, and confirm it doesn't double up when a consumer nests their own `<nav>`.

## Context

Discovered during the Atlas console sidebar work (#417). Found while reviewing an a11y follow-up: dropping `SidebarGroupLabel` in the icon rail removed the groups' accessible names, and there was no landmark to fall back on. The console fix is shipped; this issue tracks the component-level root cause for all consumers.

aishvaryagupta:
## Summary

- `SidebarContent` now renders as `<nav aria-label="Sidebar">` instead of a plain `<div>`, giving the Nexus `Sidebar` a proper navigation landmark. Assistive tech can jump straight to the sidebar's navigation via the landmarks rotor.
- The label defaults to `"Sidebar"` and is overridable through `aria-label` for pages that carry more than one navigation landmark (keeps axe's `landmark-unique` satisfied).
- `WithDataAttributes` story gains a runtime assertion that the content region resolves as a labelled `navigation` landmark.

## Why `SidebarContent`, not the whole `Sidebar`?

The landmark sits on the scrollable content region rather than the `Sidebar` root for two reasons:

- **Branch-agnostic.** `Sidebar` has three render branches — a `collapsible="none"` div, a mobile `Sheet` (a Radix Dialog), and the desktop docked panel. The mobile branch is a `role="dialog"`; making the root a `<nav>` would fight that role or force per-branch landmark logic. `SidebarContent` renders identically across all three, so one `<nav>` covers every branch.
- **Tight nav scope.** `SidebarHeader` / `SidebarFooter` are chrome (brand mark, account menu, collapse trigger), not navigation. Scoping the landmark to the content region keeps the nav to the actual menu groups, so the landmark announces only what it contains.

## GitHub Issue

Closes #418

## Test Plan

- [x] `eslint` + `prettier --check` on changed files — clean
- [x] Runtime-verified in Storybook (from this branch's `src`): `WithDataAttributes` renders exactly one `<nav data-slot="sidebar-content" aria-label="Sidebar">` — implicit `navigation` role, `landmark-unique` satisfied (one nav, uniquely named).
- [x] Probed the icon-collapsed rail (`Collapsed` story): the landmark **survives collapse** — still present, still named, in the a11y tree (`display: flex`), `overflow: hidden` applied.
- [ ] `yarn test:storybook` runs in CI (the new `getByRole('navigation', { name: 'Sidebar' })` assertion lives in `WithDataAttributes`).

### DIFF
--- a/packages/react/src/components/ui/sidebar.tsx
+++ b/packages/react/src/components/ui/sidebar.tsx
@@ -577,12 +577,16 @@ interface SidebarContentProps extends React.ComponentProps<'nav'> {}
 function SidebarContent({
   className,
   'aria-label': ariaLabel = 'Sidebar',
+  'aria-labelledby': ariaLabelledby,
   ...props
 }: SidebarContentProps) {
   return (
     <nav
       data-slot="sidebar-content"
-      aria-label={ariaLabel}
+      // A labelledby reference, when given, names the landmark — so suppress the
+      // default string label rather than emitting both on the <nav> (issue #418).
+      aria-label={ariaLabelledby ? undefined : ariaLabel}
+      aria-labelledby={ariaLabelledby}
       className={cn(
         // nexus-allow-numeric: sidebar nav-chrome rhythm
         'nx:flex nx:min-h-0 nx:flex-1 nx:flex-col nx:gap-2 nx:overflow-auto nx:group-data-[collapsible=icon]:overflow-hidden',
--- a/packages/react/src/components/ui/sidebar.tsx
+++ b/packages/react/src/components/ui/sidebar.tsx
@@ -564,18 +564,25 @@ function SidebarSeparator({ className, ...props }: SidebarSeparatorProps) {
  *
  * Props for the SidebarContent component.
  */
-interface SidebarContentProps extends React.ComponentProps<'div'> {}
+interface SidebarContentProps extends React.ComponentProps<'nav'> {}
 
 /**
  * SidebarContent
  *
- * Scrollable middle region holding the sidebar's groups and menus. Hides
- * overflow when collapsed to the icon rail.
+ * The sidebar's scrollable nav region holding its groups and menus. Renders as a
+ * `<nav>` landmark, named via `aria-label` (default "Sidebar"), so assistive tech
+ * can jump to it — pass a distinct `aria-label` when a page has more than one
+ * navigation. Hides overflow when collapsed to the icon rail.
  */
-function SidebarContent({ className, ...props }: SidebarContentProps) {
+function SidebarContent({
+  className,
+  'aria-label': ariaLabel = 'Sidebar',
+  ...props
+}: SidebarContentProps) {
   return (
-    <div
+    <nav
       data-slot="sidebar-content"
+      aria-label={ariaLabel}
       className={cn(
         // nexus-allow-numeric: sidebar nav-chrome rhythm
         'nx:flex nx:min-h-0 nx:flex-1 nx:flex-col nx:gap-2 nx:overflow-auto nx:group-data-[collapsible=icon]:overflow-hidden',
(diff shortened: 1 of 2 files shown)


examlly-tech (approved):
## 🤖 Principal Architect Review

**Verdict: APPROVED**

`SidebarContent` is the correct landmark home — it is the single node every render branch (`collapsible="none"` div, mobile Sheet, desktop docked panel) funnels children through, and it scopes the landmark to the menus while excluding the header/footer chrome. The `ComponentProps<'div'>` → `<'nav'>` change is zero-blast (identical `HTMLAttributes`; only the ref type narrows, and no consumer passes a ref to `SidebarContent`). The live console shell has exactly one navigation region after this change — no `landmark-unique` collision. The #417 per-group `role="group"` workaround is orthogonal (group ≠ navigation) and correctly stays.

### Architectural Concerns

| Severity | File | Concern |
| --- | --- | --- |
| ⚠️ Non-blocking | `sidebar.tsx` (mobile Sheet branch) | The Sheet names the region twice — sr-only `SheetTitle "Sidebar"` plus the new `<nav aria-label="Sidebar">`. Valid ARIA, distinct roles (dialog vs navigation); acceptable as-is. |

---

examlly-tech (commented):
## 🤖 SDE2 Review

**Verdict: MINOR CHANGES**

### Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `sidebar.tsx:579` | Default `aria-label="Sidebar"` is always applied, so a consumer passing `aria-labelledby` ends up with a stray redundant `aria-label` alongside it. The name still resolves via `labelledby` (accname precedence), but the extra attribute is dead markup. Mirror `slider.tsx:35-36`: destructure both and suppress the default when `labelledby` is set. |

---

packages/react/src/components/ui/sidebar.tsx
```diff
@@ -564,18 +564,25 @@ function SidebarSeparator({ className, ...props }: SidebarSeparatorProps) {
...
- * overflow when collapsed to the icon rail.
+ * The sidebar's scrollable nav region holding its groups and menus. Renders as a
+ * `<nav>` landmark, named via `aria-label` (default "Sidebar"), so assistive tech
+ * can jump to it — pass a distinct `aria-label` when a page has more than one
+ * navigation. Hides overflow when collapsed to the icon rail.
  */
-function SidebarContent({ className, ...props }: SidebarContentProps) {
+function SidebarContent({
+  className,
+  'aria-label': ariaLabel = 'Sidebar',
```
examlly-tech, line 579: **Minor — redundant label attr:** the default `aria-label="Sidebar"` always lands on the `<nav>`, so a consumer passing `aria-labelledby` (via `...props`) gets both. `labelledby` wins per accname, so the name is correct — but the stray `aria-label` is dead markup. Fix: also pull out `'aria-labelledby': ariaLabelledby`, set `aria-label={ariaLabelledby ? undefined : ariaLabel}`, and re-apply `aria-labelledby`. Precedent: `slider.tsx:35-36`.

Issue #433, opened 2026-06-08
Decide control sizing: fixed-height vs padding and labeled min-width


aishvaryagupta:
## Context

The role-spacing epic (#124, closed by #237) established role tokens — `px-control-*`, `gap-control-*`, `p-container`, etc. — as the spacing model for interactive controls, giving them per-`data-style` density response. Shipped controls (Tabs, Input, NativeSelect, Textarea, Menubar, Tooltip, Select; Button until now) consume `control-*`.

Two components have since moved to **numeric** spacing ahead of any system-wide decision:
- **Button** — PR #432 (`[codex] Audit Button against Figma anatomy`) moved to Figma-derived numeric sizing — ultimately **fixed heights** (`h-8`/`h-10`/`h-12`) + numeric padding (`px-3`, `px-2.5`).
- **AlertDialog** — earlier numeric migration (deliberate outlier).

PR #432's review flagged this as an untracked exit from a live system.

## Decision needed

Pick one direction for **all** control-tier components:
1. **Migrate the whole control tier to numeric** — drop `control-*` density tokens; accept that controls no longer rescale with `data-style`.
2. **Keep role tokens** — revert Button (#432) and AlertDialog to `control-*`, so the tier stays consistent and density-responsive.

Per-component drift (some numeric, some role) is the state to avoid.

This decision also covers the **sizing mechanism**, not just the token namespace, and the tier-wide call must settle three things: (a) `control-*` role tokens vs numeric tokens; (b) **padding-based vs fixed-height** sizing — Button #432 sizes by fixed height (`h-N`), which `components.md` §Sizing and `responsive.md` §Touch targets name as an anti-pattern (controls should size by padding so they don't clip in flex layouts and can meet the ~44px touch floor); and (c) **whether labeled buttons carry a min-width floor** — Button #432 dropped its per-size `min-w` (see Notes). PR #432 was approved accepting fixed height for Button on the explicit basis that this issue owns the tier-wide reconciliation.

## Notes
- Numeric utilities are themselves per-`data-style` (only a few keys like `spacing-2` are mode-invariant), so "numeric" ≠ "frozen" — see the per-mode `[data-style]` blocks in `packages/tailwind/nexus.css`.
[...]

PR #446, opened 2026-06-11
Audit drawer component

aishvaryagupta:
## Summary
- Migrate Drawer title/description and header/footer spacing to the approved Nexus utilities.
- Add a stable drawer-handle slot for the bottom drag affordance.
- Expand Drawer Storybook coverage for scrollable content, header close actions, direction behavior, and bottom-only drag-handle visibility.

## Validation
- pnpm --filter @nexus/react audit:storybook-coverage --component drawer
- pnpm lint
- pnpm typecheck
- pnpm format:check
- pnpm test:storybook exits 0 but reports "No test files found" — the Storybook vitest project collects 0 tests on this base (known infra, tracked in #429), so the new play functions + axe a11y checks run only once #429 lands; verified meanwhile via the browser sweep below.
- Storybook browser sweep on http://127.0.0.1:6007/ for the new Drawer stories, with no fresh console errors

## Notes
- Keeps Vaul as the Drawer primitive.
- No API changes.
- Header/footer padding moves `p-container` -> `p-6` (role -> numeric) as a deliberate decision; tracked under #433 (control-tier spacing), consistent with Button/AlertDialog. Preserves the default (vega) value; differs only in non-default density modes.
- ScrollableContent's scroll region is keyboard-focusable (`tabIndex`; axe `scrollable-region-focusable`) and its story hook uses `data-testid` rather than a component `data-slot`.
- Scope is limited to packages/react/src/components/ui/drawer/drawer.tsx and packages/react/src/components/ui/drawer/Drawer.stories.tsx.

### DIFF
--- a/packages/react/src/components/ui/drawer/Drawer.stories.tsx
+++ b/packages/react/src/components/ui/drawer/Drawer.stories.tsx
@@ -117,7 +117,7 @@ export const ScrollableContent: Story = {
         </DrawerHeader>
         <div
           data-testid="drawer-scroll-area"
-          // Keyboard-reachable scroll region (axe scrollable-region-focusable).
+          // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex -- scrollable region must be keyboard-reachable (axe scrollable-region-focusable)
           tabIndex={0}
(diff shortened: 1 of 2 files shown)


INNOVATIVEGAMER (approved):
approved by @INNOVATIVEGAMER

PR #453, opened 2026-06-14
feat(input-otp): audit component

aishvaryagupta:
## Summary

- Simplify Input OTP spacing from role-based gap-control-md to numeric gap-2, preserving the default 8px output while removing non-default spacing-mode variance.
- Move the slot character to the typography-body-small composite, make the separator decorative, add reduced-motion handling to the fake caret, and set autocomplete=one-time-code explicitly while preserving consumer overrides.
- Add Input OTP story assertions for the semantic input/separator contract and opt InputOtp into base-variant generation.

## GitHub Issue

Refs #433

## Test Plan

- [x] pnpm --filter @nexus/react audit:storybook-coverage --component input-otp
- [x] pnpm format:check
- [x] pnpm lint
- [x] pnpm typecheck
- [x] pnpm test:storybook exits 0 after regenerating base variants; current Storybook/Vitest setup reports No test files found for the storybook project.

## Spacing Note

The previous gap-control-md value varied by spacing mode: nova 6px, maia 10px, sera 12px, with root/default, mira, vega, luma, and lyra at 8px. gap-2 fixes the gap at 8px. The canonical single InputOTPGroup layout has one flex child, so the root gap is a no-op there; split group layouts keep a fixed slot rhythm because this component has no Figma source to derive a per-mode rhythm from.

### DIFF
--- a/packages/react/scripts/base-variants.config.json
+++ b/packages/react/scripts/base-variants.config.json
@@ -82,6 +82,7 @@
       }
     },
     { "name": "InputGroup", "showcase": "AllVariants" },
+    { "name": "InputOtp", "showcase": "AllVariants" },
     { "name": "Item", "showcase": "AllVariants" },
     { "name": "Kbd", "showcase": "AllVariants" },
     {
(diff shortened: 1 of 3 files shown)


INNOVATIVEGAMER (approved):
approved by @INNOVATIVEGAMER

PR #454, opened 2026-06-14
[codex] Audit Input fixed-height scale

aishvaryagupta:
## Summary
- Align Input sizing with the approved fixed-height scale: `h-8`, `h-10`, and `h-12`.
- Preserve native input semantics, full-width layout, placeholder/focus/invalid/file-input behavior, and Input-specific body typography.
- Update Input Storybook coverage for scale-aware mode heights, implicit `data-size="default"`, and class assertions for `h-*`, `px-*`, and `py-0` utilities.

## Validation
- `pnpm --filter @nexus/react audit:storybook-coverage --component input`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`
- `git diff --check prasad/components-cleanup...HEAD`
- Browser-measured Storybook on `localhost:6008`: Vega Input heights are `sm=32px`, `default=40px`, `lg=48px`; Nova/Maia also match the story sentinel values.

## Notes
- Button PR #432 is the current source for the fixed-height control scale and is still open; this Input branch follows its latest `h-12` large-size direction.
- Existing `components.md` guidance still says controls should avoid fixed `h-*`; the control-tier sizing rule needs the corresponding sanctioned update.
- Storybook gate caveat: if PR #429 is not landed, story play functions may not run in CI, so the local `:6008` browser measurement is the concrete visual proof for this pass.

## Scope note

Beyond Input, this PR also touches `select.tsx` and `textarea.tsx`: the `SelectTrigger` and `Textarea` gained `nx:enabled:hover:bg-background-hover` so the control tier shares one hover treatment after the Input hover-token correction. This resolves the review's hover-token, tautological-hover-test, and Select/Textarea tier-consistency findings. The `components.md` fixed-height rule update and #432 / #433 coordination remain intentionally out of scope here.

Input's richer **disabled** treatment (the disabled-token set, per Figma `843:71944`) is an intentional exception; the rest of the control tier stays on `disabled:opacity-50` for now, with tier convergence tracked in #456.

### DIFF
--- a/packages/react/src/components/ui/input/Input.stories.tsx
+++ b/packages/react/src/components/ui/input/Input.stories.tsx
@@ -6,7 +6,7 @@ import {
   AllModesRow,
   SPACING_MODES,
 } from '../../../stories/spacing-modes';
-import { expectHeightPinned } from '../../../stories/test-utils';
+import { expectHeightPerMode } from '../../../stories/test-utils';
 
 import { Input } from './input';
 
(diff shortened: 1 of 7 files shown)


aishvaryagupta:
Superseded by #462, which combines this Input fixed-height audit with the InputGroup audit (#455) into a single PR — InputGroup hard-depends on the new Input, so they must land together. Every commit from this branch is included in #462, rebased onto the current `prasad/components-cleanup` (now including #429).

PR #509, opened 2026-06-23
feat(table): adopt cross-DS research findings — a11y, zebra, mobile + recipes

aishvaryagupta:
## Summary

Stacks on #493. Adopts all 15 findings from a cross-design-system Table study (shadcn / Radix / Catalyst / Tremor · MUI / AntD / AG Grid · Carbon / Atlassian / Primer / React-Aria · Chakra / Mantine / Polaris / Fluent), each routed by what belongs in a primitives-first, bundle-constrained DS.

**In-component (4 — cheap, correct):**
- `scope="col"` default on `TableHead` (overridable; blank corner via `scope={undefined}`).
- New `<TableRowHeader>` → a real `<th scope="row">`, so each row gets an accessible name.
- `aria-sort` indicator — an actively-sorted header emphasizes muted→foreground (keyed on the valid `ascending`/`descending`, never `asc`/`desc`).
- `striped` prop → `data-striped` zebra; precedence is **structural** (`:not(:hover):not([data-state=selected])`), so selection and hover always win.

**Recipes (11 — zero published-bundle, tested Storybook stories):** sortable-header button · multi-select + indeterminate + bulk-action bar · whole-row link · sticky column (`bg-inherit` tracks row state) · container-query column-priority hide · numeric / tabular · width-bounded truncate · `--gutter` bleed · a11y-safe stacked-card (structure swap, not `display:block`) · live-region announcer · DataTable (TanStack — lives in `apps/console`).

## Council fixes

A 4-lens plan council caught 5 blockers the naive plan would have shipped green:
- stacked-card via `display:block` strips `<table>` semantics for screen readers → **structure swap** instead;
- `<Hide containerBelow>` can't hide a table column (wraps children) → **container-query class on cell+header**;
- `<td scope="row">` is inert (scope is `<th>`-only) → **`<TableRowHeader>`**;
- sticky-column opaque bg paints over the row tint → **`bg-inherit`**;
- zebra `nth-child` out-specifies `data-state=selected` → **structural `:not()` exclusion**.

## Guardrails
- `role="table"` stays the default — never a silent `role="grid"` (that 2-D keyboard contract is the consumer's logic layer).
- Bundle: **JS 86.53 / 88 kB**, **CSS 14.7 / 30 kB**. The 4 in-component changes cost ~0.14 kB JS; the recipes cost 0 (they live in stories).

## Test Plan
- [x] `pnpm --filter @nexus/react typecheck` · `pnpm lint` · `pnpm format:check` — clean
- [x] `pnpm size-limit` — JS 86.53/88 kB, CSS 14.7/30 kB
- [x] `pnpm test:storybook` — Table **26 stories** + base-variants grid **1/1** (27 total), with computed-style proofs for zebra precedence, sticky `bg-inherit` tracking, truncate clipping, column-priority hide, the stacked-card swap, and the `aria-sort` value
- [x] `pnpm --filter @nexus/react audit:storybook-coverage --component table` — 0 missing / 0 drift

### DIFF
--- a/packages/react/src/components/ui/table/Table.stories.tsx
+++ b/packages/react/src/components/ui/table/Table.stories.tsx
@@ -422,6 +422,50 @@ export const SemanticScope: Story = {
   },
 };
 
+// A sorted column reads as active: TableHead emphasizes its text (muted →
+// foreground) when it carries aria-sort="ascending"/"descending", but not for a
+// merely-sortable "none". The interactive sort control is a recipe (SortableHeader).
+export const AriaSortIndicator: Story = {
+  render: () => (
+    <Table>
+      <TableHeader>
+        <TableRow>
+          <TableHead aria-sort="ascending">Invoice</TableHead>
+          <TableHead aria-sort="none">Status</TableHead>
+          <TableHead aria-sort="descending" className="nx:text-right">
+            Amount
+          </TableHead>
+        </TableRow>
+      </TableHeader>
+      <TableBody>
+        <TableRow>
+          <TableRowHeader>INV001</TableRowHeader>
+          <TableCell>Paid</TableCell>
+          <TableCell className="nx:text-right">$250.00</TableCell>
+        </TableRow>
+      </TableBody>
+    </Table>
+  ),
+  play: async ({ canvasElement }) => {
+    const asc = canvasElement.querySelector('[aria-sort="ascending"]');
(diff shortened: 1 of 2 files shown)


aishvaryagupta:
Auto-merged when component-table-audit fast-forwarded to include these commits — folded into #493 so the full table audit (17 commits) reviews as a single PR. No separate merge needed; the work lands via #493 into components-cleanup.

Issue #546, opened 2026-06-27
Rename Appearance token mode IDs to friendly names everywhere


aishvaryagupta:
## Summary

We are no longer doing the smaller middle-path change where friendly public values map back to internal names like `nova`, `mira`, `maia`, and `vega`.

The goal is the full internal migration: get rid of the old token-mode names everywhere we actively ship. Public API values, runtime DOM attributes, token file names, generated CSS selectors, docs, stories, tests, and examples should all use friendly names.

In plain terms:

```txt
compact public value -> internal compact -> data-style="compact" -> spacing-compact.json -> [data-style="compact"]
```

Not:

```txt
compact public value -> internal nova -> data-style="nova"
```

## Target Direction

Use friendly names as the canonical token mode IDs.

| Layer | Before | After |
| --- | --- | --- |
| User label | Compact | Compact |
| Public API | `nova` or mapped `compact` | `compact` |
| Token mode | `nova` | `compact` |
| DOM attr | `data-style="nova"` | `data-style="compact"` |
| Token file | `spacing-nova.json` | `spacing-compact.json` |
| Generated CSS | `[data-style="nova"]` | `[data-style="compact"]` |

## Initial Mapping

PR 1 should verify this against the actual token inventory before rename work begins.

### Public Appearance Modes

| Axis | Old ID | New ID |
| --- | --- | --- |
| Density | `nova` | `compact` |
| Density | `mira` | `default` |
| Density | `luma` | `comfortable` |
| Density | `sera` | `spacious` |
| Corners | `sharp` | `square` |
| Corners | `subtle` | `subtle` |
| Corners | `smooth` | `smooth` |
| Corners | `mellow` | `round` |
| Elevation | `maia` | `quiet` |
| Elevation | `mira` | `standard` |
| Elevation | `nova` | `strong` |
| Stroke | `maia` | `fine` |
| Stroke | `vega` | `normal` |
| Stroke | `nova` | `strong` |

### Hidden or Non-Appearance Modes To Inventory

These old IDs exist outside the public curated Appearance subset and need explicit friendly names before the cutover.

| Family | Old ID | Proposed new ID |
| --- | --- | --- |
| Density | `lyra` | `tight` |
| Density | `vega` | `regular` |
| Density | `maia` | `relaxed` |
| Corners | `blunt` | `extra-round` |
| Elevation | `vega` | `flat` |
| Elevation | `lyra` | `soft` |
| Stroke | `lyra` | `medium` |
[...]

Issue #531, opened 2026-06-25
Track remaining Appearance consolidation: engine package, SSR, dogfood cleanup


aishvaryagupta:
## Summary

This issue tracked the Appearance theming epic end to end.

The original problem was that PR #437 introduced the Codex runtime theming engine while the older `/design/appearance` screen still existed, leaving the app with two theming destinations:

- `/design/appearance` - old preset/token-axis customization
- `/design/codex` - new derived theme editor

The completed direction is broader and clearer: Nexus now exposes one production-ready Appearance experience, powered by package-level APIs that console and external consumers can use out of the box.

## Product Direction

- Keep `/design/appearance` as the only theming destination.
- Delete `/design/codex` from the router and sidebar.
- Treat Codex as the internal derived-theme engine, not a user-facing page.
- Use one custom brand color selector.
- Use curated neutral base tones: Stone, Neutral, Zinc, Slate, Gray.
- Use the package Appearance typography controls.
- Show spacing/radius/shadow/border width with friendly labels instead of raw token names.
- Make `contrast` a structure control: surfaces, borders, dividers, nav states, controls, cards, and composed panels.
- Do not make `contrast` a typography hierarchy knob.
- Package the model so console and external consumers use the same behavior.

### Phase A - Full-token derivation + contrast as structure control

Status: Done, merged in #535.

- `@nexus/core` derives the full runtime color token set.
- Primary, secondary, status, surfaces, chart, alpha, focus, and text safety gates are package-owned.
- Contrast controls structure, with APCA text safety preserved.

### Phase B - Package engine and provider

Status: Done, merged in #536.

- Added the package Appearance model.
- Added `@nexus/react/appearance` provider and components.
- Added package exports, type coverage, and story coverage hooks.

### Phase C - First-paint bootstrap

Status: Done, merged in #537.

- Added the first-paint bootstrap path.
- Ensured persisted Appearance state can paint before React hydrates.
- Kept the bootstrap tied to the core-derived contract instead of duplicating formulas.

### Phase D - Console dogfood

Status: Done, merged in #545.

- Console consumes `@nexus/react/appearance` and `@nexus/tailwind`.
- `/design/appearance` is the single theming page.
- `/design/codex` is removed.
- Console-local Codex runtime/UI/CSS duplication is removed.
- Settings -> Appearance, topbar quick control, and command palette use the same package model.
- Runtime Tailwind color utilities bridge to `--nx-color-*` variables with static fallbacks.
[...]

PR #549, opened 2026-06-27
feat(core): inventory friendly token-mode migration

aishvaryagupta:
## Summary

- Add the Phase E0 friendly token-mode migration plan for #546.
- Freeze the family-scoped old-to-friendly mode map so `vega`/`nova`/`maia` collisions are resolved per token family, not by global replacement.
- Add a pre-rename value oracle for all 28 mode value-sets so the later cutover can prove every token value is preserved.
- Add a context-scoped audit tool that finds retired codenames only in load-bearing positions: token filenames, data attributes/selectors, and docs theme hrefs.

## GitHub Issue

Part of #546
Part of #531

## Test Plan

- [x] `pnpm test:unit packages/core/scripts/lib/__tests__/mode-rename-map.test.js packages/core/scripts/__tests__/audit-mode-codenames.test.js`
- [x] `pnpm --filter @nexus/core capture:mode-values`
- [x] `pnpm --filter @nexus/core typecheck`
- [x] `pnpm test:unit`
- [x] `pnpm --filter @nexus/core audit:mode-codenames` reports the expected pre-migration violations and exits 1; this is a tool-only check until the rename lands and PR 3 promotes it to a gate.

### DIFF
--- a/apps/docs/app/_lib/theme-modes.test.ts
+++ b/apps/docs/app/_lib/theme-modes.test.ts
@@ -1,4 +1,4 @@
-import { existsSync } from 'node:fs';
+import { existsSync, readdirSync } from 'node:fs';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { describe, expect, it } from 'vitest';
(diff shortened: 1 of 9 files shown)


aishvaryagupta:
Accepted as a PR-2 chore: extract one shared token-mode discovery manifest and bind the audit/capture/docs/core guards to it. Deferring because #549 is PR-1 inventory/test hardening only; adding a shared core manifest now would expand the public/internal API surface outside this PR's scope.

Issue #620, opened 2026-07-06
feat(components): add Combobox and MultiSelect primitives


aishvaryagupta:
## Problem

Nexus currently has adjacent primitives for single-select fields, command-style filtering, popovers, checkboxes, badges, fields, and input groups, but it does not expose first-class components for:

- `Combobox`: editable/searchable single selection.
- `MultiSelect`: searchable multiple selection with clear selected-state affordances.

Current nearby building blocks:

- `packages/react/src/components/select/*` - Radix Select wrapper for single-value picking.
- `packages/react/src/components/native-select/*` - native select styling.
- `packages/react/src/components/command/*` - cmdk-powered command palette/list filtering.
- `packages/react/src/components/popover/*`, `checkbox/*`, `badge/*`, `field/*`, and `input-group/*` - likely composition pieces.

Without first-class components, app surfaces will keep rebuilding these patterns ad hoc, which risks inconsistent keyboard behavior, form semantics, token usage, and Storybook coverage.

## Research notes

- [WAI-ARIA APG Combobox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) frames combobox as an input with an associated popup. The popup may be a listbox, grid, tree, or dialog, and the implementation must define the autocomplete behavior, popup state, keyboard contract, active descendant handling, accessible naming, and selected option semantics.
- [WAI-ARIA APG Listbox Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) covers single-select and multi-select listboxes. For multi-select, focus and selection are distinct states, typeahead is recommended for longer lists, and the recommended multi-select model lets users navigate without holding modifier keys while toggling selection with `Space`.
- [React Aria ComboBox](https://react-aria.adobe.com/ComboBox) treats a combobox as a text input plus listbox and shows useful requirements for static/dynamic collections, sections, disabled items, empty/loading states, and filtering.
- [React Aria Selection](https://react-aria.adobe.com/selection) is a useful reference for controlled multiple selection, toggle-vs-replace behavior, explicit checkbox affordances, and touch-friendly multi-selection.
[...]

PR #632, opened 2026-07-07
feat(react): add combobox component

aishvaryagupta:
## Summary

- Adds a clean Combobox component with a shadcn-style items API, field sizing, invalid and disabled states, grouped items, and a controlled value held at the call site.
- Exposes the reference anatomy: Combobox, ComboboxTrigger, ComboboxInput, ComboboxContent, ComboboxEmpty, ComboboxList, ComboboxGroup, ComboboxSeparator, and ComboboxItem.
- Positions a Radix Popover over the cmdk-powered Command list — cmdk owns filtering, keyboard navigation, active-descendant tracking, and ARIA; the selected value and open state are controlled at the call site (the composition/shadcn model).
- Adds focused Storybook coverage for default, controlled, sizes, grouped, disabled, invalid, keyboard, click, clear, reset, and data attributes.

## GitHub Issue

Part of #620. This split PR intentionally does not auto-close the shared Combobox + MultiSelect tracker.

## Test Plan

- [x] corepack pnpm --filter @nexus_ds/core build
- [x] corepack pnpm --filter @nexus_ds/react typecheck
- [x] corepack pnpm lint
- [x] corepack pnpm format:check
- [x] corepack pnpm test:storybook packages/react/src/components/combobox/Combobox.stories.tsx

### DIFF
--- a/packages/react/src/components/combobox/Combobox.stories.tsx
+++ b/packages/react/src/components/combobox/Combobox.stories.tsx
@@ -24,8 +24,9 @@ type ComboboxItemData = {
   group?: string;
 };
 
-// cmdk lowercases the value passed to onSelect, so keep option values lowercase
-// and compare against them directly.
+// cmdk filters on each item's `value` + `keywords`, not its rendered children,
+// so pass the label as a keyword to make it searchable. `onSelect` receives the
+// trimmed `value` (case preserved) — compare against `option.value` directly.
 const frameworks: ComboboxItemData[] = [
   { value: 'next', label: 'Next.js' },
   { value: 'sveltekit', label: 'SvelteKit' },
@@ -121,6 +122,7 @@ function ComboboxExample({
                 <ComboboxItem
                   key={option.value}
                   value={option.value}
+                  keywords={[option.label]}
                   disabled={option.disabled}
                   selected={value === option.value}
                   onSelect={handleSelect}
@@ -185,6 +187,7 @@ export const Controlled: Story = {
                   <ComboboxItem
                     key={framework.value}
                     value={framework.value}
+                    keywords={[framework.label]}
                     selected={value === framework.value}
                     onSelect={handleSelect}
                   >
@@ -276,7 +279,7 @@ export const InvalidField: Story = {
         Choose the framework used by this project.
       </FieldDescription>
       <FieldError id="framework-invalid-error">
-        Framework is required.
+        Select a supported framework.
       </FieldError>
     </Field>
   ),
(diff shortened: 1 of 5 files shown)


INNOVATIVEGAMER:
Thanks for the reviews. I've pushed the genuine fixes; a few of the "blocking" items are based on a framing I want to correct rather than implement.

## What was actually fixed

- **Label not searchable (SDE2 #1)** — real bug. cmdk scores the query against `value` + `keywords`, never the rendered children, so typing `Next.js` matched nothing. Options now pass `keywords={[label]}` and the visible label is searchable. Documented on `ComboboxItem`.
- **Stale cmdk `onSelect` docs (SDE2 #2)** — corrected. In cmdk `^1.1.1` `onSelect` receives the `.trim()`ed value, case preserved (the lowercasing was 0.2.x). The JSDoc and the story comment no longer tell callers to compare lowercase.
- **`data-slot` leak (Architect #4)** — the inner Command root now emits `data-slot="combobox"` instead of `command`. (The input-bar `command-input-wrapper` is cmdk's internal layout node, not part of the public Combobox anatomy, so it stays.)
- **Inline design-rationale comment (SDE2 #3)** — removed; rationale for the transparent Command surface lives here in the PR, per `code-comments.md`.
- **PR description "instead of cmdk" (both reviews)** — corrected. The description was backwards: the component **is** built on cmdk (`Command` / `CommandInput` / `CommandItem`). Fixed the summary.

## What I'm not implementing, and why

This Combobox is intentionally a **composition primitive** (Popover + cmdk), not a value-owning form control. That's a deliberate design choice, not a gap:

- **"No value model" (Architect #1)** — the value model is the standard composition one: the caller holds the selected `value` and wires `selected` + `onSelect` per item (see the `Controlled` story). This is the shadcn combobox idiom and matches Nexus's composition-over-configuration rules. Root-owned `value`/`onValueChange` + a hidden `<input>` for native form submission is explicitly **out of scope** for this primitive — not deferred.
- **"required" / "read-only" unimplemented (Architect #2, #5)** — these are consumer-composition concerns here, wired at the call site (`aria-required` on the field/label; a disabled or plain-text trigger for read-only), not root props on a composition component. I've dropped the stray "required" wording from the summary and the invalid-field demo copy so nothing over-claims a wired feature.
- **#620 framing** — #620 is the epic tracker for the eventual full Combobox **and** MultiSelect surface. Its full-component form ACs (root value ownership, hidden input, required/read-only props) are not obligations this composition PR is failing to meet or deferring — they're a different, larger component shape. I'm not citing #620 for "deferred" work because there isn't any.

## `aria-invalid` on the trigger vs. the popover input (Architect #3)

Intentional. The trigger button is the control in the field's tab order and the thing the `Field` label/error describe; the popover input is a transient **search** affordance, and marking a search field `aria-invalid` mislabels it. Callers who want the linkage in the open state can still pass `aria-invalid` / `aria-describedby` straight to `ComboboxInput` (it forwards them), but the field-level invalid state correctly lives on the trigger.

Typecheck, lint, and the 15 Combobox stories are green.

PR #633, opened 2026-07-07
feat(react): add multi-select component

aishvaryagupta:
## Summary

- Adds a MultiSelect component with chip display/removal, grouped options, a searchable popover, disabled/invalid states, and controlled/uncontrolled selection.
- Aligns the public anatomy with the WDS reference by exposing MultiSelectTrigger plus MultiSelectValue composition.
- Built on Radix Popover + cmdk (consistent with Combobox #632). Radix owns focus, dismissal, and open/close motion; chip labels are read from the composed children so preselected chips stay labeled while the list is closed — no always-mounted shadow option tree.
- Accessibility: multi-select listbox semantics are **not yet solved** and are tracked in #634 — cmdk binds `aria-selected` to the active descendant, not the checked options, so screen readers get no reliable per-option checked-state signal. Arbitrary-chip keyboard removal (beyond Backspace-removes-last) is tracked in the same issue.
- Adds focused Storybook coverage for default, controlled, grouped, disabled, disabled-option, invalid field, search, keyboard (incl. focus-on-open), chip removal (pointer + Backspace), overflow, long labels, and data attributes.

## GitHub Issue

Part of #620. This split PR intentionally does not auto-close the shared Combobox + MultiSelect tracker.

## Test Plan

- [x] corepack pnpm --filter @nexus_ds/react typecheck
- [x] corepack pnpm lint
- [x] corepack pnpm format:check
- [x] corepack pnpm test:storybook packages/react/src/components/multi-select packages/react/src/components/popover packages/react/src/components/combobox

### DIFF
--- a/packages/react/src/components/multi-select/MultiSelect.stories.tsx
+++ b/packages/react/src/components/multi-select/MultiSelect.stories.tsx
@@ -84,40 +84,6 @@ export const Controlled: Story = {
   },
 };
 
-export const SingleSelect: Story = {
-  render: () => (
-    <MultiSelect single>
-      <MultiSelectTrigger aria-label="Framework" className="nx:w-80">
-        <MultiSelectValue placeholder="Select a framework" />
-      </MultiSelectTrigger>
-      <MultiSelectContent
-        searchPlaceholder="Search frameworks…"
-        emptyMessage="No frameworks found."
-      >
-        {FRAMEWORKS.map((option) => (
-          <MultiSelectItem key={option.value} value={option.value}>
-            {option.label}
-          </MultiSelectItem>
-        ))}
-      </MultiSelectContent>
-    </MultiSelect>
-  ),
-  play: async ({ canvasElement }) => {
-    const canvas = within(canvasElement);
-    const body = within(document.body);
-    const trigger = canvas.getByRole('button', { name: 'Framework' });
-
-    await userEvent.click(trigger);
-    await userEvent.click(await body.findByRole('option', { name: 'Vue' }));
-
-    // Single mode collapses to a plain label and closes the popover.
-    await expect(within(trigger).getByText('Vue')).toBeInTheDocument();
-    await waitFor(() =>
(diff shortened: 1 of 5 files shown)


INNOVATIVEGAMER:
## Follow-up review addressed

Pushed `cdcffb9` (and prior `04f0831`). Summary of changes since the last review:

### `forceMount` findings (blocking + minors) — resolved by reverting it
Reverted `forceMount` and the shared `popover.tsx` change, restoring Radix's popover lifecycle:
- **Focus-on-open (blocking):** focus now enters the search box on open — cmdk arrow/type nav is live without a manual click. Proven by a new assertion in `KeyboardInteraction` (`waitFor(expect(search).toHaveFocus())` + `keyboard('sve')`, which only reaches the input if it's focused).
- **Exit motion:** restored (`display:none` no longer nullifies it).
- **`popover.tsx` footgun:** removed.

Chip labels now come from `collectLabels(children)` — a pure walk of the composed JSX — so preselected/disabled/never-opened chips are labeled synchronously with **no shadow option tree** and no per-item registration effects.

### Accessibility — not solved, now tracked in #634
The multi-select listbox a11y semantics are **not** solved in this PR. cmdk binds `aria-selected` to the active descendant rather than the checked options, and the earlier `aria-live` mitigation only announced toggle *actions* (never per-option state), so it was removed rather than shipped as a misleading half-measure. Full `aria-multiselectable` listbox semantics + arbitrary-chip keyboard removal are tracked in **#634**. The PR description has been corrected to drop the false "explicit aria-multiselectable listbox" claim.

### Other findings
- `value` + `keywords` forwarded to `CommandItem` (duplicate-label / icon-only disambiguation).
- Overflow badge driven from state only.
- Removed the avoidable `as string` cast in the trigger key handler.
- Chip keyboard removal remains `Backspace`-removes-last (accepted token-field pattern); per-chip keyboard removal is folded into #634.

Verification: typecheck, lint, format clean; storybook green for multi-select (17), popover (8), combobox (15).

PR #642, opened 2026-07-09
feat(core): add token engine instrumentation

aishvaryagupta:
## Summary

- Re-export `apcaLc` and add the semantic token registry to the core runtime export allowlist.
- Extract the surface ladder/ramp constants into `surface-ladder.ts`, keeping Phase 1 additive and preserving existing dark raw-step behavior.
- Add shade-anchor ladder drift coverage, engine self-parity registry coverage, and the frozen engine matrix snapshot generator/fixture/test.
- Add the `Tokens/SurfaceLadder` Storybook explorer with live runtime values, APCA badges, and embedded Appearance settings.

## GitHub Issue

Not linked.

## Test Plan

- [x] `corepack pnpm lint-staged`
- [x] `corepack pnpm test:unit`
- [x] `corepack pnpm test:storybook`
- [x] `corepack pnpm format:check`


aishvaryagupta:
## Review follow-up — finding → fix map

All review findings addressed across two commits. Verification for both: `@nexus_ds/core` 255/255 unit tests, `core` + `react` typecheck, `eslint --max-warnings 0`, and `prettier --check` all clean; `audit:runtime-exports` clean (35 runtime value exports unchanged — only **type** exports moved).

### Principal Architect + SDE2 review — `7d0eb42`

| # | Finding | Fix |
|---|---------|-----|
| PA#1 | Non-surface ramps (`STATUS_RAMP` / `CHART_*` / `NEUTRAL`) living in the surface-named module | Split into new `static-ramps.ts` |
| PA#2 | Ramp re-export shim in `derive-theme.ts` existing only for two tests | Deleted the block; repointed both test imports at `static-ramps` |
| PA#3 | `surface-ladder.ts` importing `Mode` / `NexusSurfaceTone` *up* from `derive-theme` (type-only cycle) | Moved both types to `palette.ts` (shared leaf) and repointed every importer — `surface-ladder` is now a true leaf |
| PA#4 | `anchorToStep`'s `contrast` param unused by all callers + double-count footgun | Dropped the param; uses the `SURFACE_DELTA_AT_ANCHOR` constant |
| PA#5 | `ShadeAnchor` / `SurfaceSteps` exported publicly with no producer/consumer | Removed from the public API (kept `SurfaceToken`) |
| SDE2#1 | Tautological `X === X` loops in `surface-ladder.test.ts` | Removed both loops + now-unused imports; the legacy-freeze guard remains |
| SDE2#2 | Commented-out `SURFACE_STEP_OVERRIDES` block | Deleted |
| SDE2#3 | Dead `...PERCEPTUAL_L_GRID` spread | Dropped; anchor map built from the virtual steps only, so unknown shades fail fast |
| SDE2#4 | `comps` / `normalizeMap` / `sameValue` / `compactSnapshot` / `deriveMatrix` duplicated (and drifted) between test and generator | Extracted shared `scripts/engine-snapshot-matrix.mjs` (+ `.d.mts`); generator and test both import it |

### Follow-up review — `7457be9`

| # | Finding | Fix |
|---|---------|-----|
| 1 | Missing token could become `undefined` and silently drop from a regenerated fixture | Restored the `snapshotValue()` guard in the shared matrix module — throws on a missing `tone.mode.token` (the generator now gains a guard it never had) |
| 2 | Changeset advertised "shade-anchor ladder types," but those are no longer public | Reworded to "the public surface-token type" |
| 3 | Generator wrote raw `JSON.stringify` output → not `format:check`-idempotent with the committed (Prettier-formatted) fixture | Generator now Prettier-formats its output — verified it reproduces the committed fixture byte-for-byte |
| 4 | `SurfaceLadder` play function only smoke-checked one swatch | Now asserts the live OKLCH value and an `Lc` badge render (async finders handle the rAF timing; verified stable across repeated runs) |

PR #650, opened 2026-07-10
fix(react): remove dense control item stagger

### LINKED ISSUE #631: Refine staggered item motion in dropdown-style controls
## Problem

Dropdown-style controls currently use shared staggered item enter motion, where each line item appears with a small delay after the previous one. This feels decorative and slightly laggy in dense interaction surfaces.

The issue is visible in components such as:

- Dropdown
- Select
- Menubar
- Combobox / MultiSelect-style popovers, if they reuse the same motion helper

For these controls, users expect the menu/list content to appear as a stable, immediately scannable surface. Per-item stagger makes the list feel like it is still loading or flickering, especially when quickly opening and closing menus.

## Desired Direction

Keep subtle motion at the popover/content container level, but remove per-item stagger from dense menu/list controls.

Good default:

- The dropdown/popover content can fade, scale, or slide in subtly as a whole.
- Menu/list items should be visible immediately once the surface opens.
- Motion should continue to respect `prefers-reduced-motion`.

Avoid:

- Cascading per-row item delays in Select, Dropdown, Menubar, Combobox, and MultiSelect popovers.
- Shared stagger helpers being applied broadly to action/menu/form-list surfaces.

## Acceptance Criteria

- Select, Dropdown, and Menubar no longer show delayed per-item cascade on open.
- Combobox and MultiSelect popover options are checked and either avoid the same stagger or use the same no-stagger behavior if they share the helper.
- Any shared helper is renamed, split, or scoped so decorative stagger is not accidentally reused by dense control components.
- Storybook examples for affected components feel immediate and stable when opened repeatedly.
- Existing reduced-motion behavior remains supported.

## Notes

This is not about removing all motion. The preferred behavior is container-level motion only, with line items rendered immediately.

aishvaryagupta:
## Summary

- Remove per-item stagger motion from DropdownMenu, Select, Menubar, and ContextMenu while keeping container-level overlay motion intact.
- Replace stagger Storybook coverage with immediate item motion assertions across affected dense controls, Combobox, and MultiSelect.
- Keep an opt-in decorative stagger CSS utility for non-dense/card-style consumers under explicitly decorative names.

## GitHub Issue

Closes #631

## Test Plan

- [x] `PATH=/private/tmp/pnpm10-bin:$PATH CI=true pnpm typecheck`
- [x] `PATH=/private/tmp/pnpm10-bin:$PATH CI=true pnpm lint`
- [x] `PATH=/private/tmp/pnpm10-bin:$PATH CI=true pnpm test:storybook`
- [x] `pnpm audit:browser-support`
- [x] `git diff --check`
- [x] Grep confirmed old stagger helper/class/story names are gone from `packages/react/src`


aishvaryagupta:
Removed the decorative stagger utility (`motion.css` + its `index.css` import) in debe4e453 so this stays a pure removal PR.

As landed, that utility had **zero first-party consumers** — Tailwind v4 tree-shook it out of the built CSS entirely — plus **no Storybook coverage** and **no `prefers-reduced-motion` guard**. That made it an inert, untested, motion-policy-noncompliant orphan inside a removal PR, against `project-stage.md` ("delete rather than preserve; clean over safe").

The decorative opt-in stagger for card/showcase surfaces is tracked in #651, to return **with** a real consumer, a Storybook story, and a reduced-motion guard.

Issue #159, opened 2026-05-25
feat(tabs): animated sliding active indicator + motion-token foundation


aishvaryagupta:
## Summary

Give Tabs a **sliding active indicator** — the active cue (pill background / underline) animates *between* tabs instead of fading out-here / in-there per trigger. This is the signature polish in the Tier-A systems (Vercel, Linear, Raycast).

Split out of #93, which scoped narrowly to replacing the `shadow-sm` active cue with a segmented-control border. That fix is class-level; this is the larger enhancement and deserves its own design.

## Motivation

Per the Tier-A visual benchmark (Linear / Vercel / Geist / Raycast) and Emil Kowalski's design-engineering guidance, the *traveling* indicator is what makes tabs feel alive. Today our active indicator has no shared motion — it just toggles per trigger.

## Scope

1. **Motion-token foundation (prerequisite).** Nexus currently has **no** easing/duration tokens (verified against `nexus.css` / `variables.css`). Add a small motion scale:
   - easing: `--ease-out` (e.g. `cubic-bezier(0.23, 1, 0.32, 1)`), plus `--ease` / `--ease-in-out` as needed
   - duration: `--duration-fast` (~150ms), `--duration-default` (~200ms)
   - emit as Tailwind utilities; document in `tokens.md` (new "Motion" semantic category).
2. **Sliding indicator in Tabs.**
   - Shared indicator element positioned/sized to the active trigger (ref + `ResizeObserver` measurement — a legit external-sync `useEffect`).
   - Animate via **CSS transform** (off the main thread) — *not* Framer Motion shared-layout, per Vercel's own lesson that JS layout animation dropped frames under page-load.
   - Apply to both `default` (pill bg) and `underline` (bottom-border) variants.
3. **Accessibility.** Respect `prefers-reduced-motion` — drop the slide, keep the instant state change.

## Non-goals

- Bounce/spring flourish — tab-switching is frequent, so keep it fast and subtle (<200ms).
- Per-variant easing — one curve for the system.

## References

- Emil Kowalski — Good vs Great Animations: https://emilkowal.ski/ui/good-vs-great-animations
- Vercel tabs recreation (Joshua Wootonn): https://www.joshuawootonn.com/vercel-tabs-component
- buildui — Animated Tabs: https://buildui.com/recipes/animated-tabs

## Relates to

- Follows #93 (segmented-control border active cue).

aishvaryagupta:
[...]

PR #669, opened 2026-08-19
feat(react): add Marker primitive for inline annotations and labelled dividers

### LINKED ISSUE #665: feat(components): add Marker primitive for inline annotations and labelled dividers
## Context

shadcn/ui ships a `marker` registry item (`@shadcn/marker`, `registry:ui`, depends on `radix-ui`) that renders a low-emphasis inline annotation row: an optional icon, muted text, and three layout variants — `default` (plain row), `separator` (centred text with a rule bleeding to both sides), and `border` (row with a bottom rule). Parts are `Marker`, `MarkerIcon`, `MarkerContent`, with `asChild` on the root and an exported `markerVariants`.

Nexus has no equivalent. The closest shipped pieces are `Separator` (a rule with no content slot) and `Badge` (an emphasis chip, wrong weight for this role). Consumers wanting a "Today", "Unread from here", "3 messages hidden", or "Edited" divider inside a list currently hand-roll a flex row with `border` and `muted-foreground` utilities, which puts the treatment outside the design system and produces inconsistent rules across surfaces.

Related component-primitive context: #658, #627.

## Goal

Add a small presentational Marker primitive for inline annotations and labelled dividers inside lists, feeds, and message streams.

## Scope

- Add the component under `packages/react/src/components/marker/` following the standard folder layout (`marker.tsx`, `Marker.stories.tsx`, `index.ts`).
- Cover the upstream part set — root, icon, content — with names and API confirmed against Nexus conventions rather than copied wholesale.
- Variants: evaluate `default` / `separator` / `border` against the Nexus variant model. `separator` is the one carrying real layout logic (the `::before` / `::after` flex rules); confirm it is built with tokenised border colour and survives a container with no free space.
- Token adaptation per `shadcn-divergences.md`: `text-muted-foreground` maps directly, `bg-border` → `border-default`. Decide the text tier deliberately — `muted-foreground` vs `muted-foreground-subtle` — since this is annotation-weight content sitting next to real content.
- Typography: the upstream `text-sm` must become a `nx:typography-*` composite, not a raw size utility (see the recurring findings in `COMPONENT-REVIEW.md`).
- Icon sizing follows the existing Nexus `[&_svg]` convention used by Button/Item rather than the upstream literal.
- `asChild` per the Nexus Slot pattern, so the row can render as a heading or a link when a consumer needs the semantics.
- Semantics: a decorative divider and a labelled section heading are different things to a screen reader. Decide whether the root should carry `role="separator"` in the `separator` variant, and make sure `MarkerIcon` stays `aria-hidden`.
- Add a sizing row to `components.md` § Sizing Convention in the same PR.
- Export the components, props types, and `markerVariants` from `packages/react/src/index.ts`.
- Storybook coverage: each variant, icon and no-icon, long text wrapping, a link inside content, and a marker sitting inside a real list to prove the rule aligns optically with surrounding rows.

## Acceptance criteria

- All styling uses `nx:`-prefixed semantic tokens with no raw primitives and no `dark:` modifiers on semantic tokens.
- Typography uses composites; no raw `text-sm` / `font-medium` survives.
- The `separator` variant's rules align to the same border colour and thickness as `Separator`, so the two never read as two different systems on one page.
- The annotation is announced sensibly — decorative rules are not read as content, and a labelled divider exposes its label.
- A sizing row for Marker lands in `components.md` in the same PR.
- Stories cover the required matrix from `testing-react.md`, including a composed-list scene.
- Typecheck, lint, and Storybook tests pass for the new exported component.

## Notes

Plan before implementing. **Sequence: 1 of 4** in this batch (Marker → Bubble → Message → MessageScroller). Fully independent — nothing else in the batch blocks it. Study `separator`, `item`, `empty-state`, and `badge` first — if `Separator` can absorb the labelled case behind an existing slot, that is a better outcome than a second component with an overlapping role, and the PR should say so explicitly either way.

Tier-A polish (optical rhythm of the rule against adjacent rows, composed-scene fit) is a separate gate per `polish.md`.

neha-tanti:
## Summary

- Adds `Marker`, `MarkerIcon`, and `MarkerContent` — a low-emphasis annotation row for lists, feeds, and message streams (optional icon + muted text).
- `variant="separator"` centres the label between two rules; `variant="border"` rests the row above a bottom rule; `default` is a plain row.
- `asChild` (Radix `Slot`) composes the row onto different semantics — a heading for a labelled section, or a link / button for an actionable row.
- The label stays real text in reading order, so it is announced where it appears. For a purely decorative rule with no label, `Separator` is still the right primitive.
- 14 stories, 5 with play functions: data attributes, `asChild` semantics (heading + link), click interaction, keyboard interaction, and the variant grid.
- Records the `Marker` sizing contract in `.claude/rules/components.md`.
- Adds a changeset (`@nexus_ds/react` minor).

## GitHub Issue

Closes #665

## Notes for review

**Why `Separator` cannot absorb the labelled case.** `Separator` wraps Radix `SeparatorPrimitive.Root`, which renders a single element carrying `role="separator"` when `decorative={false}`. ARIA defines `separator` as a children-presentational role, so any label nested inside it is stripped from the accessibility tree — the text would render but never be announced. Absorbing the labelled case would therefore mean giving `Separator` a second, non-separator rendering mode with its own two-rule flex layout and a label slot: one component with two layout systems, which #665 explicitly rules out.

**Why the `Marker` root does not carry `role="separator"`.** Same reason, from the other direction. The label is the entire point of the component — it must stay real text in reading order so it is announced where it appears. Putting `role="separator"` on the root would make its children presentational and silence exactly the content the component exists to convey. The row therefore stays a plain `div`, the two rules are `::before` / `::after` pseudo-elements (invisible to assistive tech by construction), and a genuinely label-less rule is routed to `Separator`, which owns the semantic role.

**The row answers `asChild` itself.** The base carries `nx:[a]:hover:text-foreground nx:[button]:hover:text-foreground`, compiling to `:is(a):hover` / `:is(button):hover`, so the hover affordance applies only when `asChild` actually makes the row interactive — a plain annotation row still does not light up under the cursor. The base also carries the canonical focus triple, so an interactive row gets the design-system ring rather than the UA default. This matches `Item`, which scopes its own hover the same way; interactive call sites pass no extra classes. Nested anchors inside `MarkerContent` keep their own treatment via `nx:[&_a:hover]:text-primary-subtle-foreground`, matching `Field`, `Item`, and `EmptyState`.

**Touch targets.** `nx:min-h-4` is correct for a non-interactive annotation, which has no hit area to size. The component owns the floor for the interactive case rather than delegating it: `nx:[a]:pointer-coarse:min-h-11` / `nx:[button]:pointer-coarse:min-h-11` compile to `:is(a)` / `:is(button)` inside a `pointer: coarse` query, so an `asChild` link or button clears ~44px on touch while a plain annotation row stays 16px. Same mechanism as `ChoiceRow`, scoped to the elements that are actually interactive.

**Adjacent wording fix.** The existing `ChoiceRow` / `ChoiceCard` sizing rows said "nested marker remains atom-sized", where "marker" meant the Checkbox / RadioGroupItem indicator. With a component now literally named `Marker`, that read ambiguously, so those two cells now say "nested selection marker". The rest of the table diff is prettier re-padding column widths to fit the new row.

**Polish gate not run.** This clears the build + accessibility gate. The Tier-A polish gate (`polish.md`) — Modern Web Guidance search, browser-floor decision, density/theme evidence — has not been run. `Marker` has no motion and no `@container` behaviour, so it should be short, but it is outstanding.

## Test Plan

- [x] `pnpm lint` — clean repo-wide
- [x] `tsc --noEmit` — clean
- [x] `pnpm --filter @nexus_ds/core audit:class-refs` — 439 files, all semantic colour refs resolve
- [x] `vitest --project=storybook packages/react/src/components/marker` — 15/15 pass (addon-a11y runs per story)
- [x] `pnpm --filter @nexus_ds/eslint-plugin refresh:canonical-set --check` — matches live union
- [x] `pnpm --filter @nexus_ds/react build` — `Marker` ships in the runtime bundle, the type barrel, and `dist/react.css` (verified `before\:bg-border-default`, `after\:bg-border-default`, `group\/marker` all compile)
- [x] Visual review of `Components/Marker` in Storybook, light and dark


neha-tanti:
## Follow-up review findings addressed (`d1e9ae7`)

| Finding | Fix |
| --- | --- |
| PA — `:not(:disabled)` misses `aria-disabled`; anchors can't take `disabled` | Base cva now carries `nx:disabled:pointer-events-none nx:aria-disabled:pointer-events-none`; the bespoke `[button:not(:disabled)]:hover` guard is gone |
| SDE2 #1 — `asChild` `<button>` gets no pointer cursor | `nx:[button]:cursor-pointer` |
| SDE2 #2 — tap floor and hover guard untested | `AsChildLink` asserts `nx:[a]:pointer-coarse:min-h-11`; new `Disabled` story covers a native-disabled button and an `aria-disabled` anchor |

**Scope note — this goes one step beyond the finding.** The review asked for the hover guard to exclude `[aria-disabled=true]`. Rather than grow the selector to `[button:not(:disabled):not([aria-disabled='true'])]`, the base adopts the package's existing disabled contract (`button.tsx:11`). `pointer-events: none` means the element is never hit-tested, so the hover and the cursor both fall away without a compound selector — and an `aria-disabled` `asChild` anchor becomes genuinely un-clickable rather than merely un-hovered, which is the behaviour `button.tsx:168-169` already documents as the repo's accessible-disabled path. Flagging it because it is a behaviour change the diff alone does not advertise.

`nx:disabled:cursor-default` / `nx:aria-disabled:cursor-default` were deliberately **not** added: under `pointer-events: none` the element is not hit-tested, so those rules would be dead classes of the kind flagged earlier in this PR (`before:min-w-0`).

## Verification

Emitted CSS grepped out of `dist/react.css` rather than asserted from memory:

- `.nx\:\[button\]\:cursor-pointer:is(button){cursor:pointer}`
- `.nx\:aria-disabled\:pointer-events-none[aria-disabled=true]{pointer-events:none}`
- `.nx\:disabled\:pointer-events-none:disabled{pointer-events:none}`
- `@media(hover:hover){.nx\:\[button\]\:hover\:text-foreground:is(button):hover{…}}`

The `Disabled` play function asserts computed `pointer-events: none` on both rows — the storybook project runs in real chromium, so this is a live-CSS assertion, not a class-string check. `user-event` refuses to click through `pointer-events: none`, so the click that proves the handler stays silent passes `pointerEventsCheck: 0`.

- [x] `pnpm lint` — clean repo-wide
- [x] `tsc --noEmit` (`@nexus_ds/react`) — clean
- [x] `vitest run --project=storybook packages/react/src/components/marker` — 15/15 pass
- [x] `pnpm --filter @nexus_ds/core audit:class-refs` — 439 files, all refs resolve
- [x] `pnpm --filter @nexus_ds/react build` — all seven interactive-contract rules ship in `dist/react.css`

The Tier-A polish gate (`polish.md`) remains outstanding, as noted in the PR body.

neha-tanti:
## Tier-A Polish Gate — `Marker` (`2e86936`)

Closes the gate the PR body flagged as outstanding.

### Modern Web Guidance Findings

```
npx -y modern-web-guidance@latest search \
  "labelled divider row with pseudo-element rules and interactive link semantics" \
  --skill-version 2026_05_16-c5e7870
```

Ran successfully. Guides returned: `css-layout` (0.404), `forms` (0.356), `accessibility` (0.350), `custom-select-picker-layouts` (0.313). Retrieved `accessibility`.

The tool emits a version warning — it reports our pinned `2026_05_16-c5e7870` as out of date against `2026_05_16-c5e78707`. That is a truncated-hash mismatch on the same dated release, not a stale skill, so the results were used as-is. Worth correcting the pin repo-wide in a separate change.

**Load-bearing finding from `accessibility`:** *"Be deliberate about `disabled` vs `aria-disabled`: `disabled` removes the element from the focus order entirely… `aria-disabled="true"` keeps the element focusable so users can land on it and learn it's disabled."* See Deliberately Unpolished below.

### Findings fixed

| # | Matrix area | Finding | Fix |
| --- | --- | --- | --- |
| 1 | State coverage | A `disabled` / `aria-disabled` row was inert but rendered **identically to an enabled one** — the base is already `text-muted-foreground`, so removing interactivity changed nothing visually | `nx:disabled:text-disabled-foreground nx:aria-disabled:text-disabled-foreground` (precedent: `sidebar.tsx:791`, `button.tsx` ghost) |
| 2 | Theme / browser floor | `variant="separator"` paints its two rules with `background-color` on `::before` / `::after`. The forced-color adjustment flattens that to `Canvas`, so **the divider disappeared in Windows High Contrast Mode while the label stayed** — the component reduced to floating text | `nx:forced-colors:before:bg-[CanvasText]` / `nx:forced-colors:after:bg-[CanvasText]` |

**Finding 2 also applies to `Separator`,** which paints the identical `nx:bg-border-default` hairline and had the same gap. #665 requires the two never diverge into separate systems, so `Separator` takes the same one-class fallback. That is the only file outside `marker/` touched, and it is a strict bug fix, not a refactor. Precedent for the technique is in-repo: `scroll-area.tsx:134` already does exactly this for its thumb, and its JSDoc frames it as owning chrome Nexus painted itself.

### Browser-Floor Decision

`@media (forced-colors: active)` and the `CanvasText` system colour are the only browser-platform features added. Chrome/Edge 89+, Firefox 89+ — inside the floor. **Safari does not implement `forced-colors` at all**, which is the correct failure mode: Safari users get the normal `bg-border-default` render, unchanged from today.

`pnpm audit:browser-support` — all checks passed.

### Fallback / Progressive Enhancement

Purely additive. The rule is a no-op outside HCM; the normal-render appearance is byte-identical to before. Nothing is removed from the non-forced-colors path, so there is no degradation to fall back from.

### Rejected Approaches

- **`border-top` instead of `background-color` for the rules.** Forced-colors preserves border colours, so this would fix HCM without a media query — but it changes the shipped rendering of both `Marker` and `Separator` and breaks the `nx:h-px` + `nx:bg-border-default` pairing that #665 pins the two components to. A media query that touches only HCM is the smaller change.
- **A global `[data-slot=marker]` rule in the `forced-colors` block of `nexus.css`.** That block exists for focus box-shadows that forced-colors strips wholesale. A single component's hairline does not warrant leaving the utility layer.
- **Scoping the disabled colour to `[a]` / `[button]`** as the hover and tap-floor rules are scoped. Unnecessary: `:disabled` and `[aria-disabled=true]` are already self-scoping, and a non-interactive row carrying `aria-disabled` should mute too.

### Motion

No change. `nx:transition-colors nx:duration-faster` is a colour transition with no movement, so `prefers-reduced-motion` has nothing to reduce — `polish.md`'s policy targets movement. No new motion introduced, so #159 stays unblocked.

### Validation Evidence

- [x] `pnpm lint` — clean repo-wide
- [x] `tsc --noEmit` (`@nexus_ds/react`) — clean
- [x] `vitest --project=storybook` on `marker` + `separator` — 21/21 pass (16 Marker, 5 Separator; addon-a11y per story)
- [x] `pnpm audit:browser-support` — all checks passed
- [x] `pnpm --filter @nexus_ds/core audit:class-refs` — 439 files, all refs resolve
- [x] `pnpm --filter @nexus_ds/react build` — new rules verified in `dist/react.css`, and confirmed nested inside `@media(forced-colors:active)` rather than emitted unconditionally

New `ForcedColors` story; `Disabled` extended to assert the muted colour.

### Deliberately Unpolished

- **`tabIndex={-1}` on the `aria-disabled` anchor in the `Disabled` story.** MWG's `accessibility` guide argues the opposite — keeping an `aria-disabled` element focusable is usually better, so a screen-reader user can land on it and *learn* it is disabled rather than have it silently vanish from the tab order. `button.tsx:169` documents `tabIndex={-1}` as this repo's established `asChild` disabled pattern, and the Marker story mirrors it for consistency. Changing it is a Button-wide decision, not Marker's to make unilaterally. Flagging it rather than silently picking a side.
- **No press/active state** on an interactive row. Hover → `foreground` is sufficient feedback for a 16px annotation; a press cue would be decoration.
- **No density or `@container` behaviour.** `Marker` is single-size by design (no `size` prop) and has no internal responsive behaviour, so there is no density matrix to evidence.

neha-tanti:
## Review round addressed (`a0c5061`)

Both reviews were submitted against `d1e9ae7`, one commit before the polish gate landed in `2e86936` — so PA #1 was already fixed when the review posted.

| # | Finding | Status |
| --- | --- | --- |
| PA 1 | ❌ Disabled row keeps its enabled colour | Already fixed in `2e86936` — `nx:disabled:text-disabled-foreground nx:aria-disabled:text-disabled-foreground`, same pair you cite from `sidebar.tsx:791` |
| PA 2 | ⚠️ Interaction contract in the Sizing Convention row | Fixed — row is back to sizing facts only; the contract lives in the `asChild` JSDoc |
| SDE2 1 | ⚠️ `toHaveClass` passes on every Marker | Fixed — replaced with a computed-cursor proof |
| SDE2 2 | ⚠️ `Disabled` asserts the fixture's own props | Fixed — dropped; the colour comparison replaces them |

### SDE2 #1 — the non-discriminating assertion, generalised

The finding is right, and it condemned **three more assertions than it named** — all three added in `2e86936`, all the same defect: `toHaveClass` against a string that is unconditionally in the base cva. `nx:aria-disabled:text-disabled-foreground` in `Disabled`, and both `nx:forced-colors:*:bg-[CanvasText]` in `ForcedColors`. Fixed alongside rather than left to a later round.

**Cursor proof.** `AsChildLink` was the wrong host for it — an `<a href>` gets `cursor: pointer` from the UA, so the assertion cannot discriminate there either. A bare `<button>` has **no** UA pointer cursor, so the pair that actually tests `:is(button)` scoping is:

- `ClickInteraction` — `await expect(button).toHaveStyle({ cursor: 'pointer' })` — only our rule can produce this
- `AsChild` (the `h3`) — `await expect(heading).not.toHaveStyle({ cursor: 'pointer' })` — same class string, no cursor

**Disabled colour.** `Disabled` now renders an **enabled** row alongside the two disabled ones and compares computed `color`, so the assertion tests the token swap rather than a string.

**ForcedColors.** HCM cannot be emulated from a play function, so the class assertions were theatre. The play function now asserts the `::before` / `::after` rules actually paint (equal, non-transparent `backgroundColor`); the HCM behaviour itself is documented as visual evidence, which is the honest scope.

### Mutation-verified

A green test proves nothing unless it can go red, so the new assertions were checked by deleting the classes they cover:

```
removed: nx:[button]:cursor-pointer, nx:disabled:text-disabled-foreground,
         nx:aria-disabled:text-disabled-foreground
result:  2 failed | 14 passed  (Disabled, Click Interaction)
restored: 16 passed
```

The assertions they replace would have stayed green through that same deletion.

### Validation

- [x] `pnpm lint` — clean repo-wide
- [x] `tsc --noEmit` (`@nexus_ds/react`) — no `marker` diagnostics
- [x] `vitest run --project=storybook packages/react/src/components/marker` — 16/16 pass
- [x] `pnpm --filter @nexus_ds/core audit:class-refs` — 439 files, all refs resolve

Re: *"`vitest run --project=storybook` was not run — a cold Storybook browser build risks exceeding the command timeout."* It completes here in ~19s cold, ~5s warm.

neha-tanti:
**Correction to my previous comment.** I wrote that both reviews were submitted against `d1e9ae7`, one commit before the polish gate. That is wrong — GitHub records their head commit as `2e86936`, and the timestamps agree: `2e86936` was pushed at 09:19Z and the reviews landed at 09:23Z.

What is accurate is narrower: the review bodies state they analysed `d1e9ae77` (SDE2 says so explicitly), so the analysis ran against a stale checkout while the PR head had already moved. That is why PA #1 flagged `nx:disabled:text-disabled-foreground` as missing when `2e86936` had already added it.

The fixes and their verification are unaffected — only my stated reason for PA #1 already being resolved was wrong.

neha-tanti (commented):
## 🤖 Principal Architect Follow-up Review

### Verdict: CHANGES REQUIRED

### Unresolved Issues

All previous issues resolved — the disabled foreground swap landed in `2e86936`, and the Sizing Convention row is back to sizing facts only.

### New Issues

| 1 | ❌ Blocking | `marker.tsx:15` | Six other `bg-*` hairline dividers left without the HCM fallback | Extend to `command`, `context-menu`, `dropdown-menu`, `menubar`, `select`, `resizable` |
| 2 | ⚠️ Minor | `marker.tsx:65` | Marker JSDoc documents `Separator`'s implementation; unenforced sync claim | Move the fallback note into `separator.tsx`'s own JSDoc |
| 3 | ⚠️ Minor | `separator.tsx:52` | Modified component ships zero story evidence for the change | Mirror `ScrollArea.stories.tsx:233`'s class assertion (inline comment merged into the SDE2 review) |
| 4 | ⚠️ Minor | `scripts/audit-browser-support.js:42` | `forced-colors` absent from `FEATURE_POLICIES` registry | Add as `progressive-enhancement` per `polish.md` |
| 5 | ⚠️ Minor | `marker.tsx:40` | "inert" overclaims — `aria-disabled` anchors stay keyboard-activatable | Scope the wording to pointer affordances; say the call site owns `tabIndex={-1}` |

### Architectural Concerns

**Forced-colors fallback applied per-component** (`marker.tsx:15`) — Pulling `Separator` into this PR is justified by #665's "never read as two different systems" criterion, but that criterion does not stop at `Separator`. Either cover every `bg-*` hairline divider in this PR, or move the fallback to the design-system CSS layer that already emits `@media (forced-colors: active)` in `packages/core/scripts/utils.js:2155`.

---

packages/react/src/components/marker/marker.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { cn } from '../../lib/utils';
+
+const markerVariants = cva(
+  "nx:group/marker nx:flex nx:min-h-4 nx:items-center nx:gap-2 nx:text-left nx:typography-body-small nx:text-muted-foreground nx:transition-colors nx:duration-faster nx:focus-visible:outline-2 nx:focus-visible:outline-focus-default nx:focus-visible:outline-offset-(--focus-offset) nx:disabled:pointer-events-none nx:disabled:text-disabled-foreground nx:aria-disabled:pointer-events-none nx:aria-disabled:text-disabled-foreground nx:[&_svg:not([class*='size-'])]:size-4 nx:[&_svg]:shrink-0 nx:[a]:pointer-coarse:min-h-11 nx:[a]:hover:text-foreground nx:[button]:cursor-pointer nx:[button]:pointer-coarse:min-h-11 nx:[button]:hover:text-foreground",
+  {
+    variants: {
+      variant: {
+        default: '',
+        separator:
+          'nx:w-full nx:before:mr-1 nx:before:h-px nx:before:flex-1 nx:before:bg-border-default nx:after:ml-1 nx:after:h-px nx:after:flex-1 nx:after:bg-border-default nx:forced-colors:before:bg-[CanvasText] nx:forced-colors:after:bg-[CanvasText]',
```
neha-tanti, line 15: **Blocking — partial system fix:** Six other `bg-*` hairline dividers still vanish in HCM — separators in `command:317`, `context-menu:442`, `dropdown-menu:447`, `menubar:491`, `select:336`, plus `resizable:66`. The parity argument that pulled `Separator` into scope covers them too. Extend the fallback to all of them here.

packages/react/src/components/marker/marker.tsx
```diff
@@ -0,0 +1,165 @@
...
+ * Marker
+ *
+ * A low-emphasis annotation row for lists, feeds, and message streams — an
+ * optional icon plus muted text. `variant="separator"` centres the label
+ * between two rules; `variant="border"` rests the row above a bottom rule.
+ *
+ * The label stays real text in reading order, so it is announced where it
+ * appears. For a purely decorative rule with no label, use `Separator`.
+ *
+ * The `separator` rules are `background-color` hairlines, which the forced-color
```
neha-tanti, line 65: **Minor — cross-component doc:** This paragraph documents `Separator`'s implementation from `Marker`'s JSDoc, creating an unenforced sync claim that `code-comments.md` bans. Move the fallback note to `separator.tsx`'s own JSDoc and drop the `Separator` sentence here.

packages/react/src/components/marker/marker.tsx
```diff
@@ -0,0 +1,165 @@
...
+  extends React.ComponentProps<'div'>, VariantProps<typeof markerVariants> {
+  /**
+   * Render as the child element instead of a `div`, merging Marker's styles
+   * onto it. Use when the annotation needs different semantics — a heading for
+   * a labelled section, or a link / button for an actionable row.
+   *
+   * An `a` or `button` child is answered by the row itself — hover, the
+   * pointer cursor, the design-system focus ring, and a coarse-pointer tap
+   * floor apply only when the child is interactive, so the call site needs no
+   * extra classes. A `disabled` or `aria-disabled` child is inert: the row
```
neha-tanti, line 40: **Minor — contract overclaims:** `aria-disabled` does not make an `a` inert — it stays focusable and Enter-activatable. The `Disabled` story adds `tabIndex={-1}` to compensate, which this doc says is unnecessary. Scope the wording to pointer affordances and colour, and state that the call site owns keyboard removal.

PR #682, opened 2026-09-02
feat(docs): add rehype-slug so headings get stable ids

### LINKED ISSUE #674: feat(docs): add rehype-slug so headings get stable ids
## Context

Headings on MDX pages have no `id` attributes, so nothing can deep-link to a section and nothing can build a table of contents from the page.

This is also what blocks the right rail: `RightRail` is currently a stub that renders the literal string `[ Anchor ]` three times. It cannot be filled with real anchors until heading ids exist.

## Goal

Every heading on an MDX page gets a stable, unique `id`.

## Scope

- Add `rehype-slug` to the MDX pipeline in `apps/docs/next.config.ts`.
- Register it by **string name, not by import** — same Turbopack constraint as `remark-gfm`.
- Verify ids are unique within a page and stable across builds of unchanged content.
- Confirm deep links (`#some-heading`) land on the heading rather than underneath the sticky TopNav — a `scroll-margin-top` may be needed.

## Acceptance criteria

- Every `h2` and `h3` on an MDX page has an `id`.
- Linking to `#id` scrolls to the correct heading, fully visible below the sticky header.
- Ids do not change between two builds of unchanged content.

## Notes

Ticket **2 of 7 in Stage 1** of the docs build order. Independent.

Unblocks the RightRail table-of-contents work later in the plan; that ticket assumes heading ids already exist.

neha-tanti:
## Summary

- Register `rehype-slug` in the docs MDX pipeline so every Markdown heading renders with a slugged `id`. It is registered by **string name**, not by import: `@next/mdx`'s loader resolves plugin strings itself (`require.resolve` from the `.mdx` file's own directory), and a plugin *function* cannot cross Turbopack's serialized loader-options boundary. Same constraint `remark-gfm` already lives under.
- Extract the MDX plugin options into `apps/docs/mdx-options.ts` so the test asserts against the same object `next.config.ts` ships, rather than a second copy of the literal.
- Offset the scroll port with one `nx:scroll-pt-20` on `<html>` so `#hash` links land below the sticky TopNav, and stop the nav wrapping so that single offset is correct at every width.
- Pin the resulting ids in a unit test. Heading ids are permalinks; the test makes a rename or a slugger bump show up as a failing diff instead of a silently broken deep link.

## GitHub Issue

Closes #674

## Test Plan

- [x] `pnpm build --filter=@nexus_ds/docs` → `id="the-model"` and siblings present in `apps/docs/.next/server/app/theming/appearance.html`
- [x] `next dev --turbopack` → same ids in the served DOM. `next build` takes the webpack branch, so it would pass even with an imported plugin function; the dev run is what actually exercises the constraint this PR is built around.
- [x] Anchor landing measured in headless Chromium at both reference widths, on cold load with a hash **and** on in-page anchor click after client-side nav:

  | Width | Path | Header | Heading top |
  | --- | --- | --- | --- |
  | 390px | cold load `#brand-and-surface-tone` | 57px | 80px |
  | 390px | in-page anchor click | 57px | 80px |
  | 1280px | cold load | 57px | 80px |
  | 1280px | in-page anchor click | 57px | 80px |

- [x] `pnpm test:unit` → 5 new tests pass. Mutation-checked: removing `'rehype-slug'` from `MDX_OPTIONS` fails all 5.
- [x] `pnpm lint` clean, `pnpm --filter @nexus_ds/docs typecheck` clean.

## Notes for review

**Why the TopNav changed.** The nine nav links were `flex-wrap`. At the 390px mobile reference width they wrapped to several rows, making the sticky header far taller than any fixed offset — a deep link would have landed *under* the header at a first-class target width. Switching the nav to `overflow-x-auto` with `nx:shrink-0` links keeps the header a constant 57px, which is what makes one scroll-padding value correct everywhere. It is the one edit in this PR that goes beyond "add a plugin", and it is here because AC #2 fails at 390px without it.

**Why 80px and not 64px.** `LeftRail` and `RightRail` both sit flush under the header at `nx:top-16` (64px). A heading arriving at a deep link wants clearance, not flush contact, so the scroll port uses 80px — 23px of air below a 57px header. That deliberately leaves four header-height literals in the app (64/64/80/80). Consolidating them behind one custom property was considered and rejected: flush and clearance are not the same number, so a single variable would need `calc()` at three of the four sites, and converting working Tailwind utilities into var plumbing is scope past this issue. Worth revisiting when the RightRail TOC work needs the number a fifth time.

**Why scroll-padding on `<html>` rather than `scroll-mt` per heading.** One rule on the scroll container instead of one class per anchorable element, and it keeps working for anchors that do not come from `mdx-components.tsx`. Note `--nx-spacing-20` is density-scaled (76px compact / 80px default / 84px relaxed); every step clears the header, which scales with it.

**Why the test pins values instead of asserting behaviour.** Asserting "ids are unique" or "two compiles agree" only restates `github-slugger`'s contract — those assertions cannot fail. What can fail is the id *values* changing under a dependency bump or a heading rename, which silently breaks external links. So the test carries a literal path→ids map for the three registered content files. It also uses a literal input set rather than a glob, so it cannot pass vacuously, and it resolves the plugin strings the way the loader does rather than from its own location.

**`footnote-label` in the expected list.** `theme-setup.mdx` uses a GFM footnote, and remark-rehype emits its own `<h2 id="footnote-label">`. rehype-slug skips headings that already carry an id, so that id is in the pinned list but is not slug-generated. The uniqueness assertion covers every `id` in the tree, not just heading ids, so a heading titled "Footnote label" colliding with it would be caught.

**Known constraint for the RightRail TOC ticket.** `## 1. Import the styles` slugs to `id="1-import-the-styles"`. That is a valid HTML id and native fragment navigation works, but `document.querySelector('#1-import-the-styles')` throws `SyntaxError` — any future id→element lookup must use `getElementById` or `CSS.escape`. Both leading-digit ids are pinned in the test.

**Deliberately not done.** No `rehype-autolink-headings` (clickable heading anchors) and no RightRail work — the issue scopes both to later tickets, and the RightRail one explicitly depends on this landing first.

**Pre-existing test failure, unrelated to this branch.** `scripts/audit-agent-drift.test.js` fails on `main` today, reporting `.claude` ↔ `.codex` agent-definition drift (`contrast-auditor`, `devops`, `principal-architect`, `sde2`, `tester`) plus a missing `modern-web-guidance` skill mirror. This branch touches nothing under `.claude/`, `.codex/`, or `.agents/`.


apps/docs/mdx-options.ts
```diff
@@ -0,0 +1,7 @@
+export const MDX_OPTIONS = {
```
neha-tanti, line 3: **Blocking — untyped config object.** Extracting to a variable dropped the excess-property check the inline literal in `next.config.ts` had. Verified: a typo'd `remarkRehypeOption` errors inline but compiles through the variable, silently dropping the sr-only footnote class. Annotate with `NextMDXOptions['options']` from `@next/mdx`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -46,7 +46,7 @@ export function TopNav() {
           docs
         </span>
       </div>
-      <nav className="nx:flex nx:gap-0.5 nx:flex-1 nx:flex-wrap">
+      <nav className="nx:flex nx:gap-0.5 nx:flex-1 nx:min-w-0 nx:overflow-x-auto">
```
neha-tanti: **Blocking — mobile nav regression.** At the 390px reference width this strip is ~117px once the brand, gaps, and theme toggle take their share, so nine links become a horizontal drag with the active item off-screen and no overflow affordance. Under `flex-wrap` they all stayed reachable.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -55,8 +55,8 @@ export function TopNav() {
               href={s.href}
               className={
                 active
-                  ? 'nx:px-3 nx:py-1 nx:typography-label-default nx:rounded-sm nx:bg-nav-item-active nx:text-primary-subtle-foreground'
-                  : 'nx:px-3 nx:py-1 nx:typography-label-default nx:rounded-sm nx:text-nav-muted-foreground nx:hover:text-nav-foreground nx:hover:bg-nav-item-hover'
+                  ? 'nx:px-3 nx:py-1 nx:shrink-0 nx:typography-label-default nx:rounded-sm nx:bg-nav-item-active nx:text-primary-subtle-foreground'
```
neha-tanti: **Minor — clipped focus outline.** `overflow-x-auto` on the parent makes `overflow-y` compute to `auto`, so these links' focus outline is clipped by the scroll container. Add explicit `focus-visible` outline classes and vertical room on the nav.

apps/docs/app/layout.tsx
```diff
@@ -30,6 +30,7 @@ export default function RootLayout({
   return (
     <html
       lang="en"
+      className="nx:scroll-pt-20"
```
neha-tanti: **Blocking — magic scroll offset.** 80px is the fourth independent encoding of header height (LeftRail `top-16` and `max-h-[calc(100svh-80px)]`, RightRail `top-16`), while the header actually measures 57px. Set `--docs-header-h` on the header element and consume it in `scroll-padding-top` and both rails.

apps/docs/app/_lib/mdx-pipeline.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    'install-the-published-packages',
+    'vendor-the-components-and-theme',
+    'wire-the-styles',
+    'declare-native-color-scheme',
+    'render-a-component',
+    'next-steps',
+  ],
+  'getting-started/theme-setup.mdx': [
+    'theme-setup',
+    '1-import-the-styles',
```
neha-tanti: **Minor — leading-digit id.** `1-import-the-styles` is not a valid CSS selector, and this pins it as expected while deferring the constraint to an uncited ticket. Drop the `1.` / `2.` numbering from those two H2s in `theme-setup.mdx`.

neha-tanti (commented):

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Architecture | ⚠️ | Scroll-offset contract is inverted; header height encoded in four places |
| Code Quality | ⚠️ | Extracted config object lost its type constraint |
| Testing | ⚠️ | Pinned id map has no link to the content directory |

### Verdict: CHANGES REQUESTED

### Issues

| 1 | ❌ Blocking | `apps/docs/mdx-options.ts:1` | Untyped export drops the excess-property check the inline literal had | Annotate with `NextMDXOptions['options']` from `@next/mdx` |
| 2 | ❌ Blocking | `apps/docs/app/_components/TopNav.tsx:49` | Nine-item nav becomes a ~117px horizontal scroll strip at the 390px reference width | Derive the offset from the header instead of flattening the header |
| 3 | ❌ Blocking | `apps/docs/app/layout.tsx:33` | `scroll-pt-20` is a fourth uncoupled encoding of header height | Publish `--docs-header-h` on the header; derive the offset and both rails from it |
| 4 | ⚠️ Minor | `apps/docs/app/_lib/mdx-pipeline.test.ts:34` | Leading-digit id pinned as expected, with the constraint deferred to an uncited ticket | Drop the `1.` / `2.` numbering from those two H2s in `theme-setup.mdx` |
| 5 | ⚠️ Minor | `apps/docs/app/_components/TopNav.tsx:58` | `overflow-x-auto` computes `overflow-y: auto`, clipping link focus outlines | Add explicit `focus-visible` classes and vertical room on the nav |

### Architectural Concerns

**Scroll-offset contract is inverted** — `apps/docs/app/layout.tsx:33`. The header was flattened to a constant height so a hardcoded `scroll-padding-top` would hold, rather than the offset tracking the header. Publish the height as a custom property on the `header` element and have `scroll-padding-top`, `LeftRail`, and `RightRail` all read it; the nav can then keep wrapping.

**Test placement is dictated by a runner glob** — `apps/docs/app/_lib/mdx-pipeline.test.ts`. Build config gets a test inside the Next route tree reaching `../../` because `vitest.config.ts:56` only collects `apps/**/app/**`. Broaden that include to `apps/**/*.test.{ts,tsx}` and co-locate the test as `apps/docs/mdx-options.test.ts`, matching `appearance-controls.test.ts`.

---

neha-tanti (commented):

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Architecture | — | Covered in the Principal Architect review |
| Code Quality | ⚠️ | Unchecked cast and an inaccurate resolution comment in the new test |
| Testing | ❌ | The test never touches `next.config.ts`, and new content files are never pinned |

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

### Issues

| 1 | ❌ Blocking | `apps/docs/app/_lib/mdx-pipeline.test.ts:110` | Test never imports `next.config.ts`, so `createMDX({})` would still pass | Import the config under `process.env.TURBOPACK` and assert the loader options carry the plugins |
| 2 | ⚠️ Minor | `apps/docs/app/_lib/mdx-pipeline.test.ts:22` | A fourth `.mdx` gets silent zero coverage — verified by adding one | Derive the key set from a recursive `readdir` of `CONTENT_DIR` and assert every file has an entry |
| 3 | ⚠️ Minor | `apps/docs/app/_lib/mdx-pipeline.test.ts:70` | Trailing `path.sep` is a no-op; resolution starts at `apps/docs`, not `content/` | Drop it, or pass a file path inside `content` if content-relative resolution is the intent |
| 4 | ⚠️ Minor | `apps/docs/app/_lib/mdx-pipeline.test.ts:99` | `tree as HastNode` turns a non-firing capture plugin into an unhelpful `TypeError` | Guard: `if (!tree) throw new Error('capture plugin did not run')` |

### Verified, no action needed

Emptying `rehypePlugins` does fail all five tests. `it.each` `%s` renders the path correctly in Vitest 4. `@mdx-js/mdx@3.1.1` is the exact version `@mdx-js/loader` already resolves — one lockfile entry, so the test compiles with the production MDX version. `as const` on `MDX_OPTIONS` would break assignability to `@next/mdx`'s mutable `PluggableList`, so its absence is correct. `@next/mdx`'s `mdx-js-loader.js` does resolve plugin strings via `require.resolve(name, { paths: [this.context] })`, confirming the string-not-import rationale. The `#the-model` anchor added to `appearance.mdx` matches a pinned id.

---

apps/docs/mdx-options.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    headings: elements.filter((el) => /^h[1-6]$/.test(el.tagName ?? '')),
+    ids: elements
+      .map((el) => el.properties?.id)
+      .filter((id): id is string => typeof id === 'string'),
+  };
+}
+
+describe('MDX heading ids', () => {
+  it('registers rehype-slug where the loader can resolve it', () => {
+    expect(MDX_OPTIONS.rehypePlugins).toContain('rehype-slug');
```
neha-tanti, line 124: **Blocking — production wiring untested.** The test imports `MDX_OPTIONS` but never `next.config.ts`, so changing that file to `createMDX({})` leaves all five tests green. Verified a fix works: set `process.env.TURBOPACK`, `await import('../../next.config')`, then assert `turbopack.rules['#next-mdx'].loaders[0].options` carries the plugins.

apps/docs/mdx-options.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  '..',
+  'content',
+  path.sep
+);
+
+/**
+ * `footnote-label` on theme-setup comes from remark-rehype, not rehype-slug —
+ * rehype-slug skips headings that already carry an id.
+ */
+const EXPECTED_HEADING_IDS: Record<string, string[]> = {
```
neha-tanti, line 17: **Minor — silent coverage gap.** Adding a fourth `.mdx` under `content/theming/` left all five tests passing; nothing forces a new page to be pinned. Derive the key set from a recursive `readdir` of `CONTENT_DIR` and assert every `.mdx` has an entry here.

apps/docs/app/_lib/mdx-pipeline.test.ts
```diff
@@ -0,0 +1,145 @@
...
+interface HastNode {
+  type: string;
+  tagName?: string;
+  properties?: Record<string, unknown>;
+  children?: HastNode[];
+}
+
+// Mirrors @next/mdx's loader, which resolves plugin strings with
+// require.resolve from the directory of the .mdx file being compiled.
+const requireFromContent = createRequire(CONTENT_DIR);
```
neha-tanti: **Minor — comment overstates fidelity.** `createRequire` derives paths from `path.dirname(filename)`, which strips the trailing separator either way, so both forms resolve from `apps/docs/node_modules` and never `content/node_modules`. The `path.sep` argument is a no-op — drop it, or pass a file path inside `content`.

apps/docs/app/_lib/mdx-pipeline.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  await compile(source, {
+    remarkPlugins: await Promise.all(MDX_OPTIONS.remarkPlugins.map(loadPlugin)),
+    rehypePlugins: [
+      ...(await Promise.all(MDX_OPTIONS.rehypePlugins.map(loadPlugin))),
+      capture,
+    ],
+    remarkRehypeOptions: MDX_OPTIONS.remarkRehypeOptions,
+  });
+
+  const elements = collectElements(tree as HastNode);
```
neha-tanti: **Minor — cast hides undefined.** `tree as HastNode` turns a capture plugin that never fired into `Cannot read properties of undefined` inside `collectElements`. Replace with a guard: `if (!tree) throw new Error('capture plugin did not run')`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,40 +25,61 @@ const SECTIONS = [
...
       <div className="nx:flex nx:items-center nx:gap-2 nx:font-semibold nx:typography-label-default">
         Nexus DS
         <span className="nx:text-[10px] nx:uppercase nx:tracking-wider nx:text-nav-muted-foreground nx:px-1.5 nx:py-0.5 nx:border nx:border-dashed nx:border-nav-border nx:rounded-sm">
           docs
         </span>
       </div>
-      <nav className="nx:flex nx:gap-0.5 nx:flex-1 nx:flex-wrap">
+      <nav
+        ref={navRef}
+        className="nx:flex nx:gap-0.5 nx:flex-1 nx:min-w-0 nx:overflow-x-auto nx:py-1.5"
```
neha-tanti: **⚠️ Partial — mobile nav.** Centring fixes which item shows, not that nine sections share ~150px of leftover header width with no overflow affordance. `flex-wrap` previously showed all nine at 390px. Collapse the sections into a Sheet or DropdownMenu below `lg` so the header keeps its fixed 56px.

apps/docs/mdx-options.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  '..',
+  '..',
+  'content'
+);
+
+/**
+ * `footnote-label` on theme-setup comes from remark-rehype, not rehype-slug —
+ * rehype-slug skips headings that already carry an id.
+ */
+const EXPECTED_HEADING_IDS: Record<string, string[]> = {
```
neha-tanti, line 17: **⚠️ Partial — invalid-selector constraint.** Dropping the `1.` / `2.` prefixes fixes today's ids, but nothing enforces the rule, so the next `## 1. Foo` reintroduces an id that is not a valid CSS selector. Assert selector validity over every compiled id rather than relying on content review.

apps/docs/mdx-options.test.ts
```diff
@@ -0,0 +1,185 @@
...
+    const rules = config.turbopack?.rules as
+      | Record<string, { loaders: { options: Record<string, unknown> }[] }>
+      | undefined;
+
+    expect(rules?.['#next-mdx']?.loaders[0]?.options).toMatchObject(
+      MDX_OPTIONS
+    );
+  });
+
+  it('pins every .mdx under content/', async () => {
```
neha-tanti: **⚠️ New — pinning cost grows with the docs site.** Every future page must hand-add a full id list, so a heading rename fails CI in a file unrelated to the change. Keep the pins (they are what catch permalink drift) but add a property assertion over all ids so new pages get real coverage before anyone pins them.

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up)

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

Reviewed `fc12291` against the previous architecture review.

### Unresolved Issues

| 1 | Mobile nav overflow strip | `apps/docs/app/_components/TopNav.tsx:69` | ⚠️ Partial | Centring fixes which item shows, not that nine share ~150px |
| 2 | Leading-digit id invalid | `apps/docs/app/_lib/mdx-pipeline.test.ts:21` | ⚠️ Partial | Content edited; nothing stops the next `## 1. Foo` |
| 3 | Test placement forced by runner glob | `vitest.config.ts:57` | ❌ Still Present | Glob unchanged, no issue cited; one-line fix |

### Resolved

| # | Original Issue | Status |
| --- | --- | --- |
| 1 | Untyped `MDX_OPTIONS` dropped the excess-property check | ✅ Fixed — `satisfies` verified against a `tsc` probe |
| 2 | `scroll-pt-20` a fourth uncoupled header-height encoding | ✅ Fixed — one `--docs-header-h`, consumed by header, scroll port, both rails |

### Architectural Concerns

**Mobile nav regression** — `apps/docs/app/_components/TopNav.tsx:69`. Swapping `flex-wrap` for `overflow-x-auto` traded "all nine sections visible at 390px" for "one visible in a bare strip", and the fixed 56px header now rules out restoring wrap. Collapse the section list into a disclosure below `lg` so primary IA stays reachable while the header keeps a known height.

**Correction to the previous review** — my earlier suggestion to publish `--docs-header-h` on the `header` element was wrong. `html { scroll-padding-top }` cannot read a custom property declared on a descendant, since inheritance is downward-only. The author's `:root` placement is the only workable home.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,40 +25,61 @@ const SECTIONS = [
...
+      >
         {SECTIONS.map((s) => {
           const active = isActive(pathname, s.match);
           return (
             <Link
               key={s.href}
               href={s.href}
+              data-active={active}
+              aria-current={active ? 'page' : undefined}
```
neha-tanti: **⚠️ New — inaccurate `aria-current`.** On `/getting-started/install`, `/getting-started` is the section, not the page, so `page` overstates it. Use `'true'` here and put `aria-current="page"` on LeftRail's exact-match link, which currently has none.

apps/docs/app/_components/LeftRail.tsx
```diff
@@ -8,7 +8,7 @@ import type { Section } from '../_lib/sections';
 export function LeftRail({ section }: { section: Section }) {
   const pathname = usePathname();
   return (
-    <aside className="nx:sticky nx:top-16 nx:self-start nx:max-h-[calc(100svh-80px)] nx:overflow-y-auto nx:pr-2">
+    <aside className="nx:sticky nx:top-(--docs-header-h) nx:self-start nx:max-h-[calc(100svh-var(--docs-header-h))] nx:overflow-y-auto nx:pr-2">
```
neha-tanti, line 15: **⚠️ New — missing focus ring.** TopNav links gained a `focus-visible` ring in this commit; these rail links still have none, so keyboard users lose the indicator in the sidebar. Add the same `nx:focus-visible:outline-2 nx:focus-visible:outline-focus-default nx:focus-visible:outline-offset-(--focus-offset)` trio to both branches.

apps/docs/app/_lib/mdx-pipeline.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      expect(() => requireFromContent.resolve(name)).not.toThrow();
+    }
+  });
+
+  it('ships MDX_OPTIONS to the loader through next.config', async () => {
+    // @next/mdx only emits turbopack.rules when TURBOPACK is set; the webpack
+    // branch buries the same loader object inside a config callback.
+    vi.stubEnv('TURBOPACK', '1');
+    const { default: config } = await import('../../next.config');
+    vi.unstubAllEnvs();
```
neha-tanti: **⚠️ New — env leak on failure.** If the dynamic import rejects, `TURBOPACK=1` leaks to later test files in the same worker: `process.env` is process-scoped and `unstubEnvs` is unset. Set `unstubEnvs: true` on the unit project, or move this into `afterEach`.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up)

### Verdict: MINOR CHANGES

Reviewed `fc12291` against the previous code-quality review.

### Unresolved Issues

All previous SDE2 issues resolved.

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | Test never imported `next.config.ts` | ✅ Fixed | `createMDX({})` now fails the test; `rehypePlugins: []` also fails, so `toMatchObject` is array-length-strict |
| 2 | A 4th `.mdx` got silent zero coverage | ✅ Fixed | Adding `content/probe-dir/probe.mdx` fails the test; recursive discovery works on win32 |
| 3 | Trailing `path.sep` was a no-op | ✅ Fixed | Probe package proves resolution now starts at `content/`, and the file need not exist |
| 4 | `tree as HastNode` hid `undefined` | ✅ Fixed | Throwing guard; narrowing confirmed by clean typecheck |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/_components/TopNav.tsx:78` | `aria-current="page"` on a section link that is not the current page | Use `'true'` here; `'page'` belongs on LeftRail's exact-match link |
| 2 | ⚠️ Minor | `apps/docs/app/_components/LeftRail.tsx:11` | Rail links still lack the focus ring TopNav gained in this commit | Reuse the same three `focus-visible` outline utilities |
| 3 | ⚠️ Minor | `apps/docs/app/_lib/mdx-pipeline.test.ts:142` | `vi.unstubAllEnvs()` sits outside `afterEach`; a rejected import leaks `TURBOPACK=1` | `unstubEnvs: true` on the unit project, or move to `afterEach` |

### Verified, no action needed

The centring math reduces to `contentPos - (navW - linkW)/2`, and the `scrollLeft` setter clamps, so the first link (negative) and last link (past `scrollWidth`) are safe and a non-overflowing nav is a no-op. Assigning `Element.scrollLeft` scrolls only that element's box, so a cold load at `#hash` is not perturbed — the PR's core feature holds. React stringifies booleans on `data-*`, so `[data-active="true"]` matches only the active link. `--docs-scroll-offset` resolves to `5rem`, matching the previous `scroll-pt-20`. Node floor is 20.19.0, above the `Dirent.parentPath` floor of 20.12.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
+                {s.label}
+              </Link>
+            </DropdownMenuItem>
+          ))}
+        </DropdownMenuContent>
+      </DropdownMenu>
+      <nav
+        ref={navRef}
+        aria-label="Sections"
+        className="nx:hidden nx:lg:flex nx:gap-0.5 nx:flex-1 nx:min-w-0 nx:overflow-x-auto nx:py-1.5"
```
neha-tanti: **⚠️ Partial — breakpoint one tier low.** Measured in Chromium at the real token values: at 1024px `For AI agents` and `Changelog` are clipped; at 1152px `Changelog` is. The strip only fits at 1280px. Swap the strip and the trigger at `xl`, not `lg`.

apps/docs/app/_components/nexus.tsx
```diff
@@ -21,6 +21,10 @@ export {
   CommandInput,
   CommandItem,
   CommandList,
+  DropdownMenu,
```
neha-tanti: **⚠️ New — barrel leaks into MDX.** `mdx-components.tsx:163` spreads `...Nexus`, so every barrel export becomes an MDX tag and lands in each MDX route's graph. These four are TopNav-only. Import them from `@nexus_ds/react` in TopNav, which is already `'use client'`.

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 3)

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

Reviewed `3798a03`.

### Unresolved Issues

| 1 | Nine-link strip overflows with no affordance | `TopNav.tsx:101` | ⚠️ Partial | 1024–1280 band still clips two links; measured in Chromium |
| 2 | Per-page pinned id lists burden | `mdx-options.test.ts:169` | ⚠️ Partial | Property loop gated by the hand-maintained map |

### Resolved

| # | Original Issue | Status |
| --- | --- | --- |
| 1 | Test placement forced by the runner glob | ✅ Fixed — `apps/*/*.test.{ts,tsx}` added; moved file is collected |
| 2 | Leading-digit ids unenforced | ✅ Fixed — jsdom `querySelector` guard verified to throw on `1-foo` and on an empty id |

### New Issues

| 1 | ⚠️ Minor | `nexus.tsx:24` | Barrel is spread into MDX globals; nav-only parts leak into every MDX route graph | Import `DropdownMenu*` from `@nexus_ds/react` in TopNav |
| 2 | ⚠️ Minor | `TopNav.tsx:86` | Local `SECTIONS` duplicates `_lib/sections.ts`, the declared IA source of truth | Derive from `Object.values(SECTIONS)` |
| 3 | ⚠️ Minor | `TopNav.tsx:122` | Empty spacer div holds header layout below `lg` | A margin utility on an existing element |

### Architectural Concerns

**Breakpoint is one tier too low** — `TopNav.tsx:101`. Measured at the real token values, the strip needs 829–897px of content width against 641px available at 1024px and 769px at 1152px, so the swap belongs at `xl`, not `lg`.

**Two renderings of one nav, one source** — `TopNav.tsx:86`. The strip and the menu now both walk a locally-declared `SECTIONS` that shadows `_lib/sections.ts`, which already backs LeftRail and SearchPalette. Derive the header's list from that registry so the IA has one definition.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
+            variant="ghost"
+            size="sm"
+            className="nx:lg:hidden"
+            aria-label="Browse sections"
+          >
+            ☰ {currentSection?.label ?? 'Sections'}
+          </Button>
+        </DropdownMenuTrigger>
+        <DropdownMenuContent align="start">
+          {SECTIONS.map((s) => (
```
neha-tanti: **⚠️ New — duplicated IA.** This array is a second copy of `_lib/sections.ts`, which documents itself as the single source of truth and already backs SearchPalette and LeftRail. Derive it from `Object.values(SECTIONS)` plus the two standalone routes.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
+                  ? `${NAV_LINK_BASE} nx:bg-nav-item-active nx:text-primary-subtle-foreground`
+                  : `${NAV_LINK_BASE} nx:text-nav-muted-foreground nx:hover:text-nav-foreground nx:hover:bg-nav-item-hover`
               }
             >
               {s.label}
             </Link>
           );
         })}
       </nav>
+      <div className="nx:flex-1 nx:lg:hidden" />
```
neha-tanti: **⚠️ New — spacer div.** An empty breakpoint-conditional div to absorb slack is structural. Delete it and put `nx:ms-auto` on the theme-toggle Button — at `lg`+ the nav's `flex-1` already eats the free space, so it is a no-op there.

apps/docs/mdx-options.test.ts
```diff
@@ -0,0 +1,189 @@
...
+      const source = await readFile(
+        path.join(CONTENT_DIR, contentPath),
+        'utf8'
+      );
+      const { headings, ids } = await compileMdx(source);
+
+      expect(headings.map((h) => h.properties?.id)).toEqual(expected);
+      expect(new Set(ids).size).toBe(ids.length);
+
+      for (const id of ids) {
```
neha-tanti, line 180: **⚠️ Partial — guard still gated by the pins.** The property loop runs inside the hand-pinned `it.each`, so a new page is unguarded until someone types its full id list. Move it to its own `it.each` over `contentFiles()`; keep the pins as the deep-link net.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 3)

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

Reviewed `3798a03`.

### Unresolved Issues

All previous SDE2 issues resolved.

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | `aria-current="page"` on a section link | ✅ Fixed | `'true'` on TopNav, `'page'` on LeftRail's exact-match link |
| 2 | Rail links lacked the focus ring | ✅ Fixed | Class-by-class diff of the `RAIL_LINK_BASE` restructure — nothing dropped |
| 3 | `vi.unstubAllEnvs()` could leak `TURBOPACK` | ✅ Fixed | Stubbed in one test, read `undefined` in the next in the same file |

### New Issues

| 1 | ❌ Blocking | `TopNav.tsx:80` | `aria-label` overrides the visible `☰ Getting Started` — WCAG 2.5.3, Level A | Drop the `aria-label`; wrap the glyph in `aria-hidden` |
| 2 | ⚠️ Minor | `TopNav.tsx:74` | Nav landmark is `nx:hidden` below `lg` and the menu sits outside it | Wrap the menu in its own `nav` labelled `Sections` |
| 3 | ⚠️ Minor | `TopNav.tsx:79` | Trigger is 32px; text sizes get no coarse-pointer overlay | Add the `pointer-coarse` inset overlay to clear ~44px |
| 4 | ⚠️ Minor | `TopNav.tsx:90` | Menu items convey active state only via `aria-current` | Mirror the strip's `nx:bg-nav-item-active` |

### Verified, no action needed

`DropdownMenuItem asChild` + `Link` navigates *and* closes: Radix's Slot only merges handlers present on the child, so its handler is passed to `Link` as `onClick`; `Link` calls it, checks `defaultPrevented`, then navigates, and Radix's `handleSelect` does not prevent default. Keyboard path confirmed in jsdom — ArrowDown focuses the first item, typeahead `"the"` jumps to Theming, Enter navigates and closes. The centring `useEffect` is a harmless no-op below `lg` (zeroed rects yield `+= 0`, not `NaN`). `currentSection?.label ?? 'Sections'` handles routes outside `SECTIONS`. One latent trap: an `onClick` calling `preventDefault` on that `Link` would silently skip `handleSelect` and leave the menu open.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
         </span>
       </div>
-      <nav className="nx:flex nx:gap-0.5 nx:flex-1 nx:flex-wrap">
+      <DropdownMenu>
+        <DropdownMenuTrigger asChild>
+          <Button
+            variant="ghost"
+            size="sm"
+            className="nx:lg:hidden"
+            aria-label="Browse sections"
```
neha-tanti: **❌ Blocking — label in name.** `aria-label="Browse sections"` overrides the visible `☰ Getting Started`, so the accessible name omits the visible text — WCAG 2.5.3, Level A. Drop the aria-label and hide the glyph with `<span aria-hidden="true">☰</span>`. Radix already supplies `aria-haspopup` and `aria-expanded`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
-    <header className="nx:sticky nx:top-0 nx:z-30 nx:flex nx:items-center nx:gap-6 nx:px-6 nx:py-3 nx:bg-nav-background nx:text-nav-foreground nx:border-b nx:border-nav-border">
+    <header className="nx:sticky nx:top-0 nx:z-30 nx:flex nx:items-center nx:gap-6 nx:px-6 nx:h-(--docs-header-h) nx:bg-nav-background nx:text-nav-foreground nx:border-b nx:border-nav-border">
       <div className="nx:flex nx:items-center nx:gap-2 nx:font-semibold nx:typography-label-default">
         Nexus DS
         <span className="nx:text-[10px] nx:uppercase nx:tracking-wider nx:text-nav-muted-foreground nx:px-1.5 nx:py-0.5 nx:border nx:border-dashed nx:border-nav-border nx:rounded-sm">
           docs
         </span>
       </div>
-      <nav className="nx:flex nx:gap-0.5 nx:flex-1 nx:flex-wrap">
+      <DropdownMenu>
```
neha-tanti: **⚠️ New — landmark lost on mobile.** The `nav aria-label="Sections"` element is `nx:hidden` below `lg` and this menu sits outside it, so mobile loses the navigation landmark entirely. Wrap the DropdownMenu in its own `nav aria-label="Sections"` with `nx:lg:hidden`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
+        <DropdownMenuTrigger asChild>
+          <Button
+            variant="ghost"
+            size="sm"
+            className="nx:lg:hidden"
```
neha-tanti: **⚠️ New — touch target.** `size="sm"` is `nx:h-8` (32px), and only `icon-sm`/`icon` carry a coarse-pointer overlay — text sizes do not. This is the sole mobile nav control, so add `nx:relative nx:pointer-coarse:after:absolute nx:pointer-coarse:after:-inset-2` to clear the ~44px floor in `responsive.md`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -23,47 +31,95 @@ const SECTIONS = [
...
+          >
+            ☰ {currentSection?.label ?? 'Sections'}
+          </Button>
+        </DropdownMenuTrigger>
+        <DropdownMenuContent align="start">
+          {SECTIONS.map((s) => (
+            <DropdownMenuItem key={s.href} asChild>
+              <Link
+                href={s.href}
+                aria-current={isActive(pathname, s.match) ? 'true' : undefined}
```
neha-tanti: **⚠️ New — no visible active row.** The current section is marked only with `aria-current`; the item classes are byte-identical, so sighted users see no active row once the menu opens. Add a conditional class mirroring the strip's `nx:bg-nav-item-active`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,121 @@
...
+    <header className="nx:sticky nx:top-0 nx:z-30 nx:flex nx:items-center nx:gap-6 nx:px-6 nx:h-(--docs-header-h) nx:bg-nav-background nx:text-nav-foreground nx:border-b nx:border-nav-border">
       <div className="nx:flex nx:items-center nx:gap-2 nx:font-semibold nx:typography-label-default">
         Nexus DS
         <span className="nx:text-[10px] nx:uppercase nx:tracking-wider nx:text-nav-muted-foreground nx:px-1.5 nx:py-0.5 nx:border nx:border-dashed nx:border-nav-border nx:rounded-sm">
           docs
         </span>
       </div>
-      <nav className="nx:flex nx:gap-0.5 nx:flex-1 nx:flex-wrap">
-        {SECTIONS.map((s) => {
+      <nav aria-label="Sections" className="nx:flex-1 nx:xl:hidden">
```
neha-tanti: **⚠️ New — empty landmark.** This `nav` contains zero links: `DropdownMenuContent` portals to `document.body`, so below `xl` the section links sit outside every landmark and carry `role="menuitem"`, not link. Render a disclosure button plus a `ul` of links inside this `nav` instead.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,121 @@
 'use client';
 
+import { useEffect, useRef } from 'react';
+
+import {
```
neha-tanti: **⚠️ New — two import paths.** `DropdownMenu*` comes from `@nexus_ds/react` here while `Button` comes from the `./nexus` barrel, in the same client file. Pick one path; the barrel's docstring only justifies the server-component case.

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 4)

### Verdict: NEEDS DISCUSSION

Reviewed `2ca89c6` against a real `next build`, a live server, and Playwright measurements at 360/390/430/768/1024/1279/1280/1440.

### Unresolved Issues

All previous architectural issues resolved.

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | Strip clipped through the `lg` band | ✅ Fixed | At 1280 the strip measures `width 915 / scrollWidth 915` with all 9 links; `display:none` below |
| 2 | Property loop gated by the pin map | ✅ Fixed | Now its own `it.each` over `contentFiles()` |
| 3 | Barrel leaked nav parts into MDX globals | ✅ Fixed | `DropdownMenu*` no longer exported from `nexus.tsx` |
| 4 | Local `SECTIONS` duplicated the IA registry | ✅ Fixed | Derived; `Object.values` ordering is spec-guaranteed for string keys and already the precedent in `SearchPalette` |
| 5 | Spacer div | ✅ Fixed | Deleted; no horizontal document overflow at any tested width |

### New Issues

| 1 | ⚠️ Minor | `TopNav.tsx:70` | The new nav landmark holds zero links — menu content portals to `body` | Disclosure button + `<ul>` of links inside the `nav` |
| 2 | ⚠️ Minor | `TopNav.tsx:5` | Two import paths for `@nexus_ds/react` in one client file | Pick one path in this file |

### Architectural Concerns

**The mobile landmark is empty** — `TopNav.tsx:70`. Measured with the menu open: `menu.closest('nav')` is `false`, the content's parent is `BODY`, and the visible mobile `nav` reports `links: 0`. Wrapping the trigger does not bring the portalled items into the landmark.

**Barrel answers two questions** — `app/_components/nexus.tsx:1`. Its docstring scopes it to the RSC client boundary, but `mdx-components.tsx:163` spreads it into MDX globals, so membership is now decided by MDX-tag policy rather than the boundary it documents. Replace the `...Nexus` spread with an explicit tag map.

**Home and Changelog stay hardcoded, correctly** — folding them into `SECTIONS` would corrupt `Object.keys(SECTIONS)` in `app/[section]/page.tsx`'s `generateStaticParams`.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,121 @@
...
+          <DropdownMenuContent align="start">
+            {NAV_LINKS.map((s) => {
+              const active = isActive(pathname, s.match);
+              return (
+                <DropdownMenuItem key={s.href} asChild>
+                  <Link
+                    href={s.href}
+                    aria-current={active ? 'true' : undefined}
+                    className={
+                      active ? 'nx:text-primary-subtle-foreground' : undefined
```
neha-tanti: **⚠️ Partial — invisible in dark.** At the docs' default brand, dark `primary-subtle-foreground` is `oklch(1 0 0)` against `popover-foreground` `oklch(0.9848)` — ΔL 0.015. `nx:focus:text-popover-foreground` also erases it on hover. Add a non-colour cue such as `nx:font-semibold`.

apps/docs/mdx-options.test.ts
```diff
@@ -0,0 +1,204 @@
...
+  const elements = collectElements(tree);
+  return {
+    headings: elements.filter((el) => /^h[1-6]$/.test(el.tagName ?? '')),
+    ids: elements
+      .map((el) => el.properties?.id)
+      .filter((id): id is string => typeof id === 'string'),
+  };
+}
+
+const CONTENT_FILES = await contentFiles();
```
neha-tanti, line 120: **⚠️ New — directory read twice.** `pins every .mdx under content/` still calls `contentFiles()` again at line 153, so the same directory is read from two independent lists. Use `CONTENT_FILES` there.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,121 @@
...
   if (match === '/') return pathname === '/';
   return pathname === match || pathname.startsWith(match + '/');
 }
 
 export function TopNav() {
   const pathname = usePathname();
+  const navRef = useRef<HTMLElement>(null);
   const { resolvedMode, setState } = useNexusAppearance();
   const isDark = resolvedMode === 'dark';
+  const currentSection = NAV_LINKS.find((s) => isActive(pathname, s.match));
```
neha-tanti: **⚠️ New — misleading name.** `currentSection` resolves to a `NAV_LINKS` entry, which may be Home or Changelog — neither is a section — and `SECTIONS` now means the imported registry in this file. Rename to `currentLink`.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 4)

### Verdict: MINOR CHANGES

Reviewed `2ca89c6`. Every claim below was checked by running code, not by inspection.

### Unresolved Issues

| 4 | Active menu item conveyed only via `aria-current` | `TopNav.tsx:91` | ⚠️ Partial | Chosen token is invisible in dark; still colour-only |

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | `aria-label` overrode the visible trigger text | ✅ Fixed | Prerendered accessible name is now `Getting Started`; glyph is `aria-hidden` |
| 2 | Nav landmark lost below the breakpoint | ✅ Fixed | Both `nav aria-label="Sections"` are mutually exclusive, so only one is ever in the a11y tree |
| 3 | 32px trigger with no coarse-pointer overlay | ✅ Fixed | Emitted CSS confirms `content:var(--tw-content)` with `--tw-content:""` registered; 48px hit area, unclipped |

### New Issues

| 1 | ⚠️ Minor | `TopNav.tsx:127` | Sibling mode toggle left at 32px with no coarse overlay | Apply the same `after:-inset-2` pattern |
| 2 | ⚠️ Minor | `mdx-options.test.ts:123` | `contentFiles()` is read twice from two lists | Use `CONTENT_FILES` at line 153 |
| 3 | ⚠️ Minor | `TopNav.tsx:43` | `currentSection` names a `NAV_LINKS` entry, not a `SECTIONS` entry | Rename to `currentLink` |

_Issue 1 has no inline comment: it lands on a line this PR did not touch._

### Verified, no action needed

`NAV_LINKS` infers as `{ href: string; label: string; match: string }[]`; `tsc --noEmit` clean. Top-level `await` is reliable here — 10 tests collect under `--project unit`, a module-scope `readdir` failure surfaces as a named `Failed Suite` with a line-accurate stack, and Vitest's `experimental.preParse` caveat is off by default. Adding `content/theming/zz-probe.mdx` with `## 1. Foo` failed both the new `querySelector` guard and `pins every .mdx`; a heading-free file failed the `>0` guard, so that assertion is a real vacuity guard; renaming a pinned heading still fails only the pinned test.

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 5)

### Verdict: NEEDS DISCUSSION

Reviewed `4c95170` against a real `npx next build` and the shipped Nexus components.

### Unresolved Issues

| 1 | Barrel spread into MDX globals | `mdx-components.tsx:163` | 🔄 Changed | Both files left the diff; now unchanged `main` code, out of scope |

### Resolved

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | Nav landmark held zero links | ✅ Fixed | Links now render inside the `nav`; no portal involved |
| 2 | Two import paths in one client file | ✅ Fixed | `DropdownMenu*` import removed; `nexus.tsx` is back to its `main` state |

### New Issues

| 1 | ⚠️ Minor | `TopNav.tsx:102` | Floating flyout painted with the `nav-*` family | Move the surface to the popover family |
| 2 | ⚠️ Minor | `TopNav.tsx:101` | Registry-derived list has no height bound | `max-h` + `overflow-y-auto` on the panel |

### Architectural Concerns

**Token ownership ruling** — `TopNav.tsx:102`. The panel is transient floating content, which `components.md` assigns to the popover family; `nav-*` is scoped to persistent chrome. The shipped `NavigationMenuViewport` uses `popoverSurfaceClassName`, and the docs site is the surface consumers copy from.

### Ruled on, no change requested

**Hand-rolling the disclosure is correct here.** None of the three shipped candidates fits: `PopoverContent` hardcodes `PopoverPrimitive.Portal` (`popover.tsx:72`) — the exact portal-out-of-landmark bug from round 4; `NavigationMenu` is hover-opened with no click-only trigger mode; `Collapsible` is in-flow with no dismissal layer. The real gap is `PopoverContent`'s non-optional portal, which does not belong in a heading-ids PR.

**`nx:z-popover` inside the `nx:z-30` header stacks correctly.** The sticky header forms a stacking context, so the panel's 70 orders only against header siblings while the header sits at 30 above `auto` page content — locally inert but matching the DS's own `NavigationMenuViewport` convention.

**Two `nav aria-label="Sections"` landmarks are sound.** The breakpoint classes are mutually exclusive, so exactly one is in the a11y tree, and the two presentations share `NAV_LINKS`, `NAV_LINK_BASE`, and `isActive` — the per-mode split `composition-over-render-props.md` prescribes.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,144 @@
...
+          size="sm"
+          aria-expanded={menuOpen}
+          onClick={() => setMenuOpen((open) => !open)}
+          className="nx:relative nx:pointer-coarse:after:absolute nx:pointer-coarse:after:-inset-2"
+        >
+          <span aria-hidden="true">☰</span>
+          {currentLink?.label ?? 'Sections'}
+        </Button>
+        {menuOpen && (
+          <ul className="nx:absolute nx:top-full nx:left-0 nx:mt-1 nx:z-popover nx:min-w-48 nx:list-none nx:m-0 nx:flex nx:flex-col nx:gap-0.5 nx:p-1 nx:rounded-md nx:border nx:border-nav-border nx:bg-nav-background nx:shadow-lg">
```
neha-tanti: **⚠️ New — wrong token family.** This is transient floating content, not persistent chrome. `components.md` assigns floating content and menu rows to the popover family, and the shipped `NavigationMenuViewport` uses `popoverSurfaceClassName`. Move the surface to popover and keep a visibly distinct active row.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,144 @@
...
+          variant="ghost"
+          size="sm"
+          aria-expanded={menuOpen}
+          onClick={() => setMenuOpen((open) => !open)}
+          className="nx:relative nx:pointer-coarse:after:absolute nx:pointer-coarse:after:-inset-2"
+        >
+          <span aria-hidden="true">☰</span>
+          {currentLink?.label ?? 'Sections'}
+        </Button>
+        {menuOpen && (
```
neha-tanti, line 115: **⚠️ New — unbounded panel.** `NAV_LINKS` grows with the `SECTIONS` registry, but the panel has no height bound. Pinned inside a sticky header, overflow rows land off-screen and are unreachable — page scroll will not reveal them. Add `max-h` plus `overflow-y-auto`.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,144 @@
...
+  useEffect(() => {
+    if (!menuOpen) return;
+
+    const closeOnOutsidePointer = (event: PointerEvent) => {
+      if (menuRef.current?.contains(event.target as Node)) return;
+      setMenuOpen(false);
+    };
+
+    const closeOnEscape = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') setMenuOpen(false);
```
neha-tanti: **❌ Blocking — focus lost on Escape.** Verified at 390px: Escape with focus on a panel link leaves `document.activeElement === document.body`; Radix previously returned focus to the trigger. Ref the trigger and `.focus()` it before closing. Nothing closes on focus-out either, so Tab past the last link leaves the panel open over the page.

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,62 +1,144 @@
...
 }
 
 export function TopNav() {
   const pathname = usePathname();
+  const navRef = useRef<HTMLElement>(null);
+  const menuRef = useRef<HTMLElement>(null);
+  const [menuOpen, setMenuOpen] = useState(false);
   const { resolvedMode, setState } = useNexusAppearance();
   const isDark = resolvedMode === 'dark';
+  const currentLink = NAV_LINKS.find((s) => isActive(pathname, s.match));
```
neha-tanti: **⚠️ New — leftover naming.** `currentSection` became `currentLink`, but this callback and both `NAV_LINKS.map` callbacks still bind `s` for section. Rename to `link`.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 5)

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

Reviewed `4c95170`. Every result below came from driving the built app in Playwright at 390px.

### Unresolved Issues

All previous SDE2 issues resolved.

| # | Original Issue | Status |
| --- | --- | --- |
| 1 | Active row invisible in dark | ✅ Fixed — `font-semibold` is a non-colour cue |
| 2 | Mode toggle touch target | ✅ Fixed |
| 3 | `contentFiles()` read twice | ✅ Fixed — injected `.mdx` still fails the pin test correctly |
| 4 | `currentSection` misnamed | ✅ Fixed |

### New Issues

| 1 | ❌ Blocking | `TopNav.tsx:50` | Escape from a panel link drops focus to `body` | Ref the trigger and `.focus()` it before closing |
| 2 | ⚠️ Minor | `TopNav.tsx:39` | Callback params still named `s` after the `currentLink` rename | Rename to `link` |

### Verified, no action needed

| Check | Result |
| --- | --- |
| Escape with focus on the trigger | Stays on `BUTTON` |
| Outside pointerdown, and drag starting inside | Closes / stays open |
| Listener leak | None — the guard plus `[menuOpen]` dep register and clean up correctly |
| Hydration | Zero console errors; `ul` absent from SSR HTML and initial render |
| Back-nav with the panel open | Panel closed, label correct |

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 6)

### Verdict: APPROVED

Reviewed `5903f81`. Ran `pnpm --filter @nexus_ds/docs typecheck`, `pnpm lint` (`--max-warnings 0`), and `pnpm --filter @nexus_ds/core audit:class-refs` — all clean.

### Unresolved Issues

All previous architectural issues resolved.

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | Flyout painted with the `nav-*` family | ✅ Fixed | Panel now `nx:bg-popover nx:text-popover-foreground nx:border-border-default`, rows `nx:hover:bg-popover-hover` (`TopNav.tsx:113–125`). `audit:class-refs` scans `apps/` and passes. |
| 2 | Registry-derived list had no height bound | ✅ Fixed | `nx:max-h-[70svh] nx:overflow-y-auto` on the panel. `svh` keeps the cap on the chrome-shown viewport, so the panel cannot outgrow the screen as `SECTIONS` grows. |

### New Issues

None.

### Ruled on, no change requested

**Solid `bg-popover` instead of the translucent recipe.** `popoverSurfaceClassName` is not exported from `@nexus_ds/react`, and its `reduce-transparency` fallback is a `@custom-variant` declared in `packages/react/src/index.css`, which the docs app deliberately does not import. Copying the translucent classes without that variant would ship blur with no `prefers-reduced-transparency` fallback, so the solid fill is the correct choice here.

**`max-h-[70svh]` arbitrary value.** `responsive.md` bans raw `vh` and endorses `svh` for content that must not clip; `clamp()`/`svh` are not tokenised yet, so the bracket value is the sanctioned form.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,65 +1,158 @@
...
+  const menuRef = useRef<HTMLElement>(null);
+  const triggerRef = useRef<HTMLButtonElement>(null);
+  const [menuOpen, setMenuOpen] = useState(false);
   const { resolvedMode, setState } = useNexusAppearance();
   const isDark = resolvedMode === 'dark';
+  const currentLink = NAV_LINKS.find((link) => isActive(pathname, link.match));
+
+  // Tabbing off the last row leaves the panel open over the page otherwise.
+  const closeOnFocusLeave = (event: React.FocusEvent<HTMLElement>) => {
+    if (event.currentTarget.contains(event.relatedTarget)) return;
```
neha-tanti, line 59: **❌ Blocking — null `relatedTarget`.** `contains(null)` is false, so a focusout with no next target closes the panel. Verified in Chromium: `blur()` on alt-tab unmounts it. On Safari/Firefox-macOS anchors take no focus on mousedown, so a row can unmount before its click lands. Add `if (!event.relatedTarget) return;`.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 6)

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

Reviewed `5903f81` against a real `next build`, a live server, and Playwright Chromium.

### Unresolved Issues

All previous SDE2 issues resolved.

| # | Original Issue | Status | Evidence |
| --- | --- | --- | --- |
| 1 | Escape from a panel link dropped focus to `body` | ✅ Fixed | Playwright: `activeElement` is the trigger button, panel unmounted |
| 2 | Callback params still named `s` | ✅ Fixed | All three map/find callbacks renamed |

### New Issues

| 1 | ❌ Blocking | `TopNav.tsx:44` | Null `relatedTarget` closes the panel, and can swallow a row click | `if (!event.relatedTarget) return;` before the `contains` guard |

### Verified, no action needed

| Behaviour | Evidence |
| --- | --- |
| `focus()` before `setMenuOpen(false)` ordering | React flushes after the native handler; trigger still mounted when focused |
| Tab off the last row closes the panel | 9 rows tabbed, panel count 0, focus lands on the theme toggle |
| Escape vs `closeOnFocusLeave` double-close | Focus stays inside `nav`, guard returns early; no second `setMenuOpen` |
| Pointer click on a row navigates (Chromium) | `focusin` on the anchor precedes `mousedown`; URL → `/changelog` |
| Types and lint | `pnpm --filter @nexus_ds/docs typecheck` and `pnpm lint` both clean |

WebKit and Firefox Playwright browsers are not installed here, so the Safari path was modelled in Chromium rather than confirmed on real WebKit.

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 7)

### Verdict: APPROVED

Reviewed `c0edd92`. Ran `pnpm --filter @nexus_ds/docs typecheck` and `pnpm lint` — both clean.

### Unresolved Issues

All previous architectural issues resolved.

### New Issues

None.

### Ruled on, no change requested

**Narrowing focus-leave dismissal does not break the disclosure contract.** Escape, outside `pointerdown`, the trigger toggle, a link click, and focus moving to any real element outside the nav all still close the panel; only document/window blur is excluded. On return, focus is still inside the panel, so the next focus move dismisses — and a non-modal disclosure has no APG obligation to close on window blur.

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,65 +1,161 @@
...
+  const menuRef = useRef<HTMLElement>(null);
+  const triggerRef = useRef<HTMLButtonElement>(null);
+  const [menuOpen, setMenuOpen] = useState(false);
   const { resolvedMode, setState } = useNexusAppearance();
   const isDark = resolvedMode === 'dark';
+  const currentLink = NAV_LINKS.find((link) => isActive(pathname, link.match));
+
+  // Tabbing off the last row leaves the panel open over the page otherwise.
+  // A null relatedTarget is focus leaving the document rather than moving on,
+  // which closeOnOutsidePointer already covers.
```
neha-tanti: **⚠️ New — comment states rationale, not logic.** Null `relatedTarget` also fires when focus drops to `body` inside the document, and the row-click case is closed by the `Link` `onClick`, not `closeOnOutsidePointer`. Drop both lines, or keep one line stating only the DOM fact.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 7)

### Verdict: MINOR CHANGES

Reviewed `c0edd92`. Ran `pnpm --filter @nexus_ds/docs typecheck` (clean), `pnpm lint` (clean), `next build` + `next start`, an 11-assertion Playwright/Chromium drive at 390px, and a mutation rerun with the guard deleted.

### Unresolved Issues

All previous SDE2 issues resolved.

| # | Original Issue (round 6) | File | Status | Notes |
| 1 | `contains(null)` closed the panel on any focusout with no next target (❌ Blocking) | `TopNav.tsx:45–49` | ✅ Fixed | Guard verified live; mutation rerun fails without |

### New Issues

| 1 | ⚠️ Minor | `TopNav.tsx:43–44` | Comment is rationale, and mis-states when `relatedTarget` is null | Drop both lines, or keep one accurate line: null also means focus dropped to `body` |

### Verified, no action needed

| Behaviour | Evidence |
| --- | --- |
| Null `relatedTarget` keeps panel mounted | Native `blur()` on a row and synthetic null-`relatedTarget` `focusout` both keep it open |
| Guard is load-bearing | Guard deleted + rebuilt: exactly those two assertions fail, other nine still pass |
| Tab out to a real element closes | Focus lands on the button outside `nav`; panel unmounts |
| Outside pointerdown closes | Click at (200,700) after a null-`focusout` dismisses |
| Escape closes and returns focus to trigger | `document.activeElement === trigger`, panel gone |
| Row click navigates | Reaches `/changelog`, panel unmounted; `pointerdown` inside `menuRef` does not pre-close |
| No stuck-open state | After each null-`relatedTarget` case, both Escape and outside pointerdown still dismiss |
| Blur while panel closed | No-op; `setMenuOpen(false)` on an already-false state |
| Type narrowing without cast | Typecheck clean; `relatedTarget` narrows from nullable `Element` to a `Node` for `contains` |
| Two-guard shape | Sequential early returns match `guard-clauses.md`; combining them would lose the narrowing |
| Safari/Firefox-macOS anchor path | Not executed — only Chromium binaries installed; headless cross-tab blur is non-discriminating |

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 8)

### Verdict: APPROVED

Reviewed `1bc9144`. Ran `git diff c0edd92..1bc9144` (comment-only, one file), `pnpm --filter @nexus_ds/docs typecheck` (clean), and `pnpm lint` (clean).

### Unresolved Issues

All previous architectural issues resolved.

### New Issues

None.

---

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 8)

### Verdict: APPROVED

> Posted as a comment, not a formal `APPROVE`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects approving a self-authored PR (verified — HTTP 422 "Can not approve your own pull request").

Reviewed `1bc9144`. Ran `pnpm --filter @nexus_ds/docs typecheck` (pass) and `pnpm lint` (pass, 0 warnings); verified the `relatedTarget` claim against MDN `FocusEvent.relatedTarget` — no rebuild or Playwright run, since the commit changes only a comment.

### Unresolved Issues

All previous SDE2 issues resolved.

| # | Original Issue (round 7) | File | Status | Notes |
| 1 | Comment stated rationale and mis-stated the null case (⚠️ Minor) | `TopNav.tsx:42–43` | ✅ Fixed | States DOM behaviour, matches MDN, no duplication |

The replacement line states DOM behaviour — `relatedTarget` is null when no element receives focus — rather than rationale, and is distinct from the retained first line, which names the behaviour the handler exists to prevent. The contrast with "leaves the nav" is correct: that case is non-null and handled by the `contains` guard on the next line.

### New Issues

None.

`nexus-docs-mcp` was not exposed in this session, so the DOM-API check fell back to MDN.

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 9)

### Verdict: APPROVED

> Posted as a comment, not a formal `APPROVE`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects approving a self-authored PR.

Reviewed `1bc9144..bc64af6`.

### Unresolved Issues

### New Issues

None.

### Verified

| Check | Result |
| --- | --- |
| `nexus.tsx` barrel is components-only; `mdx-components.tsx:163` spread receives no utilities | ✅ |
| `data-active` removal left no orphaned consumer (`git grep` → only unrelated `LiveThemeSwapper.tsx:85`) | ✅ |
| `--docs-header-h` / `--docs-scroll-offset` still single-sourced; comment removal is text-only | ✅ |
| Import split is a rule, not an accident — components from `./nexus`, the `cn` utility from `@nexus_ds/react` | ✅ |
| `scroll-initial-target` is Chrome/Edge 133+ only, outside the Nexus browser floor — the JS path is correct | ✅ |
| `pnpm --filter @nexus_ds/docs typecheck`, `pnpm lint` | clean |

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,65 +1,157 @@
...
+const COARSE_HIT_AREA =
+  'nx:relative nx:pointer-coarse:after:absolute nx:pointer-coarse:after:-inset-2';
+
 function isActive(pathname: string, match: string) {
   if (match === '/') return pathname === '/';
   return pathname === match || pathname.startsWith(match + '/');
 }
 
+function centerInStrip(link: HTMLAnchorElement | null) {
+  link?.scrollIntoView({ block: 'nearest', inline: 'center' });
```
neha-tanti: **❌ Blocking — vertical scroll jump.** `html` sets `scroll-padding-top: 70px`, so `block: 'nearest'` treats this sticky-header link as out of view and scrolls the document up 60px. Measured on `#next-steps`: scrollY 1481 → 1421. Set the strip's `scrollLeft` from rect math instead.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 9)

### Verdict: CHANGES REQUESTED

> Posted as a comment, not a formal `REQUEST_CHANGES`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects requesting changes on a self-authored PR.

Reviewed `1bc9144..bc64af6`. `eeceed1` had not been code-quality reviewed before this round.

### Unresolved Issues

All previous SDE2 issues resolved.

### New Issues

| 1 | ❌ Blocking | `apps/docs/app/_components/TopNav.tsx:39` | `scrollIntoView` drags the document up 60px on `#hash` landing | Set the strip's own `scrollLeft` from the rect delta; leave the block axis alone |

### Verified, no action needed

`cn()` emits a class-for-class identical set to the previous interpolation across all six branches, compared against the built `cn` with the `nx:` prefix and custom border groups configured. The module-scope ref fires once on mount and once per active-link change, and not on unrelated re-renders — a theme toggle produces zero calls. `inline: 'center'` does centre correctly once the strip overflows, but at 1280/1320/1500px `scrollWidth` equals `clientWidth`, so the horizontal effect is currently a no-op and only the vertical side effect is observable. No selector still reads the removed `data-active`. `scroll-behavior` stays `auto`, so nothing animates under reduced motion. `pnpm lint`, `pnpm --filter @nexus_ds/docs typecheck`, and all 10 `mdx-options.test.ts` tests pass.

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up, round 10)

### Verdict: APPROVED

> Posted as a comment, not a formal `APPROVE`: the `nexuslabs-ai-bot` account is not configured in this `gh` install, and GitHub rejects approving a self-authored PR.

Reviewed `bc64af6..1cc8de5`.

### Unresolved Issues

### New Issues

None.

### Verified

| Check | Result |
| --- | --- |
| `centerInStrip`'s `parentElement` read stays module-private — helper, `ref` call site, and the `overflow-x-auto` strip all live in `TopNav.tsx` | ✅ |
| Module-scope helper preserves ref-callback identity; closing over a `stripRef` would re-centre on every unrelated re-render | ✅ |
| Mutation is scoped to the strip's own scroller, not the document scroller — no new coupling to `--docs-scroll-offset` | ✅ |
| No scalability, security, API-contract, data-model, or stub-route surface touched | ✅ |

---

apps/docs/app/_components/TopNav.tsx
```diff
@@ -1,65 +1,162 @@
...
+const COARSE_HIT_AREA =
+  'nx:relative nx:pointer-coarse:after:absolute nx:pointer-coarse:after:-inset-2';
+
 function isActive(pathname: string, match: string) {
   if (match === '/') return pathname === '/';
   return pathname === match || pathname.startsWith(match + '/');
 }
 
+function centerInStrip(link: HTMLAnchorElement | null) {
+  const strip = link?.parentElement;
```
neha-tanti, line 39: **⚠️ New — implicit markup coupling:** `centerInStrip` infers the strip from `parentElement`, so adding any wrapper element (the mobile panel already wraps its links in `<li>`) turns this into a silent no-op. Use `link.closest('nav')` — that resolves the scroll container regardless of nesting.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up, round 10)

### Verdict: MINOR CHANGES

> Posted as a comment: the `nexuslabs-ai-bot` account is not configured in this `gh` install.

Reviewed `bc64af6..1cc8de5`.

### Unresolved Issues

All previous SDE2 issues resolved. The round-9 blocking scroll jump is fixed — A/B on `/getting-started/install#next-steps` @1280: old `scrollIntoView` moved `scrollY` 1481 → 1421; the rect math leaves it at 1481.

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/_components/TopNav.tsx:39` | Strip inferred from `parentElement`; a wrapper element silently breaks it | Use `link.closest('nav')` |

### Verified in Chromium (React 19.2.7)

| Check | Result |
| --- | --- |
| Overflow case (strip clamped to 300px) — centres, clamps at both ends, `inView: true`, document `scrollY` delta 0 | ✅ |
| Non-overflow (1280–1500px) — browser clamps `scrollLeft`, no-op as documented | ✅ |
| Client-side nav re-fires the ref (Home → Changelog → Theming → Agents → Home all recentred) | ✅ |
| React 19 ref detach passes `null`; the `link?.parentElement` guard handles it | ✅ |
| `dir="rtl"` — `scrollLeft` reaches the RTL min `-529`, link fully in view | ✅ |
| Nav computes `scroll-behavior: auto` — nothing animates under reduced motion | ✅ |
| `pnpm lint`, `pnpm --filter @nexus_ds/docs typecheck` | clean |
| `pnpm test:unit` | only the documented pre-existing `scripts/audit-agent-drift.test.js` failure |

---

PR #683, opened 2026-09-04
feat(docs): add Shiki syntax highlighting themed with Nexus tokens

neha-tanti:
## Summary

- Register `rehype-pretty-code` (Shiki) by string name alongside `rehype-slug` in `mdx-options.ts`, so Turbopack can serialise the loader options.
- Theme it with `NEXUS_CODE_THEME`, whose every colour is a Nexus token reference (`var(--nx-color-*)`) rather than a resolved value. Shiki writes those references straight into the markup, so **no hex colour is emitted anywhere**.
- `keepBackground: false` drops Shiki's own surface — the block keeps the Nexus `container` fill and `border-default` it already had.
- `defaultLang: { block: 'plaintext' }` substitutes a language for a fence that declares none; a fence whose language has no grammar is caught by the plugin's own try/catch.
- Shiki makes the scroll container a tab stop (`tabindex="0"`), so the `pre` takes the same focus ring as the scrollable table wrapper.
- `mdx-options.test.ts` pins tokenisation, the token-reference-only output, the plaintext fallback, and the untouched inline code.

### Dual-theming comes from the tokens, not from Shiki

The original approach used Shiki's dual-theme mode: two baked palettes written to `--shiki-light` / `--shiki-dark`, with CSS choosing between them on the `.dark` class. That mechanism exists because hex values cannot adapt.

Token references can. Each token already carries a light value and a `.dark` override, so a single theme is enough and the CSS mapping is gone:

```
Shiki emits  var(--nx-color-chart-categorical-4)
Nexus token   light  oklch(0.58  0.2489 17.585)
              .dark  oklch(0.885 0.0771 10.001)
```

Keeping Shiki's dual-theme mode would write the _identical_ reference into both custom properties and then spend a CSS rule choosing between two equal values — two switching layers on one colour, the outer one a no-op. Code colour now lives in exactly one place: the design system.

### Why `chart-categorical-*` for the syntax roles

Nexus's semantic foreground tokens are **achromatic under a neutral brand** — `primary-subtle-foreground` resolves to `oklch(0.1448 0 0)` in light and pure white in dark, so keywords rendered identically to plain text. Five of the eight syntax roles came out greyscale; 85% of spans had no hue.

`chart-categorical-*` is the only shipped family whose members are guaranteed mutually distinguishable (chroma 0.12–0.29 across hues 186°/131°/41°/17°/277°), APCA-gated and mode-aware. Syntax highlighting is categorical encoding, so the family fits by purpose.

[...]

PR #685, opened 2026-09-04
fix(tooling): let each machine point the browser MCP at its own binary

### LINKED ISSUE #684: fix(tooling): chrome-devtools MCP hardcodes a macOS browser path
`.mcp.json` pins the `chrome-devtools` MCP server to a macOS Brave install:

```json
"--executable-path=/Applications/Brave Browser.app/Contents/MacOS/Brave Browser"
```

On Linux or Windows every browser tool call fails with:

```
Browser was not found at the configured executablePath (/Applications/Brave Browser.app/Contents/MacOS/Brave Browser)
```

That silently removes browser-driven verification (screenshots, `ui-audit`, checking a change in the running app) for anyone not on a Mac.

## Goal

The shared MCP config works on every platform the team develops on, without contributors editing a checked-in file.

## Scope

- Make the executable path overridable per machine, keeping the current macOS path as the default so Mac contributors see no change.
- Document the override in `CONTRIBUTING.md` alongside the docs-MCP setup.

## Acceptance criteria

- A non-macOS contributor can point the server at their own Chromium-family binary without touching `.mcp.json`.
- macOS contributors need no change.
- The override is documented next to the existing MCP setup instructions.

neha-tanti:
## Summary

- `.mcp.json` pinned `chrome-devtools-mcp` to `/Applications/Brave Browser.app/...`, so every browser tool call failed on Linux and Windows with `Browser was not found at the configured executablePath`. The path now reads `${NEXUS_BROWSER_PATH:-<the macOS path>}`, so macOS contributors see no change and everyone else overrides it from their gitignored `.claude/settings.local.json`.
- `CONTRIBUTING.md` documents the override next to the existing docs-MCP setup.

## GitHub Issue

Closes #684

## Test Plan

- [x] Drove the server over stdio with the Windows Brave path (`C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe`): `initialize` returns `chrome_devtools 1.8.0` and `new_page` opens the target URL. Before the fix the same call returned `Browser was not found at the configured executablePath`.
- [x] Used the now-working server to complete the browser verification that was missing from #683 — loaded the built docs, measured token colours in both themes, and captured light/dark screenshots.
- [x] `.mcp.json` and `.claude/settings.local.json` both parse as JSON.
- [x] Default unchanged for macOS: with `NEXUS_BROWSER_PATH` unset the expansion yields the original Brave path byte-for-byte.

## Notes

The documented value uses forward slashes. A Windows path with backslashes is legal JSON only when escaped (`C:\Program Files\...`), and an unescaped one silently produces a broken string — forward slashes avoid the trap and Node accepts them.

The other MCP server in `.mcp.json`, `nexus-docs-mcp`, is still unreachable on this machine, but that is not a config bug: it needs Docker (`make up`), which is not installed here. No change made.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -8,7 +8,7 @@
       "command": "npx",
       "args": [
         "chrome-devtools-mcp@latest",
-        "--executable-path=/Applications/Brave Browser.app/Contents/MacOS/Brave Browser"
+        "--executable-path=${NEXUS_BROWSER_PATH:-/Applications/Brave Browser.app/Contents/MacOS/Brave Browser}"
       ]
     },
     "nexus-docs-mcp": {
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -305,6 +305,29 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 
 ---
 
+## Browser MCP (chrome-devtools)
+
+`.mcp.json` also wires up `chrome-devtools-mcp`, which drives a real browser so
+Claude Code can screenshot and inspect a running app (see
+[`ui-audit`](.agents/skills/ui-audit-guide/SKILL.md)).
+
+It defaults to Brave at the macOS install path. On Linux or Windows, point
+`NEXUS_BROWSER_PATH` at your own Chromium-family binary — in your gitignored
+`.claude/settings.local.json`, so the shared config stays untouched:
+
+```json
+{
+  "env": {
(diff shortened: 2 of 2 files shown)


INNOVATIVEGAMER:
Closing — we don't need this change.

This was superseded by #693, which carried the same `.mcp.json` fix plus the Codex side. Both are being closed; the per-machine browser MCP path isn't something we want to take on right now. Leaving the branch in place in case we revisit it.

Issue #675, opened 2026-08-27
feat(docs): add Shiki syntax highlighting themed with Nexus tokens


neha-tanti:
## Context

`mdx-components.tsx` gives `<pre>` a background and a border and nothing else — no tokenisation at all.

Under the copy-paste distribution model the code block **is** the product: it is how a consumer gets a Nexus component into their project. Unhighlighted code is the least readable form of the one thing the site exists to hand over.

Shiki is the right choice over Prism here. It tokenises with TextMate grammars at build time and ships no runtime library, where Prism is regex-based and breaks on nested languages. In dual-theme mode Shiki emits both themes' colours as CSS custom properties, so the existing `NexusAppearanceProvider` toggle recolours every block with zero JavaScript and no flash — which fits the flash-free theming already in place.

## Goal

Code blocks are tokenised at build time and follow the site theme with no client-side work.

## Scope

- Add `rehype-pretty-code` with Shiki in **dual-theme mode**, registered by string name (Turbopack constraint).
- Choose the light and dark themes, then map Shiki's emitted CSS variables onto Nexus token variables so code sits in the same colour system as the rest of the page rather than beside it.
- Cover the languages the docs actually use: `tsx`, `ts`, `bash`, `json`, `css`.
- Confirm no highlighting library reaches the client bundle.

## Acceptance criteria

- A `tsx` block is tokenised and readable.
- Toggling the theme recolours every code block with no flash and no client-side JavaScript.
- Bundle analysis shows no highlighter in the client bundle.
- An unknown or missing language degrades to plain text instead of failing the build.

## Notes

Ticket **3 of 7 in Stage 1** of the docs build order. Independent.

Ticket 4 verifies this output against the Content-Security-Policy. **Do not switch the CSP from Report-Only to enforcing before that lands** — Shiki writes inline styles.

PR #689, opened 2026-09-08
feat(docs): add a copy button to every code block

### LINKED ISSUE #677: feat(docs): add a copy button to every code block
## Context

Copy-paste is the only supported way to adopt a Nexus component. `pnpm export` forks the entire monorepo, and there is no registry. That makes the copy button the **distribution mechanism**, not a convenience.

Today `mdx-components.tsx` styles `<pre>` but offers no way to get the code out of the page except selecting it by hand.

## Goal

Every code block on the docs site can be copied in one click, complete and paste-ready.

## Scope

- Add a copy control to the `<pre>` override in `mdx-components.tsx`.
- Copy the **full block, imports included** — never a trimmed excerpt. A snippet without its import lines cannot be pasted into a project, which defeats the entire distribution model.
- Show a confirmation state after copying, and reset it.
- Keyboard reachable, with the canonical focus treatment from `components.md` § Focus States and an accessible name.
- Make sure the control's own label never ends up inside the copied string.

## Acceptance criteria

- Copied text matches the source block exactly, including imports and trailing-newline handling.
- The control is reachable and operable by keyboard with a visible focus ring.
- The confirmation is announced to assistive technology, not conveyed by colour alone.
- Works in both light and dark themes.

## Notes

Ticket **5 of 7 in Stage 1** of the docs build order. Independent, though it will look better once syntax highlighting is in.

neha-tanti:
## Summary

- Replace the MDX `<pre>` override with a `CodeBlock` client component that pairs every fenced block with an icon copy control.
- Copy `<pre>.textContent` **verbatim** — imports and the fence's trailing newline included, nothing trimmed. The control is a *sibling* of the `<pre>`, not a child, so its own label can never land in the copied string (and a hand-selection of the block doesn't pick it up either). Reading the DOM rather than the React children also keeps the copy exact once Shiki (#675) wraps the source in nested spans.
- Confirmation is announced through a `role="status"` region and carried by an icon swap, so it is never conveyed by colour alone; it resets to idle after 2s. A rejected clipboard write reports rather than failing silently.
- Reserve a focus-safe band at the bottom of the scrollport, sized from the theme panel's measured box plus the gap it floats in, so Tab never parks a copy control underneath it (WCAG 2.4.11) — see *Focus obstruction* below.
- Give `@nexus_ds/react` a `./utils` subpath so the server-side `mdx-components.tsx` can reach `cn` without importing the client-only root barrel, and drop `cn` from that barrel so it has one entry point.

## GitHub Issue

Closes #677

## Focus obstruction (WCAG 2.4.11)

The copy control sits at the code block's top-right, which is exactly where the fixed `ThemePicker` panel (`nx:fixed nx:bottom-6 nx:right-6 nx:z-popover`, x 956–1256) overlays the prose column (right edge 1024).

Measured by tabbing through `/getting-started/install` at 1280×900 and hit-testing each focused control:

| | copy controls focused | **entirely** occluded while focused |
| --- | --- | --- |
| Before | 10 | **8** |
| After | 10 | **0** |

The fix follows the shell's existing idiom — the layout already sets `nx:scroll-pt-(--docs-scroll-offset)` on `<html>` so the sticky header can't cover a scroll target. This adds the symmetric `nx:scroll-pb-(--docs-panel-offset)`, which `ThemePanel` fills under a `ResizeObserver` from its own `getBoundingClientRect()`: the panel's height plus **twice** the measured gap beneath it, so the band clears the panel and leaves the same breathing room above it that `bottom-6` leaves below. The reservation therefore tracks the panel it reserves for, and the two cannot drift apart. The panel caps itself at `calc(100svh - 2*var(--nx-spacing-6))` with `overflow-y-auto`, gaining a scrollbar instead of overflowing a short viewport. This fixes the class of bug for every focusable in the docs, not just the copy control.

Because that value feeds the global reservation directly, the panel's default state below `lg` is a layout constraint, not just a preference: an expanded panel at 390×844 publishes roughly half the scrollport as `scroll-pb`. That is why `EXPAND_QUERY` is written as `(min-width: 64rem)` rather than a `max-width` collapse query — its *unmatched* value, the one the prerendered HTML and the first client render both use, is then the collapsed one, so a narrow viewport opens collapsed and never publishes the oversized reservation on first paint. The query governs the default, not a ceiling: tapping the toggle on mobile still expands the panel and still publishes the larger band, which is the correct trade once the user has asked for the panel. Staying in rem keeps the threshold aligned with the `nx:lg:` utilities as the user's base font size changes.

## Design notes

- **Icon-only, always visible.** Copy-paste is the distribution mechanism for Nexus components, not a convenience — so the control does not hide behind hover, which would strand touch and keyboard users.
- **Stable accessible name.** The button stays `aria-label="Copy code"` in every state; the state change is announced once by the live region, rather than twice by also mutating the button's name under a focused screen reader.
- **`textContent`, not `innerText`.** `innerText` would collapse the block's whitespace; `textContent` returns the source byte-for-byte.
- Reached for `@tabler/icons-react` (already the monorepo's icon set, used by `apps/console` and `packages/react/src/lib/icons.ts`) rather than inlining SVGs.

## Test Plan

- [x] `apps/docs/app/_components/CodeBlock.test.tsx` — 10 tests: verbatim copy (imports + trailing newline, no control label), announce-then-idle, failed-write announcement, empty-block reporting, repeat copy, per-block icon isolation, a sibling's confirmation surviving another block's reset, a caller-injected `tabIndex` losing to the scroll container's tab stop, and the focusable native button. 10/10 pass.
- [x] Full unit suite: 662 passed / 1 failed / 55 skipped. The one failure (`scripts/audit-agent-drift.test.js`) and the `generate-tailwind-package` suite error reproduce identically on `main` with these changes stashed — pre-existing and unrelated.
- [x] `pnpm turbo run typecheck` — 7/7 successful.
- [x] `npx eslint . --max-warnings 0` — clean. Prettier clean on all changed files.
- [x] `pnpm --filter @nexus_ds/docs build` — 44/44 static pages; `audit:csp` inventory unchanged in shape.
- [x] Real Chromium (Playwright), light **and** dark: focus ring paints (2px surface gap + 2px outer ring, per `components.md` § Focus States), Enter copies, focus is retained through the state change, status region announces then clears, 0 console errors.
- [x] Focus-obstruction sweep at 1280×900, 1280×600, and 390×844 (mobile reference): panel not clipped, 0 of 10–15 copy controls occluded on focus.
- [x] Verified the button never overlaps code: `pre` carries `nx:pe-14`, button starts 16px clear of the content edge.

**Note on clipboard read-back:** a Playwright `navigator.clipboard.readText()` round-trip returns `"…core\r\n"` for a source of `"…core\n"`. That is the Windows OS clipboard normalising LF→CRLF on read, not a transform in this code — the string handed to `writeText` is pinned byte-for-byte by the unit test.

## Not done

- `#681` (`wldkk/issue-677-add-code-copy-button`) is an earlier open PR against this same issue, currently conflicting with `main`. This PR supersedes it; #681 should be closed if this one lands.
- Browser MCP verification was unavailable on this machine (it points at a hardcoded macOS Brave path — the subject of open PR #685), so the browser evidence above came from Playwright directly.

### DIFF
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -228,7 +228,7 @@ describe('CodeBlock', () => {
   });
 
   it('gives the scroll container a tab stop a caller cannot remove', () => {
-    // The props type omits `tabIndex`, but a rehype plugin can still inject one
+    // The props type omits `tabIndex`, but an MDX plugin can still inject one
     // at runtime — the guard below the spread is what has to hold there.
     const injected: object = { tabIndex: -1 };
     const { container } = render(
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -90,6 +90,69 @@ describe('CodeBlock', () => {
     expect(copy.dataset.copyStatus).toBe('idle');
   });
 
+  it('restarts the reset window and re-announces on a repeat copy', async () => {
+    const copy = renderBlock();
+    const region = screen.getByRole('status');
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+    const firstAnnouncement = region.firstElementChild;
+
+    act(() => {
+      vi.advanceTimersByTime(1900);
+    });
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+
+    // The first click's timer would have fired 100ms from here.
+    act(() => {
+      vi.advanceTimersByTime(1900);
+    });
+    expect(copy.dataset.copyStatus).toBe('copied');
+    expect(region.firstElementChild).not.toBe(firstAnnouncement);
+    expect(region.textContent).toBe('Code copied to clipboard');
+
+    act(() => {
+      vi.advanceTimersByTime(100);
+    });
+    expect(copy.dataset.copyStatus).toBe('idle');
+  });
+
+  it('reports a failure when the clipboard API is unavailable', async () => {
+    Object.defineProperty(navigator, 'clipboard', {
+      configurable: true,
+      value: undefined,
+    });
+    const copy = renderBlock();
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+
+    expect(screen.getByRole('status').textContent).toBe('Could not copy code');
+    expect(copy.dataset.copyStatus).toBe('failed');
+  });
+
+  it('reports a failure rather than a silent no-op for an empty block', async () => {
+    render(
+      <CodeBlock>
+        <code />
+      </CodeBlock>
+    );
+    const copy = screen.getByRole('button', { name: 'Copy code' });
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+
+    expect(writeText).not.toHaveBeenCalled();
+    expect(screen.getByRole('status').textContent).toBe('Could not copy code');
+    expect(copy.dataset.copyStatus).toBe('failed');
+  });
+
   it('exposes the control as a focusable native button', () => {
     const copy = renderBlock();
 
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -77,8 +77,6 @@ describe('CodeBlock', () => {
     });
 
     expect(copy.dataset.copyStatus).toBe('idle');
-    // The block restores its own icon; the provider clears the region it owns,
-    // in place, so emptying it is not itself read out.
     expect(screen.getByRole('status').textContent).toBe('');
     expect(screen.getByRole('status').firstElementChild).toBe(announcement);
   });
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -57,7 +57,7 @@ describe('CodeBlock', () => {
     expect(writeText.mock.calls[0]![0]).not.toMatch(/copy/i);
   });
 
-  it('announces the confirmation, then resets it', async () => {
+  it('announces the confirmation, then returns the control to idle', async () => {
     const copy = renderBlock();
 
     expect(screen.getByRole('status').textContent).toBe('');
@@ -70,13 +70,17 @@ describe('CodeBlock', () => {
       'Code copied to clipboard'
     );
     expect(copy.dataset.copyStatus).toBe('copied');
+    const announcement = screen.getByRole('status').firstElementChild;
 
     act(() => {
       vi.advanceTimersByTime(2000);
     });
 
-    expect(screen.getByRole('status').textContent).toBe('');
     expect(copy.dataset.copyStatus).toBe('idle');
+    // The block restores its own icon; the provider clears the region it owns,
+    // in place, so emptying it is not itself read out.
+    expect(screen.getByRole('status').textContent).toBe('');
+    expect(screen.getByRole('status').firstElementChild).toBe(announcement);
   });
 
   it('announces a failed write instead of leaving a silent no-op', async () => {
@@ -184,6 +188,68 @@ describe('CodeBlock', () => {
       'Code copied to clipboard'
     );
     expect(controls[0]!.dataset.copyStatus).toBe('idle');
+  });
+
+  it("leaves another block's confirmation in the shared region on reset", async () => {
+    render(
+      <CopyAnnouncerProvider>
+        <CodeBlock>
+          <code>{SNIPPET}</code>
+        </CodeBlock>
+        <CodeBlock>
+          <code>{'const x = 1;\n'}</code>
+        </CodeBlock>
+      </CopyAnnouncerProvider>
+    );
+    const controls = screen.getAllByRole('button', { name: 'Copy code' });
+    const region = screen.getByRole('status');
+
+    await act(async () => {
+      fireEvent.click(controls[0]!);
+    });
+    act(() => {
+      vi.advanceTimersByTime(1000);
+    });
+    await act(async () => {
+      fireEvent.click(controls[1]!);
+    });
+    // Both blocks announce the same words, so identity — not text — is what
+    // separates "left alone" from "cleared" and from "announced again".
+    const announcement = region.firstElementChild;
+
+    // The first block's 2s reset lands here, a second after the other block
+    // announced. It must return only its own icon to idle.
+    act(() => {
+      vi.advanceTimersByTime(1000);
+    });
+
+    expect(controls[0]!.dataset.copyStatus).toBe('idle');
+    expect(controls[1]!.dataset.copyStatus).toBe('copied');
+    expect(region.firstElementChild).toBe(announcement);
+    expect(region.textContent).toBe('Code copied to clipboard');
+  });
+
+  it('gives the scroll container a tab stop a caller cannot remove', () => {
+    // The props type omits `tabIndex`, but a rehype plugin can still inject one
+    // at runtime — the guard below the spread is what has to hold there.
+    const injected: object = { tabIndex: -1 };
+    const { container } = render(
+      <CopyAnnouncerProvider>
+        <CodeBlock>
+          <code>{SNIPPET}</code>
+        </CodeBlock>
+        <CodeBlock {...injected}>
+          <code>{SNIPPET}</code>
+        </CodeBlock>
+      </CopyAnnouncerProvider>
+    );
+    const [own, overridden] = Array.from(container.querySelectorAll('pre'));
+
+    expect(own!.tabIndex).toBe(0);
+    expect(overridden!.tabIndex).toBe(0);
+
+    own!.focus();
+    expect(document.activeElement).toBe(own);
   });
 
   it('exposes the control as a focusable native button', () => {
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -6,8 +6,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { CodeBlock } from './CodeBlock';
 import { CopyAnnouncerProvider } from './CopyAnnouncer';
 
-// Shaped like a real block from content/getting-started/install.mdx: the import
-// line is what a trimmed excerpt would drop, and the fence leaves a trailing \n.
+// A real block shape: an import line, and the fence's trailing newline.
 const SNIPPET = `import { Button } from '@nexus_ds/react';
 
 export function Example() {
@@ -17,8 +16,7 @@ export function Example() {
 
 const writeText = vi.fn<(text: string) => Promise<void>>();
 
-// The live region is shared page-wide, so a block is only testable inside the
-// provider that owns it — the same wiring `app/layout.tsx` supplies.
+// The live region lives in the provider, so a block is only testable inside one.
 function renderInDocs(block: React.ReactNode) {
   render(<CopyAnnouncerProvider>{block}</CopyAnnouncerProvider>);
   return screen.getByRole('button', { name: 'Copy code' });
@@ -211,12 +209,10 @@ describe('CodeBlock', () => {
     await act(async () => {
       fireEvent.click(controls[1]!);
     });
-    // Both blocks announce the same words, so identity — not text — is what
-    // separates "left alone" from "cleared" and from "announced again".
+    // Both blocks announce the same words, so identity separates the cases.
     const announcement = region.firstElementChild;
 
-    // The first block's 2s reset lands here, a second after the other block
-    // announced. It must return only its own icon to idle.
+    // The first block's 2s reset lands here, after the other block announced.
     act(() => {
       vi.advanceTimersByTime(1000);
     });
@@ -228,8 +224,7 @@ describe('CodeBlock', () => {
   });
 
   it('gives the scroll container a tab stop a caller cannot remove', () => {
-    // The props type omits `tabIndex`, but an MDX plugin can still inject one
-    // at runtime — the guard below the spread is what has to hold there.
+    // The props type omits `tabIndex`, but an MDX plugin can still inject one.
     const injected: object = { tabIndex: -1 };
     const { container } = render(
       <CopyAnnouncerProvider>
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -4,6 +4,7 @@ import { act, fireEvent, render, screen } from '@testing-library/react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import { CodeBlock } from './CodeBlock';
+import { CopyAnnouncerProvider } from './CopyAnnouncer';
 
 // Shaped like a real block from content/getting-started/install.mdx: the import
 // line is what a trimmed excerpt would drop, and the fence leaves a trailing \n.
@@ -16,13 +17,19 @@ export function Example() {
 
 const writeText = vi.fn<(text: string) => Promise<void>>();
 
+// The live region is shared page-wide, so a block is only testable inside the
+// provider that owns it — the same wiring `app/layout.tsx` supplies.
+function renderInDocs(block: React.ReactNode) {
+  render(<CopyAnnouncerProvider>{block}</CopyAnnouncerProvider>);
+  return screen.getByRole('button', { name: 'Copy code' });
+}
+
 function renderBlock() {
-  render(
+  return renderInDocs(
     <CodeBlock>
       <code className="language-tsx">{SNIPPET}</code>
     </CodeBlock>
   );
-  return screen.getByRole('button', { name: 'Copy code' });
 }
 
 describe('CodeBlock', () => {
@@ -137,12 +144,11 @@ describe('CodeBlock', () => {
   });
 
   it('reports a failure rather than a silent no-op for an empty block', async () => {
-    render(
+    const copy = renderInDocs(
       <CodeBlock>
         <code />
       </CodeBlock>
     );
-    const copy = screen.getByRole('button', { name: 'Copy code' });
 
     await act(async () => {
       fireEvent.click(copy);
@@ -151,6 +157,33 @@ describe('CodeBlock', () => {
     expect(writeText).not.toHaveBeenCalled();
     expect(screen.getByRole('status').textContent).toBe('Could not copy code');
     expect(copy.dataset.copyStatus).toBe('failed');
+  });
+
+  it('shares one live region across every block on the page', async () => {
+    render(
+      <CopyAnnouncerProvider>
+        <CodeBlock>
+          <code>{SNIPPET}</code>
+        </CodeBlock>
+        <CodeBlock>
+          <code>{'const x = 1;\n'}</code>
+        </CodeBlock>
+      </CopyAnnouncerProvider>
+    );
+    const controls = screen.getAllByRole('button', { name: 'Copy code' });
+
+    expect(controls).toHaveLength(2);
+    expect(screen.getAllByRole('status')).toHaveLength(1);
+
+    await act(async () => {
+      fireEvent.click(controls[1]!);
+    });
+
+    expect(writeText).toHaveBeenCalledWith('const x = 1;\n');
+    expect(screen.getByRole('status').textContent).toBe(
+      'Code copied to clipboard'
+    );
+    expect(controls[0]!.dataset.copyStatus).toBe('idle');
   });
 
   it('exposes the control as a focusable native button', () => {
--- a/apps/docs/app/_components/CodeBlock.test.tsx
+++ b/apps/docs/app/_components/CodeBlock.test.tsx
@@ -1,0 +1,102 @@
+import * as React from 'react';
+
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { CodeBlock } from './CodeBlock';
+
+// Shaped like a real block from content/getting-started/install.mdx: the import
+// line is what a trimmed excerpt would drop, and the fence leaves a trailing \n.
+const SNIPPET = `import { Button } from '@nexus_ds/react';
+
+export function Example() {
+  return <Button>Save</Button>;
+}
+`;
+
+const writeText = vi.fn<(text: string) => Promise<void>>();
+
+function renderBlock() {
+  render(
+    <CodeBlock>
+      <code className="language-tsx">{SNIPPET}</code>
+    </CodeBlock>
+  );
+  return screen.getByRole('button', { name: 'Copy code' });
+}
+
+describe('CodeBlock', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    writeText.mockReset().mockResolvedValue(undefined);
+    Object.defineProperty(navigator, 'clipboard', {
+      configurable: true,
+      value: { writeText },
+    });
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('copies the block verbatim — imports, trailing newline, no control label', async () => {
+    const copy = renderBlock();
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+
+    expect(writeText).toHaveBeenCalledWith(SNIPPET);
+    expect(writeText.mock.calls[0]![0]).not.toMatch(/copy/i);
+  });
+
+  it('announces the confirmation, then resets it', async () => {
+    const copy = renderBlock();
+
+    expect(screen.getByRole('status').textContent).toBe('');
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+
+    expect(screen.getByRole('status').textContent).toBe(
+      'Code copied to clipboard'
+    );
+    expect(copy.dataset.copyStatus).toBe('copied');
+
+    act(() => {
+      vi.advanceTimersByTime(2000);
+    });
+
+    expect(screen.getByRole('status').textContent).toBe('');
+    expect(copy.dataset.copyStatus).toBe('idle');
+  });
+
+  it('announces a failed write instead of leaving a silent no-op', async () => {
+    writeText.mockRejectedValue(new DOMException('denied', 'NotAllowedError'));
+    const copy = renderBlock();
+
+    await act(async () => {
+      fireEvent.click(copy);
+    });
+
+    expect(screen.getByRole('status').textContent).toBe('Could not copy code');
+    expect(copy.dataset.copyStatus).toBe('failed');
+
+    act(() => {
+      vi.advanceTimersByTime(2000);
+    });
+
+    expect(copy.dataset.copyStatus).toBe('idle');
+  });
+
+  it('exposes the control as a focusable native button', () => {
+    const copy = renderBlock();
+
+    expect(copy.tagName).toBe('BUTTON');
+    expect(copy.hasAttribute('tabindex')).toBe(false);
+
+    copy.focus();
+    expect(document.activeElement).toBe(copy);
+  });
+});
--- a/apps/docs/app/_components/CodeBlock.tsx
+++ b/apps/docs/app/_components/CodeBlock.tsx
@@ -41,7 +41,6 @@ export function CodeBlock({
     setStatus(next);
     announce(COPY_STATUS[next].message);
     window.clearTimeout(timerRef.current);
-    // Icon only — the provider owns how long the announcement stays up.
     timerRef.current = window.setTimeout(
       () => setStatus('idle'),
       RESET_DELAY_MS
@@ -68,9 +67,9 @@ export function CodeBlock({
       <pre
         className={cn(PRE_CLASS, className)}
         {...props}
-        // Both below the spread: the props type omits them, but MDX plugins
-        // inject props at runtime where the type cannot reach.
         ref={preRef}
+        // Below the spread: a rehype plugin can set `tabIndex` through
+        // `hProperties`, and the scroll container's own tab stop has to win.
         // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}
       >
--- a/apps/docs/app/_components/CodeBlock.tsx
+++ b/apps/docs/app/_components/CodeBlock.tsx
@@ -68,8 +68,7 @@ export function CodeBlock({
         className={cn(PRE_CLASS, className)}
         {...props}
         ref={preRef}
-        // Below the spread: an MDX plugin can set `tabIndex`, and the scroll
-        // container's own tab stop has to win.
+        // Below the spread so an injected `tabIndex` cannot replace this one.
         // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}
       >
--- a/apps/docs/app/_components/CodeBlock.tsx
+++ b/apps/docs/app/_components/CodeBlock.tsx
@@ -10,39 +10,26 @@ import { Button } from './nexus';
 
 const RESET_DELAY_MS = 2000;
 
-type CopyStatus = 'idle' | 'copied' | 'failed';
-
-const COPY_STATUS: Record<
-  CopyStatus,
-  { icon: React.ReactNode; message: string; textClass: string }
-> = {
-  idle: { icon: <IconCopy />, message: '', textClass: '' },
+const COPY_STATUS = {
+  idle: { icon: <IconCopy />, message: '', className: '' },
   copied: {
     icon: <IconCheck />,
     message: 'Code copied to clipboard',
-    textClass: 'nx:text-success-subtle-foreground',
+    className: 'nx:text-success-subtle-foreground',
   },
   failed: {
     icon: <IconX />,
     message: 'Could not copy code',
-    textClass: 'nx:text-error-subtle-foreground',
+    className: 'nx:text-error-subtle-foreground',
   },
 };
+
+type CopyStatus = keyof typeof COPY_STATUS;
 
 const PRE_CLASS =
   'nx:bg-muted nx:border nx:border-border-default nx:rounded-md nx:p-4 nx:pe-14 nx:overflow-x-auto nx:typography-code-block nx:[&_code]:bg-transparent nx:[&_code]:p-0 nx:[&_code]:typography-code-block';
 
-/**
- * The MDX `<pre>` override. Copy-paste is how Nexus components are adopted, so
- * every fenced block carries a control that yields the whole snippet — imports
- * included — in one click.
- *
- * The control is a sibling of the `<pre>`, not a child: the copied string is
- * read straight off `<pre>.textContent`, so the button's own label can never
- * leak into it, and a hand-selection of the block does not pick it up either.
- * Reading the DOM rather than the React children also keeps the copy exact once
- * syntax highlighting wraps the source in nested spans.
- */
+/** The MDX `<pre>` override: the code block plus a control that copies it. */
 export function CodeBlock({
   children,
   className,
@@ -50,7 +37,7 @@ export function CodeBlock({
 }: React.ComponentProps<'pre'>) {
   const preRef = React.useRef<HTMLPreElement>(null);
   const [status, setStatus] = React.useState<CopyStatus>('idle');
-  const { icon, message, textClass } = COPY_STATUS[status];
+  const { icon, message, className: statusClass } = COPY_STATUS[status];
 
   React.useEffect(() => {
     if (status === 'idle') return;
@@ -60,8 +47,6 @@ export function CodeBlock({
   }, [status]);
 
   const handleCopy = async () => {
-    // textContent, not innerText: it keeps the source verbatim, including the
-    // fence's trailing newline, where innerText would collapse the whitespace.
     const source = preRef.current?.textContent;
     if (!source) return;
 
@@ -79,12 +64,11 @@ export function CodeBlock({
         {children}
       </pre>
       <Button
-        type="button"
         variant="outline"
         size="icon-sm"
         aria-label="Copy code"
         data-copy-status={status}
-        className={join('nx:absolute nx:top-2 nx:end-2', textClass)}
+        className={join('nx:absolute nx:top-2 nx:end-2', statusClass)}
         onClick={handleCopy}
       >
         {icon}
--- a/apps/docs/app/_components/CodeBlock.tsx
+++ b/apps/docs/app/_components/CodeBlock.tsx
@@ -1,0 +1,97 @@
+'use client';
+
+import * as React from 'react';
+
+import { IconCheck, IconCopy, IconX } from '@tabler/icons-react';
+
+import { join } from '../_lib/class-names';
+
+import { Button } from './nexus';
+
+const RESET_DELAY_MS = 2000;
+
+type CopyStatus = 'idle' | 'copied' | 'failed';
+
+const COPY_STATUS: Record<
+  CopyStatus,
+  { icon: React.ReactNode; message: string; textClass: string }
+> = {
+  idle: { icon: <IconCopy />, message: '', textClass: '' },
+  copied: {
+    icon: <IconCheck />,
+    message: 'Code copied to clipboard',
+    textClass: 'nx:text-success-subtle-foreground',
+  },
+  failed: {
+    icon: <IconX />,
+    message: 'Could not copy code',
+    textClass: 'nx:text-error-subtle-foreground',
+  },
+};
+
+const PRE_CLASS =
+  'nx:bg-muted nx:border nx:border-border-default nx:rounded-md nx:p-4 nx:pe-14 nx:overflow-x-auto nx:typography-code-block nx:[&_code]:bg-transparent nx:[&_code]:p-0 nx:[&_code]:typography-code-block';
+
+/**
+ * The MDX `<pre>` override. Copy-paste is how Nexus components are adopted, so
+ * every fenced block carries a control that yields the whole snippet — imports
+ * included — in one click.
+ *
+ * The control is a sibling of the `<pre>`, not a child: the copied string is
+ * read straight off `<pre>.textContent`, so the button's own label can never
+ * leak into it, and a hand-selection of the block does not pick it up either.
+ * Reading the DOM rather than the React children also keeps the copy exact once
+ * syntax highlighting wraps the source in nested spans.
+ */
+export function CodeBlock({
+  children,
+  className,
+  ...props
+}: React.ComponentProps<'pre'>) {
+  const preRef = React.useRef<HTMLPreElement>(null);
+  const [status, setStatus] = React.useState<CopyStatus>('idle');
+  const { icon, message, textClass } = COPY_STATUS[status];
+
+  React.useEffect(() => {
+    if (status === 'idle') return;
+
+    const timer = window.setTimeout(() => setStatus('idle'), RESET_DELAY_MS);
+    return () => window.clearTimeout(timer);
+  }, [status]);
+
+  const handleCopy = async () => {
+    // textContent, not innerText: it keeps the source verbatim, including the
+    // fence's trailing newline, where innerText would collapse the whitespace.
+    const source = preRef.current?.textContent;
+    if (!source) return;
+
+    try {
+      await navigator.clipboard.writeText(source);
+      setStatus('copied');
+    } catch {
+      setStatus('failed');
+    }
+  };
+
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre ref={preRef} className={join(PRE_CLASS, className)} {...props}>
+        {children}
+      </pre>
+      <Button
+        type="button"
+        variant="outline"
+        size="icon-sm"
+        aria-label="Copy code"
+        data-copy-status={status}
+        className={join('nx:absolute nx:top-2 nx:end-2', textClass)}
+        onClick={handleCopy}
+      >
+        {icon}
+      </Button>
+      <span role="status" className="nx:sr-only">
+        {message}
+      </span>
+    </div>
+  );
+}
--- a/apps/docs/app/_components/CodeBlock.tsx
+++ b/apps/docs/app/_components/CodeBlock.tsx
@@ -68,8 +68,8 @@ export function CodeBlock({
         className={cn(PRE_CLASS, className)}
         {...props}
         ref={preRef}
-        // Below the spread: a rehype plugin can set `tabIndex` through
-        // `hProperties`, and the scroll container's own tab stop has to win.
+        // Below the spread: an MDX plugin can set `tabIndex`, and the scroll
+        // container's own tab stop has to win.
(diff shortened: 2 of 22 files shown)


apps/docs/app/globals.css
```diff
@@ -17,4 +17,8 @@
 :root {
   --docs-header-h: 3.5rem;
   --docs-scroll-offset: calc(var(--docs-header-h) + 1.5rem);
+  /* Caps the floating theme panel and reserves the same band at the bottom of
+     the scrollport, so Tab never parks a control underneath it. */
+  --docs-panel-h: 30rem;
```
neha-tanti: **Blocking — fixed rem cap on a viewport-anchored panel:** `30rem` plus `bottom-6` overflows any viewport shorter than ~504px (landscape phone), clipping the panel head off-screen with no way to reach it. Cap with `max-h-[calc(100svh-3rem)]` instead.

apps/docs/app/layout.tsx
```diff
@@ -30,7 +30,7 @@ export default function RootLayout({
   return (
     <html
       lang="en"
-      className="nx:scroll-pt-(--docs-scroll-offset)"
+      className="nx:scroll-pt-(--docs-scroll-offset) nx:scroll-pb-(--docs-panel-offset)"
```
neha-tanti, line 34: **Blocking — global scroll reservation is oversized:** 504px of `scroll-pb` leaves ~260px of usable scrollport at 390×844, and still applies on `/`, where `ThemePicker` returns `null` (ThemePicker.tsx:31). Size the reservation to the panel actually rendered.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  React.useEffect(() => {
+    if (status === 'idle') return;
+
+    const timer = window.setTimeout(() => setStatus('idle'), RESET_DELAY_MS);
+    return () => window.clearTimeout(timer);
+  }, [status]);
+
+  const handleCopy = async () => {
+    const source = preRef.current?.textContent;
```
neha-tanti, line 51: **Concern — DOM scrape breaks under #675:** `textContent` reads rendered output, which Shiki will restructure into per-line spans. Carry the raw fence value through the MDX pipeline as a prop instead of reading it back off the DOM.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+        variant="outline"
+        size="icon-sm"
+        aria-label="Copy code"
+        data-copy-status={status}
+        className={join('nx:absolute nx:top-2 nx:end-2', statusClass)}
+        onClick={handleCopy}
+      >
+        {icon}
+      </Button>
+      <span role="status" className="nx:sr-only">
```
neha-tanti: **Minor — live region per block:** a page with twenty fences mounts twenty `role="status"` regions. Hoist one shared region into the layout and publish the message to it.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <div className="nx:relative nx:mb-4">
+      <pre ref={preRef} className={join(PRE_CLASS, className)} {...props}>
+        {children}
+      </pre>
+      <Button
+        variant="outline"
+        size="icon-sm"
+        aria-label="Copy code"
+        data-copy-status={status}
+        className={join('nx:absolute nx:top-2 nx:end-2', statusClass)}
```
neha-tanti: **Minor — state expressed twice:** `statusClass` and `data-copy-status` both encode status. Per `components.md` the data attribute is the styling hook — key the colour off `nx:data-[copy-status=copied]:…` and drop `className` from the record.

neha-tanti (commented):

### Verdict: CHANGES REQUIRED

### Issues

| 1 | ❌ Blocking | `_pages/Typography.tsx:184`, `Responsive.tsx:159`, `MultiBrand.tsx:164` | Hand-rolled `<pre>` blocks on three live routes bypass the MDX override, so they get no copy button | Export `CodeBlock` and render it in those three pages; delete the duplicated `pre` class strings |
| 2 | ❌ Blocking | `app/globals.css:22` | `--docs-panel-h: 30rem` is a fixed rem cap on a viewport-anchored panel | Cap with `max-h-[calc(100svh-3rem)]` so the panel cannot exceed the viewport |
| 3 | ❌ Blocking | `app/layout.tsx:33` | 504px global `scroll-pb` eats ~60% of a 390×844 scrollport, and applies on `/` where `ThemePicker` returns `null` | Size the reservation to the panel actually rendered |
| 4 | ⚠️ Minor | `_components/CodeBlock.tsx:76` | One `role="status"` live region per code block instead of one shared region | Hoist a single region into the layout |
| 5 | ⚠️ Minor | `_components/CodeBlock.tsx:71` | Status is expressed twice — `statusClass` in `className` and `data-copy-status` | Key the colour off `nx:data-[copy-status=copied]:…` so the data attribute is the single hook |
| 6 | ⚠️ Minor | `app/_lib/class-names.ts:2` | `join` concatenates without merge semantics, so a caller `className` cannot override a base utility | Use `tailwind-merge` |

### Architectural Concerns

**Copy source is scraped from the DOM** (`CodeBlock.tsx:50`) — `preRef.current?.textContent` reads rendered output, which #675 will replace with nested Shiki line spans whose newline handling is a renderer detail. Stash the raw fence value on the `pre` node in the MDX pipeline and read it as a prop.

**`<html>` scroll padding is hand-tuned to one component's height** (`globals.css:22`) — a guessed `30rem` constant now governs global anchor and focus scrolling, and it scales past the viewport under an enlarged browser default font (see `responsive.md`). Derive the reservation from the rendered panel or scope it to the pages that render `ThemePicker`.

---

apps/docs/app/_lib/class-names.ts
```diff
@@ -0,0 +1,3 @@
+/** Appends a caller-supplied className to a base class string. */
+export const join = (base: string, incoming?: string) =>
```
neha-tanti: **Minor — no merge semantics:** promoting this to a shared module blesses plain concatenation app-wide. A caller `className` cannot override a base utility, since attribute order carries no specificity. Use `tailwind-merge` here.

neha-tanti (commented):

### Verdict: CHANGES REQUIRED

### Issues

| 1 | ❌ Blocking | `_components/CodeBlock.tsx:55` | Repeat copy while already `copied`: React bails out, so the reset timer never restarts and nothing is announced | Store a changing token (`{ status, seq }`) so the state genuinely changes |
| 2 | ❌ Blocking | `_components/CodeBlock.test.tsx:73` | No test for the repeat-click path the acceptance criteria depend on | Add a second-click assertion plus an absent-`navigator.clipboard` case |
| 3 | ⚠️ Minor | `_components/CodeBlock.tsx:51` | Empty block: silent no-op, so the button reads as broken | Set `failed` rather than returning |
| 4 | ⚠️ Minor | `_components/CodeBlock.tsx:63` | `overflow-x-auto` `<pre>` has no tab stop or focus ring | Mirror the `table` wrapper in `mdx-components.tsx:110-113` |

### Checked and not flagged

`COPY_STATUS` without `as const` still yields the intended union and rejects a typo'd key (`tsc --noEmit` clean). The `TypeError` from an undefined `navigator.clipboard` is raised inside the `try`, so the bare `catch` does surface `failed` — confirmed at runtime. `nx:text-success-subtle-foreground` and `nx:text-error-subtle-foreground` are both in `SEMANTIC_TOKEN_REGISTRY`. `{...props}` does not land on the wrong element; `rehype-slug` only touches headings.

---

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return () => window.clearTimeout(timer);
+  }, [status]);
+
+  const handleCopy = async () => {
+    const source = preRef.current?.textContent;
+    if (!source) return;
+
+    try {
+      await navigator.clipboard.writeText(source);
+      setStatus('copied');
```
neha-tanti: **Blocking — repeat copy is silent:** `setStatus('copied')` while already `copied` is a React bailout: the effect never re-runs and the live-region text node is untouched. A second click at t=1.9s reverts the icon 100ms later and announces nothing. Store a changing token (`{ status, seq }`).

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+  React.useEffect(() => {
+    if (status === 'idle') return;
+
+    const timer = window.setTimeout(() => setStatus('idle'), RESET_DELAY_MS);
+    return () => window.clearTimeout(timer);
+  }, [status]);
+
+  const handleCopy = async () => {
+    const source = preRef.current?.textContent;
+    if (!source) return;
```
neha-tanti: **Minor — empty block gives no feedback:** verified `writeText` is never called and `status` stays `idle`, so the button reads as broken. Set `failed` instead of returning.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+      await navigator.clipboard.writeText(source);
+      setStatus('copied');
+    } catch {
+      setStatus('failed');
+    }
+  };
+
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre ref={preRef} className={join(PRE_CLASS, className)} {...props}>
```
neha-tanti: **Minor — scroll region not keyboard reachable:** this `overflow-x-auto` `<pre>` has no tab stop, unlike the `table` wrapper in `mdx-components.tsx` which adds `tabIndex={0}` plus a focus ring for exactly this. Wide blocks can't be scrolled without a mouse.

apps/docs/app/_components/CodeBlock.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+    );
+    expect(copy.dataset.copyStatus).toBe('copied');
+
+    act(() => {
+      vi.advanceTimersByTime(2000);
+    });
+
+    expect(screen.getByRole('status').textContent).toBe('');
+    expect(copy.dataset.copyStatus).toBe('idle');
+  });
```
neha-tanti, line 80: **Blocking — missing coverage:** no test for a second click while `copied`. That path is where the announcement and the reset window actually break; add it alongside the absent-`navigator.clipboard` case.

apps/docs/app/_components/ThemePicker.tsx
```diff
@@ -21,10 +21,34 @@ import {
...
+  // Publish the panel's real height so `scroll-pb` clears it and Tab never
+  // parks a control underneath (WCAG 2.4.11). Re-runs per route, so the
+  // reservation drops to 0 wherever the panel does not render.
+  useEffect(() => {
+    const panel = panelRef.current;
+    if (!panel) return;
+
+    const root = document.documentElement;
+    const observer = new ResizeObserver(() => {
+      const clearance = panel.offsetHeight + PANEL_INSET_PX * 2;
```
neha-tanti: **❌ New — blocking:** `bottom-6` is density-scaled via `--nx-spacing-6` (22px compact, 28px relaxed), so the hardcoded 24 goes wrong the moment a user changes this panel's own Size control. Derive it: `const r = panel.getBoundingClientRect(); const clearance = r.height + (innerHeight - r.bottom) * 2;`

apps/docs/app/_components/ThemePicker.tsx
```diff
@@ -35,7 +59,10 @@ export function ThemePicker() {
   };
 
   return (
-    <aside className="nx:fixed nx:bottom-6 nx:right-6 nx:z-popover nx:w-[300px] nx:bg-popover nx:text-popover-foreground nx:border nx:border-border-default nx:rounded-lg nx:shadow-lg">
+    <aside
+      ref={panelRef}
+      className="nx:fixed nx:bottom-6 nx:right-6 nx:z-popover nx:w-[300px] nx:max-h-[calc(100svh-3rem)] nx:overflow-y-auto nx:bg-popover nx:text-popover-foreground nx:border nx:border-border-default nx:rounded-lg nx:shadow-lg"
```
neha-tanti: **⚠️ New:** `3rem` is a second hardcoded copy of `bottom-6`, which density remaps to 22-28px. Use `nx:max-h-[calc(100svh-2*var(--nx-spacing-6))]` so the cap tracks the token instead of restating it.

apps/docs/app/_components/ThemePicker.tsx
```diff
@@ -21,10 +21,34 @@ import {
...
+      const clearance = panel.offsetHeight + PANEL_INSET_PX * 2;
+      root.style.setProperty('--docs-panel-offset', `${clearance}px`);
+    });
+    observer.observe(panel);
+
+    return () => {
+      observer.disconnect();
+      root.style.removeProperty('--docs-panel-offset');
+    };
+  }, [pathname]);
```
neha-tanti: **⚠️ New:** Keying the measurement to routing is a proxy for "does the panel exist". Move the `aside` into its own child component so the early return unmounts it; the effect then belongs to the panel and needs no dep array.

apps/docs/app/globals.css
```diff
@@ -17,4 +17,8 @@
 :root {
   --docs-header-h: 3.5rem;
   --docs-scroll-offset: calc(var(--docs-header-h) + 1.5rem);
+  /* Reserves a band at the bottom of the scrollport so Tab never parks a
+     control under the floating theme panel. ThemePicker overwrites this with
+     the panel's measured height; it stays 0 wherever no panel renders. */
+  --docs-panel-offset: 0px;
```
neha-tanti, line 21: **⚠️ Partial:** The value is real now, but the panel still defaults to expanded, so on a 390x844 viewport it measures ~430px — about half the scrollport reserved. Default `collapsed` to true below the `lg` breakpoint.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre
+        ref={preRef}
+        // A scroll container with no focusable children needs its own tab stop.
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+        tabIndex={0}
+        className={join(PRE_CLASS, className)}
```
neha-tanti: **❌ Still present:** `join` cannot override — Tailwind resolves conflicts by CSS source order, not string order, so a caller's `className` silently loses. Add a `./utils` subpath export to `@nexus_ds/react` mirroring `./appearance`, then use `cn`. No tracked issue owns this.

apps/docs/app/_pages/Typography.tsx
```diff
@@ -181,11 +182,11 @@ export function Typography() {
           </code>{' '}
           in running prose.
         </p>
-        <pre className="nx:typography-code-block nx:rounded-lg nx:border nx:border-border-default nx:bg-muted nx:p-4 nx:overflow-x-auto">
+        <CodeBlock>
```
neha-tanti, line 185: **⚠️ New:** MDX fences hand `CodeBlock` a `<code>` child, which `mdx-components.tsx` documents as the contract. This passes a bare string, so the block renders `<pre>text</pre>` and the `[&_code]` rules are dead. Wrap the template in `<code>` here, in `Responsive.tsx`, and in `MultiBrand.tsx`.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: CHANGES REQUIRED

### Unresolved Issues

| 1 | Hand-rolled `<pre>` blocks bypass the copy button | `_pages/*.tsx` | ✅ Fixed | All three render `CodeBlock`; duplicated class strings gone |
| 2 | Fixed rem cap on a viewport-anchored panel | `ThemePicker.tsx:64` | ✅ Fixed | `30rem` replaced by a viewport-relative cap |
| 3 | `scroll-pb` eats the mobile scrollport | `globals.css:23` | ⚠️ Partial | Measured now, but expanded panel still reserves ~50% at 390x844 |
| 4 | Status expressed twice | `CodeBlock.tsx` | ✅ Fixed | Colour now keys off `data-copy-status` |
| 5 | `join` has no merge semantics | `CodeBlock.tsx:77` | ❌ Still Present | `./utils` subpath export is the answer; no tracked issue cited |

### New Issues

| 1 | ❌ Blocking | `ThemePicker.tsx:42` | `PANEL_INSET_PX = 24` hardcodes a density-scaled token | Derive the inset from `getBoundingClientRect()` |
| 2 | ⚠️ Minor | `ThemePicker.tsx:64` | `calc(100svh-3rem)` is a second hardcoded copy of `bottom-6` | `calc(100svh-2*var(--nx-spacing-6))` |
| 3 | ⚠️ Minor | `ThemePicker.tsx:51` | Effect lifecycle keyed on `pathname`, not the panel's mount | Split the panel into its own child component |
| 4 | ⚠️ Minor | `_pages/Typography.tsx:185` | Three call sites render `<pre>` with no `<code>` child | Wrap the template string in `<code>` |

### Architectural Concerns

**The panel re-themes itself.** `--nx-spacing-6` is remapped to 22px under `[data-density=compact]` and 28px under `relaxed`, and `_lib/appearance-controls.ts:37` wires the panel's own Size control straight to `density`. `PANEL_INSET_PX` stays 24 while the panel moves, reintroducing the drift the measured-height fix was meant to remove.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved.

| 1 | Repeat copy is silent | `CodeBlock.tsx:46` | ✅ Fixed | Mutation-tested: both `key={seq}` and the clearTimeout are load-bearing |
| 2 | Missing repeat-click coverage | `CodeBlock.test.tsx` | ✅ Fixed | 7 tests pass; assertions fail when the fix is reverted |
| 3 | Empty block silent no-op | `CodeBlock.tsx:57` | ✅ Fixed | `textContent` is `''`, branch reports `failed`, `writeText` uncalled |
| 4 | Scroll region not keyboard reachable | `CodeBlock.tsx:76` | ✅ Fixed | `tabIndex={0}` plus canonical focus-ring utilities |

### New Issues

| 1 | ⚠️ Minor | `CodeBlock.tsx:46` | `settle` param permits `'idle'`, scheduling a redundant reset | `next: Exclude<CopyStatus, 'idle'>` |
| 2 | ⚠️ Minor | `CodeBlock.tsx:78` | `{...props}` after `tabIndex={0}` can clobber the tab stop | Move `tabIndex` below the spread, as `className` already is |

### Checked and not flagged

Unmount-only cleanup is correct — `timerRef.current` is read at cleanup time. `ThemePicker`'s observer reading `panel.offsetHeight` is right, not a smell: the default content-box observation makes `borderBoxSize` wrong without `{ box: 'border-box' }`, and writing `scroll-padding-bottom` cannot resize a fixed panel, so there is no feedback loop. `useRef<HTMLElement>(null)` is correct for `<aside>` under React 19, and `[pathname]` is load-bearing since the component returns `null` without unmounting.

Commands run: `vitest run --project=unit CodeBlock.test.tsx` (7/7), `tsc --noEmit` (clean), `eslint --max-warnings 0` (clean).

---

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,97 @@
...
+  // announce a second time.
+  const [{ status, seq }, setCopy] = React.useState({
+    status: 'idle' as CopyStatus,
+    seq: 0,
+  });
+  const { icon, message } = COPY_STATUS[status];
+
+  React.useEffect(() => () => window.clearTimeout(timerRef.current), []);
+
+  const settle = (next: CopyStatus) => {
```
neha-tanti: **⚠️ New:** `settle` accepts `'idle'`, which would schedule a redundant 2s reset back to `'idle'`. Narrow the parameter to the terminal statuses — `next: Exclude<CopyStatus, 'idle'>` — so the type states what the helper is for.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre
+        ref={preRef}
+        // A scroll container with no focusable children needs its own tab stop.
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+        tabIndex={0}
+        className={join(PRE_CLASS, className)}
+        {...props}
```
neha-tanti: **⚠️ New:** `{...props}` spreads after `tabIndex={0}`, so a caller- or plugin-supplied `tabIndex` silently removes the scroll container's tab stop. `className` is already protected by destructuring; move `tabIndex` below the spread to protect it too.

packages/react/package.json
```diff
@@ -17,6 +17,11 @@
       "import": "./dist/index.mjs",
       "require": "./dist/index.js"
     },
+    "./utils": {
```
neha-tanti, line 20: **Minor — dual entry points:** `src/index.ts:9` still exports `cn` from the root barrel, so the same function is now reachable two ways with very different bundle cost and no documented canonical path. Drop the root export and repoint `apps/console/src/lib/rhf-field.tsx:11` at `@nexus_ds/react/utils`.

packages/react/vite.config.ts
```diff
@@ -36,6 +36,7 @@ export default defineConfig({
     lib: {
       entry: {
         index: path.resolve(__dirname, 'src/index.ts'),
+        utils: path.resolve(__dirname, 'src/lib/utils.ts'),
```
neha-tanti, line 39: **Minor — unguarded public entry:** `scripts/verify-export.mjs:176` smoke-imports `index.mjs` and `appearance.mjs`, so the exports map stays honest for those. The new `./utils` subpath has no such assertion, leaving the entry/exports/dts wiring free to drift. Add a `utils.mjs` import asserting `cn`.

apps/docs/app/_components/ThemePicker.tsx
```diff
@@ -163,3 +207,18 @@ function ModeSelect({
     </Select>
   );
 }
+
+/** Subscribes to a `matchMedia` query, staying in sync as the viewport changes. */
+function useMediaQuery(query: string) {
+  const [matches, setMatches] = useState(false);
```
neha-tanti, line 207: **Minor — expanded flash:** `matches` starts `false`, so at 390x844 the first client paint renders the panel expanded and the ResizeObserver publishes the oversized reservation before the effect collapses it. Seed the initial state from `window.matchMedia(query).matches` instead of starting at `false`.

apps/docs/app/layout.tsx
```diff
@@ -50,7 +51,9 @@ export default function RootLayout({
           defaultState={DOCS_APPEARANCE_DEFAULT_STATE}
         >
           <TopNav />
-          <main>{children}</main>
+          <CopyAnnouncerProvider>
```
neha-tanti, line 53: **Minor — provider scope:** `mdx-components.tsx:104` maps every `pre` to `CodeBlock` app-wide, but the provider only covers `<main>`. A fence rendered in nav, footer, or any future non-`main` slot throws from `useCopyAnnouncer`. Wrap the whole body content so the scopes match.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: NEEDS DISCUSSION

### Unresolved Issues

| 1 | One live region per code block | `_components/CodeBlock.tsx` | ✅ Fixed | Hoisted to one `CopyAnnouncerProvider` region in the layout |
| 2 | Panel expanded by default on mobile | `app/globals.css:23` | ✅ Fixed | Defaults collapsed below `lg`; see New Issue 3 |
| 3 | `join` cannot override base utilities | `_components/CodeBlock.tsx` | ✅ Fixed | `cn` via new `./utils` subpath; helper deleted |
| 4 | Hardcoded `PANEL_INSET_PX = 24` | `_components/ThemePicker.tsx` | ✅ Fixed | Measured from `getBoundingClientRect()` |
| 5 | Second hardcoded `bottom-6` copy | `_components/ThemePicker.tsx` | ✅ Fixed | `calc(100svh - 2*var(--nx-spacing-6))`; token verified |
| 6 | Effect keyed on `pathname` as a proxy | `_components/ThemePicker.tsx` | ✅ Fixed | `ThemePanel` split out; effect owns real lifecycle |
| 7 | Bare-string CodeBlock call sites | `_pages/{Typography,Responsive,MultiBrand}.tsx` | ✅ Fixed | All three wrapped in `<code>` |

### New Issues

| 1 | ⚠️ Minor | `packages/react/package.json:20` | `cn` now has two public entry points, neither canonical | Drop `cn` from the root barrel; repoint `apps/console/src/lib/rhf-field.tsx` |
| 2 | ⚠️ Minor | `packages/react/vite.config.ts:39` | New public entry has no guard in the export harness | Add a `utils.mjs` smoke import beside the `appearance.mjs` one |
| 3 | ⚠️ Minor | `_components/ThemePicker.tsx:213` | Initial `false` renders expanded, then collapses on mobile | Seed from `window.matchMedia` in the initializer |
| 4 | ⚠️ Minor | `app/layout.tsx:54` | Provider scope narrower than the global `pre` mapping | Wrap the whole body content, not just `<main>` |

### Architectural Concerns

**Dual entry points for `cn`** (`packages/react/package.json:20`) — the subpath is correctly isolated (built `utils.mjs` pulls no React), but `src/index.ts:9` still exports `cn` from the root barrel, so the same function is reachable two ways with very different bundle cost. Pick one and migrate the existing console caller.

---

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  React.useEffect(() => () => window.clearTimeout(timerRef.current), []);
+
+  const settle = (next: Exclude<CopyStatus, 'idle'>) => {
+    setStatus(next);
+    announce(COPY_STATUS[next].message);
+    window.clearTimeout(timerRef.current);
+    timerRef.current = window.setTimeout(() => {
+      setStatus('idle');
+      announce(COPY_STATUS.idle.message);
```
neha-tanti: **⚠️ New — cross-block clearing:** Any block's 2s reset blanks the now-shared region, wiping a confirmation a *different* block announced moments earlier. Drop this `announce` call — `seq` already forces a re-announce, and stale sr-only text is harmless — then remove the dead `idle.message`.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre
+        className={cn(PRE_CLASS, className)}
+        {...props}
+        // Both below the spread: a caller-supplied value must not replace the
+        // ref the copy control reads, nor the scroll container's own tab stop.
+        ref={preRef}
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+        tabIndex={0}
```
neha-tanti, line 73: **⚠️ New — untested fix:** No test touches the `<pre>`, so deleting `tabIndex={0}` or moving it back above the spread still passes all 8 tests. Add an assertion that the pre carries the tab stop and that a caller-supplied `tabIndex` cannot remove it.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved.

| 1 | `settle` accepted `'idle'` | `CodeBlock.tsx:40` | ✅ Fixed | Narrowed to `Exclude<CopyStatus, 'idle'>`; typecheck clean |
| 2 | `{...props}` clobbers the tab stop | `CodeBlock.tsx:74` | ✅ Fixed | `ref` and `tabIndex` moved below the spread; untested |
| 3 | Repeat copy is silent | `CopyAnnouncer.tsx` | ✅ No regression | `seq` moved into the provider; repeat-copy test still meaningful |
| 4 | Empty block silent no-op | `CodeBlock.test.tsx:146` | ✅ No regression | Branch reports `failed`, `writeText` uncalled |

### New Issues

| 1 | ⚠️ Minor | `_components/CodeBlock.tsx:46` | Any block's idle reset blanks the shared region another block just wrote | Drop the idle `announce`; `seq` already forces re-announce. Remove the dead `idle.message` |
| 2 | ⚠️ Minor | `_components/CodeBlock.tsx:74` | `<pre>` tab stop and spread ordering have no test; deleting them still passes | Assert `tabIndex` on the `pre`, and that a caller `tabIndex` cannot replace it |

### Checked and not flagged

`announce` referential stability plus the unchanged `children` reference means an announcement re-renders only the provider's own JSX, not every `CodeBlock`. `--nx-spacing-6` is a real emitted variable (`packages/tailwind/nexus.css:1237`), so the `max-h` calc resolves. `ResizeObserver` now mounts and unmounts with the panel, and its cleanup falls back to the `0px` declaration in `globals.css:23`. `63.99rem` is the correct choice over media-query range syntax given the Safari 15.4 browser floor. The `./utils` subpath is justified rather than redundant, since `@nexus_ds/react`'s root barrel is client-only and `mdx-components.tsx` is a server module.

Commands run: `vitest run --project=unit CodeBlock.test.tsx` (8/8), `tsc --noEmit` (clean), `eslint --max-warnings 0` (clean).

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: APPROVED

### Unresolved Issues

| 1 | `cn` has two public entry points | `packages/react/src/index.ts` | ✅ Fixed | Root barrel export dropped; `./utils` is canonical |
| 2 | New public entry unguarded in export harness | `scripts/verify-export.mjs:176` | ✅ Fixed | Smoke imports `dist/utils.mjs`, asserts `cn` |
| 3 | Panel renders expanded, then collapses on mobile | `_components/ThemePicker.tsx:32` | ✅ Fixed | Query inverted; unmatched value is collapsed |
| 4 | Provider scope narrower than the `pre` mapping | `app/layout.tsx:53` | ✅ Fixed | Wraps all of `<body>`; nested layout inherits |

### New Issues

None.

### Architectural Concerns

None outstanding. On issue 3, the shipped fix diverges from the suggested one and is the better call: seeding `useState` from `window.matchMedia` would desync the prerendered HTML from the first client render, while inverting the query polarity makes the collapsed state the unmatched default that SSR and hydration both agree on.

---

_Verified: export map has one `cn` entry; `git grep` finds zero root-barrel `cn` importers across all six call sites; `@nexus_ds/docs` and `@nexus_ds/console` typecheck clean; no `global-error.tsx` bypasses the root layout._

apps/docs/app/_components/CopyAnnouncer.tsx
```diff
@@ -0,0 +1,66 @@
...
+const CopyAnnouncerContext = React.createContext<
+  ((message: string) => void) | null
+>(null);
+
+/**
+ * The page's single `role="status"` region. Every CodeBlock publishes its copy
+ * confirmation here rather than mounting a live region of its own, so a page of
+ * twenty fences registers one region with assistive tech instead of twenty.
+ *
+ * The provider owns how long a message lives. A block that published two
```
neha-tanti: **⚠️ New — rationale comment:** This paragraph explains why the timer belongs to the provider rather than describing non-obvious logic. `code-comments.md` routes design rationale to the PR body. The first paragraph is enough here.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const [status, setStatus] = React.useState<CopyStatus>('idle');
+  const announce = useCopyAnnouncer();
+
+  React.useEffect(() => () => window.clearTimeout(timerRef.current), []);
+
+  const settle = (next: Exclude<CopyStatus, 'idle'>) => {
+    setStatus(next);
+    announce(COPY_STATUS[next].message);
+    window.clearTimeout(timerRef.current);
+    // Icon only — the provider owns how long the announcement stays up.
```
neha-tanti: **⚠️ New — rationale comment:** This explains what the timeout deliberately no longer does, which is tradeoff rationale rather than logic. Drop it; the provider's own clear timer is where a reader looks next.

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,91 @@
...
+      settle('failed');
+    }
+  };
+
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre
+        className={cn(PRE_CLASS, className)}
+        {...props}
+        // Both below the spread: the props type omits them, but MDX plugins
```
neha-tanti: **⚠️ New — inaccurate comment:** MDX/rehype can inject `tabIndex` via `hProperties`, but it cannot inject a React `ref`, and the props type omits it. Scope the comment to `tabIndex` so the stated reason matches what the ordering actually guards.

apps/docs/app/_components/ThemePicker.tsx
```diff
@@ -21,39 +21,85 @@ import {
   SelectValue,
 } from './nexus';
 
+/**
+ * Below `lg` an expanded panel would reserve roughly half the scrollport, so it
```
neha-tanti: **⚠️ New — rationale block:** Only the query-polarity point is non-obvious logic; the rest is design rationale for the PR body. Trim to the one sentence explaining why the query asks for the expanded case.

apps/docs/app/_components/CodeBlock.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+    );
+    expect(copy.dataset.copyStatus).toBe('copied');
+    const announcement = screen.getByRole('status').firstElementChild;
+
+    act(() => {
+      vi.advanceTimersByTime(2000);
+    });
+
+    expect(copy.dataset.copyStatus).toBe('idle');
+    // The block restores its own icon; the provider clears the region it owns,
```
neha-tanti: **⚠️ New — redundant comment:** This restates the production rationale rather than explaining the assertion. The `firstElementChild` identity check below is already self-describing next to the text assertion.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved.

| 1 | Idle reset blanks a sibling's announcement | `CopyAnnouncer.tsx` | ✅ Fixed | Message lifetime moved into the provider |
| 2 | `pre` tab stop and spread ordering untested | `CodeBlock.test.tsx` | ✅ Fixed | Both new tests fail when the fix is reverted |

### New Issues

| 1 | ⚠️ Minor | `CopyAnnouncer.tsx:16` | Added JSDoc paragraph is design rationale, not logic | Move to the PR body |
| 2 | ⚠️ Minor | `CodeBlock.tsx:44` | Comment explains what the timer deliberately omits | Move to the PR body |
| 3 | ⚠️ Minor | `ThemePicker.tsx:25` | Six-line block; only query polarity is non-obvious | Keep the polarity sentence, cut the rest |
| 4 | ⚠️ Minor | `CodeBlock.tsx:71` | Comment blames MDX injection for the `ref` ordering | Scope the comment to `tabIndex` |
| 5 | ⚠️ Minor | `CodeBlock.test.tsx:80` | Comment restates production rationale, not the assertion | Drop it |

All five are `code-comments.md` violations — rationale belongs in the PR body. Issue 4 is additionally inaccurate: rehype can inject `tabIndex` via `hProperties`, but it cannot inject a React `ref`.

### Checked and not flagged

`Omit<React.ComponentProps<'pre'>, 'tabIndex' | 'ref'>` — `ref` is a real key on React 19's `ComponentProps<'pre'>`, and the narrowed type still satisfies `MDXComponents['pre']` (contravariant param). Two-timer interleaving (block idle at 2s vs. region clear restarting per announcement) holds for the repeat-copy, cross-block, and failure paths. Blanking in place without advancing `seq` matches `aria-relevant="additions text"`, so the clear is not read out. Unmount cleanup covers both timers. `dist/utils.mjs` exists and the export map resolves.

### Commands run

| Command | Result |
| --- | --- |
| `vitest run --project=unit CodeBlock.test.tsx` | 10/10 pass |
| Same, with each fix reverted (mutation test) | 3 fail — both fixes load-bearing |
| `pnpm turbo run typecheck` | 7/7 pass |
| `eslint --max-warnings 0` (11 changed files) | clean |
| `prettier --check` | clean |

---

apps/docs/app/_components/ThemePicker.tsx
```diff
@@ -21,39 +21,81 @@ import {
   SelectValue,
 } from './nexus';
 
+/**
+ * Asks for the *expanded* case so the unmatched value — the one the prerendered
+ * HTML and the first client render both use — is the collapsed one.
```
neha-tanti, line 27: **⚠️ New — polarity's layout half is gone:** The trimmed doc keeps the hydration reason but drops the layout one: expanding below `lg` would feed `--docs-panel-offset` more than half a 390px scrollport. Record that constraint in the PR body's Focus obstruction section.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: NEEDS DISCUSSION

### Unresolved Issues

All previous issues resolved.

### New Issues

| 1 | ⚠️ Minor | `_components/ThemePicker.tsx:26` | `EXPAND_QUERY`'s tie to `--docs-panel-offset` sizing is now documented nowhere | Add the mobile-reservation constraint to the PR body's Focus obstruction section |

### Architectural Concerns

**Query polarity is a layout invariant, not only a hydration one** (`ThemePicker.tsx:26`) — the surviving sentence explains SSR/first-render agreement, but the removed half explained that an expanded panel below `lg` publishes a `--docs-panel-offset` worth over half a 390px scrollport. Neither the code nor the PR body now carries that constraint, so the two mechanisms in this file read as independent.

---

_Reviewed the `6819fee` diff from git. Comment-only, +4/-16._

apps/docs/app/_components/CodeBlock.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+  };
+
+  return (
+    <div className="nx:relative nx:mb-4">
+      <pre
+        className={cn(PRE_CLASS, className)}
+        {...props}
+        ref={preRef}
+        // Below the spread: a rehype plugin can set `tabIndex` through
```
neha-tanti: **⚠️ Partial — wrong mechanism:** `hProperties` lives on mdast `node.data` and is consumed by remark-rehype; a rehype plugin sets hast `properties` directly. Drop the mechanism: "Below the spread: an MDX plugin can set `tabIndex`, and the scroll container's own tab stop has to win."

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: MINOR CHANGES

### Unresolved Issues

| 1 | JSDoc paragraph is design rationale | `CopyAnnouncer.tsx` | ✅ Fixed | Paragraph removed; region-identity doc kept |
| 2 | Rationale comment above the reset timer | `CodeBlock.tsx` | ✅ Fixed | Removed |
| 3 | Six-line rationale block on the query | `ThemePicker.tsx:24` | ✅ Fixed | Trimmed to the polarity sentence |
| 4 | Prop-ordering comment blames the wrong injection path | `CodeBlock.tsx:71` | ⚠️ Partial | Scoped to `tabIndex`, but `hProperties` is remark-side, not rehype |
| 5 | Test comment restates production rationale | `CodeBlock.test.tsx` | ✅ Fixed | Removed |

### New Issues

None.

### Checked and not flagged

`ref={preRef}` and `tabIndex={0}` both remain below `{...props}`, so the reorder is presentational only — no behaviour change. ESLint reports no unused-disable directive, confirming `// eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex` still sits immediately above the line it disables. Surviving comments in `CopyAnnouncer.tsx` (`seq` remount, in-place blanking) and `ThemePicker.tsx` (query polarity) each describe non-obvious logic and are factually correct — `useMediaQuery` initialises `matches` to `false`, so the unmatched value is the collapsed one on both prerender and first client render.

### Commands run

| Command | Result |
| --- | --- |
| `vitest run --project=unit CodeBlock.test.tsx` | 10/10 pass |
| `pnpm turbo run typecheck` | 7/7 pass |
| `eslint <4 changed files> --max-warnings 0` | clean |

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: NEEDS DISCUSSION

### Unresolved Issues

All previous issues resolved. `EXPAND_QUERY`'s tie to `--docs-panel-offset` sizing is now recorded in the PR body's *Focus obstruction* section, and the query-direction rationale remains in the `ThemePicker.tsx` docblock.

### New Issues

| 1 | ⚠️ Minor | PR body, *Focus obstruction* | "mobile never paints an expanded panel" contradicts `expandedOverride` | Qualify the claim to the default/first-render state; a mobile tap on the toggle does publish the oversized reservation |
| 2 | ⚠️ Minor | PR body, *Focus obstruction* + summary bullet 4 | Reservation described as the panel's height; code adds twice the bottom gap | State the published value as height + 2× the measured bottom gap |

### Architectural Concerns

None. Both findings are description accuracy, not structure.

---

_Verified against `HEAD`: `app/_lib/class-names.ts` absent from the tree and from `git diff main...HEAD`; `./utils` subpath present in `package.json` with matching `vite.config.ts` entry and `verify-export.mjs` assertion; `cn` removed from the root barrel with all five in-repo consumers repointed; `--docs-panel-h` gone, replaced by `--docs-panel-offset`; `CodeBlock.test.tsx` contains exactly 10 `it(...)` blocks._

apps/docs/app/_components/CodeBlock.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+    });
+
+    expect(controls[0]!.dataset.copyStatus).toBe('idle');
+    expect(controls[1]!.dataset.copyStatus).toBe('copied');
+    expect(region.firstElementChild).toBe(announcement);
+    expect(region.textContent).toBe('Code copied to clipboard');
+  });
+
+  it('gives the scroll container a tab stop a caller cannot remove', () => {
+    // The props type omits `tabIndex`, but a rehype plugin can still inject one
```
neha-tanti: **⚠️ New:** The source comment was just reworded to "an MDX plugin" so it names no stage, but this comment on the same guard still says "a rehype plugin". The claim is true, just narrower — align it to "an MDX plugin".

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved.

| 1 | Prop-ordering comment named the wrong plugin stage | `CodeBlock.tsx:71` | ✅ Fixed | Reworded to "an MDX plugin"; stage-agnostic and accurate |

### New Issues

| 1 | ⚠️ Minor | `_components/CodeBlock.test.tsx:231` | Test comment still names a plugin stage the source comment dropped | Use "an MDX plugin" here too |

### Checked and not flagged

`apps/docs/mdx-options.ts` configures `remarkPlugins: ['remark-gfm']` and `rehypePlugins: ['rehype-slug']` through `createMDX` in `next.config.ts`. Both stages can produce a `tabIndex` prop on the emitted `<pre>` — a remark plugin via `node.data.hProperties`, a rehype plugin via hast `properties` — so "an MDX plugin can set `tabIndex`" is true for this pipeline and correctly stage-agnostic. It still names the mechanism the reader needs (prop ordering vs. the spread), so it is not vague to the point of uselessness.

### Commands run

| Command | Result |
| --- | --- |
| `vitest run --project=unit CodeBlock.test.tsx` | 10/10 pass |
| `pnpm turbo run typecheck` | 7/7 pass |
| `eslint CodeBlock.tsx --max-warnings 0` | clean; no unused-disable directive, so the `jsx-a11y` disable still sits directly above `tabIndex={0}` |

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: APPROVED

### Unresolved Issues

| 1 | "Mobile never paints an expanded panel" contradicts `expandedOverride` | PR body, *Focus obstruction* | ✅ Fixed | Claim now scoped to the prerender/first-render default |
| 2 | Reservation described as the panel's height alone | PR body + summary bullet 4 | ✅ Fixed | Now states height + 2× the measured bottom gap |

### New Issues

None.

### Architectural Concerns

None. The description now matches `rect.height + (window.innerHeight - rect.bottom) * 2` in `ThemePicker.tsx:61`.

---

_Verified against `HEAD` `b29d964` (equals the PR head): the only code change since the last review is a one-line test comment in `apps/docs/app/_components/CodeBlock.test.tsx:231` — no module boundary, type, export-map entry, or security surface touched._

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: APPROVED

### Unresolved Issues

| 1 | Test comment named a plugin stage the source dropped | `CodeBlock.test.tsx:231` | ✅ Fixed | Now "an MDX plugin", matching `CodeBlock.tsx:71` |

### New Issues

None.

### Checked and not flagged

The surviving comment explains why the test passes a prop the `Omit<…, 'tabIndex' | 'ref'>` type excludes — non-obvious logic, not design rationale, so it stays within `code-comments.md`. The commit touches one comment line; no behavioural surface changed.

### Commands run

| Command | Result |
| --- | --- |
| `vitest run --project=unit CodeBlock.test.tsx` | 10/10 pass |
| `tsc --noEmit` (`apps/docs`) | clean |
| `eslint CodeBlock.test.tsx --max-warnings 0` | clean |

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Verdict: APPROVED

### Unresolved Issues

All previous issues resolved.

### New Issues

None.

### Architectural Concerns

None. The `--docs-panel-offset` invariant survives the comment removals at all three touch points: `globals.css:20` names the writer, `ThemePicker.tsx:49-50` names the consumer mechanism and the WCAG criterion, and the shared variable name links both to `layout.tsx:34`.

---

_Verified: the deleted `0px`-default sentence carried no safety property — an undefined custom property makes `scroll-padding-bottom` invalid at computed-value time and falls back to `0` regardless. The new `cn` import matches the `@nexus_ds/react/utils` subpath used by all four other docs consumers; `cn` is absent from the root barrel and `scripts/verify-export.mjs` still guards the map._

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Verdict: APPROVED

### Unresolved Issues

All previous issues resolved.

### New Issues

None.

### Checked and not flagged

The chevron `cn()` dedupe was run through the repo's actual merge config (`extendTailwindMerge({ prefix: 'nx' })`, `packages/react/src/lib/utils.ts:115`): both branches emit byte-identical strings to the ternary they replaced, with no drop and no reorder — `transition-transform` and `rotate-180` are different class groups. `expanded` is a `boolean`, so the `&&` short-circuit yields `false` and `clsx` drops it.

The four load-bearing comments all survive and are accurate: the below-the-spread `tabIndex` guard (`CodeBlock.tsx:71`), `seq` remount vs. in-place blanking (`CopyAnnouncer.tsx:17,29`), the `* 2` clearance and `EXPAND_QUERY` polarity (`ThemePicker.tsx:25-28,58`), and the fake-timer offsets (`CodeBlock.test.tsx:117,215`), each re-derived against its test timeline. Removals were rationale-only.

### Commands run

| Command | Result |
| --- | --- |
| `vitest run --project=unit CodeBlock.test.tsx` | 10/10 pass |
| `pnpm turbo run typecheck` | 7/7 successful |
| `eslint` (5 changed files) | clean |

---

PR #691, opened 2026-09-08
fix(docs): publish Storybook and repair the dead footer links

### LINKED ISSUE #678: fix(docs): publish Storybook and repair the dead footer links
## Context

The docs footer links to Figma and to Storybook. Both are `href="#"`.

This looked cosmetic until the documentation plan decided that consumers get **Storybook stories plus component source** instead of dedicated consumer-facing demos. Demos were scoped as docs-site infrastructure only, on the reasoning that anyone adopting Nexus is fine reading the stories.

That reasoning rests on one assumption: that a consumer can actually reach the stories. A `#` link means they cannot. So this is now a distribution gap, not a broken anchor.

## Goal

Nothing in the footer points nowhere, and the stories are genuinely reachable by someone outside the repo.

## Scope

- Publish the static Storybook build to a stable URL and point the footer link at it.
- Wire publishing into CI so the published build tracks `main` rather than whenever someone last ran it by hand.
- Fix the Figma link or remove it. A dead link is worse than no link.
- Check the remaining footer links while in there.

## Acceptance criteria

- No `href="#"` remains in the docs footer.
- The Storybook URL loads and shows the current component set.
- Publishing runs in CI, so the link cannot go stale.

## Notes

Ticket **6 of 7 in Stage 1** of the docs build order. Independent.

If Storybook is **not** going to be published publicly, say so on this issue rather than closing it quietly — the decision to skip consumer-facing demos was justified by story access, and it needs revisiting if that access does not exist.

neha-tanti:
## Summary

- Extracted the inline footer out of `apps/docs/app/layout.tsx` into `apps/docs/app/_components/Footer.tsx`, so the link table is importable and testable.
- **Storybook** link now points at the published build, `https://nexuslabs-ai.github.io/nexus/`.
- **Figma** link removed. I grepped the whole tree — the only `figma.com` references are `https://mcp.figma.com/mcp` MCP server endpoints in `.mcp.json`, i.e. tooling config, not a public design-library URL. There was nothing to point the link at, and per the issue a dead link is worse than no link.
- **GitHub** link was already live; left as-is.
- Added `.github/workflows/deploy-storybook.yml` — builds Storybook and deploys it to GitHub Pages on every push to `main`, so the published catalog tracks `main` rather than whenever someone last built by hand.
- Added `apps/docs/app/_components/Footer.test.ts` asserting every link is an absolute `https://` URL and that Storybook points at the published build.
- Noted the published catalog URL in `README.md`.

## GitHub Issue

Closes #678

## Notes for reviewers

**Why the deploy workflow is standalone rather than a job in `ci.yml`.** It is a post-merge `push: main` workflow, not a PR check, so folding it into `ci.yml` would drag it into the `CI Status` branch-protection aggregator and make Pages availability a merge gate. `ci.yml` already builds Storybook on PRs (`build-storybook`, `ci.yml:450`), which is where a broken build gets caught; this workflow only publishes what `main` already proved buildable. Its steps mirror that job exactly, including the `@nexus_ds/core` pre-build the `AdjustContrast` story needs.

**Pages is created by the workflow, not by hand.** `configure-pages` runs with `enablement: true`, so the first run provisions the site — no manual repo-settings step. The repo is public, so Pages is available. This does mean **the Storybook URL will 404 until this PR merges and that first run completes** — expected, and the reason the URL cannot be smoke-tested from the PR itself.

**Subpath hosting works without a `base` config.** Storybook's Vite production build emits relative asset paths — verified in the local build output, `index.html` and `iframe.html` reference `./assets/…`, `./sb-manager/…`, `./sb-addons/…` — so serving from `/nexus/` resolves correctly. `deploy-pages` also bypasses Jekyll, so no `.nojekyll` is needed.

**Trigger paths** are `packages/**`, `pnpm-lock.yaml`, and the workflow file, plus `workflow_dispatch`. The lockfile is included so a dependency bump that changes rendering still republishes.

**Known pre-existing issue, not touched here:** the Storybook static build also emits `index.d.ts`, `components/`, and `hooks/` into `storybook-static` (the react package's dts plugin running during the build). Harmless — extra files served alongside the catalog — and it predates this PR, affecting the existing `ci.yml` job identically. Flagging rather than fixing, since it is outside this issue's scope; happy to open a separate issue if you want it cleaned up.

## Test Plan

- [x] `pnpm build-storybook` succeeds locally — **899 stories across 70 components**.
- [x] Asset paths in `index.html` / `iframe.html` confirmed relative, so `/nexus/` subpath hosting resolves.
- [x] `Footer.test.ts` passes under the `unit` project (2 tests).
- [x] ESLint clean on all changed files; Prettier clean on all changed files.
- [x] No `href="#"` remains anywhere in `apps/`.
- [ ] After merge: confirm the first `Deploy Storybook` run provisions Pages and `https://nexuslabs-ai.github.io/nexus/` loads the current component set.
- [ ] After merge: confirm the docs footer Storybook link resolves.

---

_Supersedes #690, which GitHub auto-closed when the head branch was renamed from `prasad/...` (a placeholder copied from the branch-naming example in `.claude/rules/github.md`) to `neha-tanti/...`. Same commit, same content._

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -5,6 +5,10 @@ name: Deploy Storybook
 # Standalone from ci.yml: this is a post-merge push:main workflow, not a PR
 # check, so it is not part of the `CI Status` branch-protection aggregator.
 # ci.yml still builds Storybook on PRs, which is where a broken build is caught.
+#
+# One-time repo setup: Settings > Pages > Source: "GitHub Actions". The
+# `actions/configure-pages` `enablement` input cannot do this for us — it
+# requires a token other than GITHUB_TOKEN.
 
 on:
   push:
@@ -12,6 +16,7 @@ on:
     paths:
       - 'packages/**'
       - 'pnpm-lock.yaml'
+      - '.github/actions/build-storybook/action.yml'
       - '.github/workflows/deploy-storybook.yml'
   workflow_dispatch:
 
@@ -28,6 +33,8 @@ jobs:
   deploy:
     name: Build and Deploy
     runs-on: ubuntu-latest
+    # workflow_dispatch accepts any ref; the published site only ever tracks main.
+    if: github.ref == 'refs/heads/main'
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
@@ -35,33 +42,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Install pnpm
-        uses: pnpm/action-setup@v4
-
-      - name: Setup Node
-        uses: actions/setup-node@v4
-        with:
-          node-version-file: '.nvmrc'
-          cache: 'pnpm'
-
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-
-      # The AdjustContrast story imports @nexus_ds/core, which is consumed via
-      # its built dist/runtime entry (package.json exports). Build it so the
-      # Storybook production build can resolve the package.
-      - name: Build @nexus_ds/core
-        run: pnpm --filter @nexus_ds/core build
-
       - name: Build Storybook
-        run: pnpm build-storybook
-
-      # `enablement` creates the Pages site on the first run, so publishing
-      # needs no manual repo-settings step.
-      - name: Configure Pages
-        uses: actions/configure-pages@v5
-        with:
-          enablement: true
+        uses: ./.github/actions/build-storybook
 
       - name: Upload Pages artifact
         uses: actions/upload-pages-artifact@v3
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -16,7 +16,7 @@ on:
     paths:
       - 'packages/**'
       - 'pnpm-lock.yaml'
-      - '.github/actions/build-storybook/action.yml'
+      - '.github/actions/build-storybook/**'
       - '.github/workflows/deploy-storybook.yml'
   workflow_dispatch:
 
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -1,0 +1,73 @@
+name: Deploy Storybook
+
+# Publishes the static Storybook build to GitHub Pages
+# (https://nexuslabs-ai.github.io/nexus/), the URL the docs footer links to.
+# Standalone from ci.yml: this is a post-merge push:main workflow, not a PR
+# check, so it is not part of the `CI Status` branch-protection aggregator.
+# ci.yml still builds Storybook on PRs, which is where a broken build is caught.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/**'
+      - 'pnpm-lock.yaml'
+      - '.github/workflows/deploy-storybook.yml'
+  workflow_dispatch:
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+jobs:
+  deploy:
+    name: Build and Deploy
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      # The AdjustContrast story imports @nexus_ds/core, which is consumed via
+      # its built dist/runtime entry (package.json exports). Build it so the
+      # Storybook production build can resolve the package.
+      - name: Build @nexus_ds/core
+        run: pnpm --filter @nexus_ds/core build
+
+      - name: Build Storybook
+        run: pnpm build-storybook
+
+      # `enablement` creates the Pages site on the first run, so publishing
+      # needs no manual repo-settings step.
+      - name: Configure Pages
+        uses: actions/configure-pages@v5
+        with:
+          enablement: true
+
+      - name: Upload Pages artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: packages/react/storybook-static
+
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
--- a/apps/docs/app/_components/Footer.test.ts
+++ b/apps/docs/app/_components/Footer.test.ts
@@ -1,19 +1,32 @@
+import { createElement } from 'react';
+import { renderToStaticMarkup } from 'react-dom/server';
+
 import { describe, expect, it } from 'vitest';
 
-import { FOOTER_LINKS } from './Footer';
+import { Footer } from './Footer';
 
-describe('FOOTER_LINKS', () => {
-  it('points every link at a reachable absolute URL', () => {
-    expect(FOOTER_LINKS.length).toBeGreaterThan(0);
+function renderFooterAnchors(): { href: string; text: string }[] {
+  const html = renderToStaticMarkup(createElement(Footer));
 
-    for (const link of FOOTER_LINKS) {
-      expect(link.href).toMatch(/^https:\/\//);
+  return [...html.matchAll(/<a\b[^>]*href="([^"]*)"[^>]*>(.*?)<\/a>/g)].map(
+    ([, href = '', text = '']) => ({ href, text })
+  );
+}
+
+describe('Footer', () => {
+  it('renders every link as a reachable absolute URL', () => {
+    const anchors = renderFooterAnchors();
+
+    expect(anchors.length).toBeGreaterThan(0);
+
+    for (const anchor of anchors) {
+      expect(anchor.href).toMatch(/^https:\/\//);
     }
   });
 
-  it('links to the published Storybook', () => {
-    const storybook = FOOTER_LINKS.find((link) => link.label === 'Storybook');
-
-    expect(storybook?.href).toBe('https://nexuslabs-ai.github.io/nexus/');
+  it('tells assistive tech that each link opens a new tab', () => {
+    for (const anchor of renderFooterAnchors()) {
+      expect(anchor.text).toContain('(opens in a new tab)');
+    }
   });
 });
--- a/apps/docs/app/_components/Footer.test.ts
+++ b/apps/docs/app/_components/Footer.test.ts
@@ -1,0 +1,19 @@
+import { describe, expect, it } from 'vitest';
+
+import { FOOTER_LINKS } from './Footer';
+
+describe('FOOTER_LINKS', () => {
+  it('points every link at a reachable absolute URL', () => {
+    expect(FOOTER_LINKS.length).toBeGreaterThan(0);
+
+    for (const link of FOOTER_LINKS) {
+      expect(link.href).toMatch(/^https:\/\//);
+    }
+  });
+
+  it('links to the published Storybook', () => {
+    const storybook = FOOTER_LINKS.find((link) => link.label === 'Storybook');
+
+    expect(storybook?.href).toBe('https://nexuslabs-ai.github.io/nexus/');
+  });
+});
--- a/apps/docs/app/_components/Footer.test.ts
+++ b/apps/docs/app/_components/Footer.test.ts
@@ -5,28 +5,33 @@ import { describe, expect, it } from 'vitest';
 
 import { Footer } from './Footer';
 
-function renderFooterAnchors(): { href: string; text: string }[] {
-  const html = renderToStaticMarkup(createElement(Footer));
+function renderFooterAnchors(): HTMLAnchorElement[] {
+  document.body.innerHTML = renderToStaticMarkup(createElement(Footer));
 
-  return [...html.matchAll(/<a\b[^>]*href="([^"]*)"[^>]*>(.*?)<\/a>/g)].map(
-    ([, href = '', text = '']) => ({ href, text })
-  );
+  const anchors = [...document.body.querySelectorAll('a')];
+
+  expect(anchors.length).toBeGreaterThan(0);
+
+  return anchors;
 }
 
 describe('Footer', () => {
   it('renders every link as a reachable absolute URL', () => {
-    const anchors = renderFooterAnchors();
+    for (const anchor of renderFooterAnchors()) {
+      expect(anchor.getAttribute('href')).toMatch(/^https:\/\//);
+    }
+  });
 
-    expect(anchors.length).toBeGreaterThan(0);
-
-    for (const anchor of anchors) {
-      expect(anchor.href).toMatch(/^https:\/\//);
+  it('opens every link in a new tab without leaking the referrer', () => {
+    for (const anchor of renderFooterAnchors()) {
+      expect(anchor.target).toBe('_blank');
+      expect(anchor.rel).toBe('noreferrer');
     }
   });
 
   it('tells assistive tech that each link opens a new tab', () => {
     for (const anchor of renderFooterAnchors()) {
-      expect(anchor.text).toContain('(opens in a new tab)');
+      expect(anchor.textContent).toContain('(opens in a new tab)');
     }
   });
 });
--- a/apps/docs/app/_components/Footer.tsx
+++ b/apps/docs/app/_components/Footer.tsx
@@ -1,4 +1,6 @@
-export const FOOTER_LINKS = [
+import * as React from 'react';
+
+const FOOTER_LINKS: { label: string; href: `https://${string}` }[] = [
   { label: 'Storybook', href: 'https://nexuslabs-ai.github.io/nexus/' },
   { label: 'GitHub', href: 'https://github.com/nexuslabs-ai/nexus' },
 ];
@@ -20,9 +22,10 @@ export function Footer() {
               href={link.href}
               target="_blank"
               rel="noreferrer"
-              className="nx:rounded-sm nx:hover:text-foreground nx:transition-colors nx:focus-visible:outline-2 nx:focus-visible:outline-focus-default nx:focus-visible:outline-offset-2"
+              className="nx:rounded-sm nx:hover:text-foreground nx:transition-colors nx:focus-visible:outline-2 nx:focus-visible:outline-focus-default nx:focus-visible:outline-offset-(--focus-offset)"
             >
               {link.label}
+              <span className="nx:sr-only"> (opens in a new tab)</span>
             </a>
           ))}
         </nav>
--- a/apps/docs/app/_components/Footer.tsx
+++ b/apps/docs/app/_components/Footer.tsx
@@ -1,0 +1,32 @@
+export const FOOTER_LINKS = [
+  { label: 'Storybook', href: 'https://nexuslabs-ai.github.io/nexus/' },
+  { label: 'GitHub', href: 'https://github.com/nexuslabs-ai/nexus' },
+];
+
+export function Footer() {
+  return (
+    <footer className="nx:border-t nx:border-border-default nx:mt-16">
+      <div className="nx:max-w-[1280px] nx:mx-auto nx:px-6 nx:py-8 nx:flex nx:flex-col nx:gap-4 nx:sm:flex-row nx:sm:items-center nx:sm:justify-between">
+        <div className="nx:flex nx:items-center nx:gap-3">
+          <span className="nx:font-semibold">Nexus</span>
+          <span className="nx:font-mono nx:text-[11px] nx:text-muted-foreground-subtle">
+            MIT · v0.0.1
+          </span>
+        </div>
+        <nav className="nx:flex nx:items-center nx:gap-5 nx:typography-label-default nx:text-muted-foreground">
+          {FOOTER_LINKS.map((link) => (
+            <a
+              key={link.label}
+              href={link.href}
+              target="_blank"
+              rel="noreferrer"
+              className="nx:rounded-sm nx:hover:text-foreground nx:transition-colors nx:focus-visible:outline-2 nx:focus-visible:outline-focus-default nx:focus-visible:outline-offset-2"
+            >
+              {link.label}
+            </a>
+          ))}
+        </nav>
+      </div>
+    </footer>
+  );
+}
--- a/apps/docs/app/_components/Footer.tsx
+++ b/apps/docs/app/_components/Footer.tsx
@@ -1,5 +1,3 @@
-import * as React from 'react';
-
 const FOOTER_LINKS: { label: string; href: `https://${string}` }[] = [
   { label: 'Storybook', href: 'https://nexuslabs-ai.github.io/nexus/' },
   { label: 'GitHub', href: 'https://github.com/nexuslabs-ai/nexus' },
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -49,6 +49,10 @@ export default defineConfig({
       // Unit tests (hooks, utilities) - jsdom
       {
         extends: true,
+        // apps/docs sets `jsx: preserve` for Next, which esbuild reads as the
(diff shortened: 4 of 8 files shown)


.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Build Storybook
+        run: pnpm build-storybook
+
+      # `enablement` creates the Pages site on the first run, so publishing
+      # needs no manual repo-settings step.
+      - name: Configure Pages
+        uses: actions/configure-pages@v5
+        with:
+          enablement: true
```
neha-tanti: **Blocking — `enablement` needs a non-default token:** `configure-pages` documents this input as requiring a token other than `GITHUB_TOKEN`, and none is passed here. The repo reports `has_pages: false`, so the first run fails at this step. Enable Pages in Settings with source `GitHub Actions`, then drop this input.

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,73 @@
...
+# ci.yml still builds Storybook on PRs, which is where a broken build is caught.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/**'
+      - 'pnpm-lock.yaml'
+      - '.github/workflows/deploy-storybook.yml'
+  workflow_dispatch:
```
neha-tanti, line 21: **Minor — deploy surface:** `workflow_dispatch` accepts any ref, so anyone with write access can publish a branch build to the live URL. Confirm the `github-pages` environment exists with a deployment branch policy restricted to `main`.

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,73 @@
...
+        run: pnpm install --frozen-lockfile
+
+      # The AdjustContrast story imports @nexus_ds/core, which is consumed via
+      # its built dist/runtime entry (package.json exports). Build it so the
+      # Storybook production build can resolve the package.
+      - name: Build @nexus_ds/core
+        run: pnpm --filter @nexus_ds/core build
+
+      - name: Build Storybook
+        run: pnpm build-storybook
```
neha-tanti: **Minor — duplicated build:** these steps are identical to `ci.yml`'s `build-storybook`, down to the `@nexus_ds/core` pre-build comment. Extract `.github/actions/build-storybook/action.yml` and call it from both so the pre-build requirement has one definition.

apps/docs/app/_components/Footer.tsx
```diff
@@ -0,0 +1,32 @@
+export const FOOTER_LINKS = [
```
neha-tanti: **Architecture — export boundary:** `TopNav.tsx` keeps its equivalent `NAV_LINKS` unexported. Exporting this table only so a test can read it leaks a component internal, and the Storybook URL now sits in four files with nothing enforcing sync.

**Code quality — invariant not typed:** `href` infers as `string`, but the unconditional `target="_blank"` below now assumes every entry is absolute. Type it as `` href: `https://${string}` `` so a relative link fails at compile time.

neha-tanti (commented):

### Verdict: CHANGES REQUIRED

### Issues

| 1 | ❌ Blocking | `.github/workflows/deploy-storybook.yml:64` | `enablement: true` cannot provision Pages with `GITHUB_TOKEN` | Enable Pages in Settings (source: GitHub Actions), drop the `enablement` input |
| 2 | ⚠️ Minor | `.github/workflows/deploy-storybook.yml:16` | `workflow_dispatch` accepts any ref; branch policy unverified | Confirm the `github-pages` environment limits deployments to `main` |
| 3 | ⚠️ Minor | `.github/workflows/deploy-storybook.yml:57` | Build steps duplicate `ci.yml`'s `build-storybook` verbatim | Extract a composite action used by both |
| 4 | ⚠️ Minor | `apps/docs/app/_components/Footer.tsx:1` | `FOOTER_LINKS` exported only for the test; `NAV_LINKS` is not | Keep the table unexported |

### Architectural Concerns

**Pages provisioning is not self-service** (`.github/workflows/deploy-storybook.yml:64`) — the PR's stated contract is that the first run creates the site, but the action rejects `enablement` under the default token and the repo reports `has_pages: false`. Standalone-vs-integrated is the right call; the provisioning step is the part that does not hold.

**Published URL has four owners** (`apps/docs/app/_components/Footer.tsx:2`) — the same literal appears in the component, the test, `README.md`, and a workflow comment, with no mechanism keeping them aligned. Hold it in one exported constant that the README and workflow reference by name rather than by copy.

### Research Flags

| --- | --- | --- |
| `actions/configure-pages@v5` | `enablement` "requires a token other than `GITHUB_TOKEN`"; the step passes no `token` input | Provision Pages via repo settings and remove `enablement`, or supply a PAT/App token |
| `actions/deploy-pages@v4` | Deployment authorization rides OIDC claims validated against the environment's branch policy | Verify the `github-pages` environment exists with a `main`-only deployment rule |

Sources: [configure-pages action.yml](https://github.com/actions/configure-pages/blob/main/action.yml), [configure-pages#40](https://github.com/actions/configure-pages/issues/40), [Using custom workflows with GitHub Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages)

---

apps/docs/app/_components/Footer.test.ts
```diff
@@ -0,0 +1,19 @@
...
+describe('FOOTER_LINKS', () => {
+  it('points every link at a reachable absolute URL', () => {
+    expect(FOOTER_LINKS.length).toBeGreaterThan(0);
+
+    for (const link of FOOTER_LINKS) {
+      expect(link.href).toMatch(/^https:\/\//);
+    }
+  });
+
+  it('links to the published Storybook', () => {
```
neha-tanti: **Minor — tautological test:** Both cases restate the imported constant instead of exercising `Footer`. The exact-URL assertion only fails when someone edits `FOOTER_LINKS`, and the fix is editing the test. Render with `renderToStaticMarkup` (precedent: `appearance-ssr/page.test.tsx`) and assert the rendered anchor hrefs; then the export isn't needed.

apps/docs/app/_components/Footer.tsx
```diff
@@ -0,0 +1,32 @@
...
+          <span className="nx:font-mono nx:text-[11px] nx:text-muted-foreground-subtle">
+            MIT · v0.0.1
+          </span>
+        </div>
+        <nav className="nx:flex nx:items-center nx:gap-5 nx:typography-label-default nx:text-muted-foreground">
+          {FOOTER_LINKS.map((link) => (
+            <a
+              key={link.label}
+              href={link.href}
+              target="_blank"
```
neha-tanti, line 21: **Minor — new-tab a11y:** Both links open a new tab with no visible or screen-reader cue. Add an `nx:sr-only` span such as `(opens in a new tab)` inside each anchor.

apps/docs/app/_components/Footer.tsx
```diff
@@ -0,0 +1,32 @@
...
+              rel="noreferrer"
+              className="nx:rounded-sm nx:hover:text-foreground nx:transition-colors nx:focus-visible:outline-2 nx:focus-visible:outline-focus-default nx:focus-visible:outline-offset-2"
```
neha-tanti: **Minor — focus offset:** Use `nx:focus-visible:outline-offset-(--focus-offset)`, the tokenised value from `components.md` § Focus States. Siblings `TopNav.tsx:26` and `LeftRail.tsx:10` already use it; this new file hardcodes `outline-offset-2`.

neha-tanti (commented):

### Verdict: MINOR CHANGES

> **Combined verdict across both reviews is CHANGES REQUESTED** (Architect issue #1 is blocking). Posted as `COMMENT` because GitHub rejects `REQUEST_CHANGES` from the PR author and the `nexuslabs-ai-bot` account is not configured on this machine.

### Issues

| 1 | ⚠️ Minor | `apps/docs/app/_components/Footer.test.ts:14` | Both cases restate the imported constant; `Footer` is never rendered | Render via `renderToStaticMarkup` and assert the anchor hrefs |
| 2 | ⚠️ Minor | `apps/docs/app/_components/Footer.tsx:1` | Inferred `href: string` doesn't encode the absolute-URL invariant line 21 assumes | Type as `` href: `https://${string}` `` (inline comment on the Architect review) |
| 3 | ⚠️ Minor | `apps/docs/app/_components/Footer.tsx:23` | `outline-offset-2` instead of the tokenised `(--focus-offset)` | Siblings `TopNav.tsx:26` / `LeftRail.tsx:10` already use the token |
| 4 | ⚠️ Minor | `apps/docs/app/_components/Footer.tsx:21` | Links open a new tab with no screen-reader cue | `nx:sr-only` span per anchor |

### Checked and clear

| Area | Result |
| --- | --- |
| `rel="noreferrer"` without `noopener` | Sufficient — `noreferrer` implies `noopener` per the HTML spec; matches `changelog/page.tsx:143` |
| `.test.ts` as the vehicle | Correct — `testing-react.md`'s "stories are tests" governs `packages/react`; `apps/docs` already ships three `.test.ts(x)` files and the vitest `unit` project globs `apps/**/app/**` |
| `nx:sm:` inside the footer | Not a `responsive.md` violation — the footer is page-shell chrome |
| Workflow YAML syntax, step order, action inputs | Valid; mirrors `ci.yml:455-478` including the `@nexus_ds/core` pre-build |

---

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,55 @@
...
+# `actions/configure-pages` `enablement` input cannot do this for us — it
+# requires a token other than GITHUB_TOKEN.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/**'
+      - 'pnpm-lock.yaml'
+      - '.github/actions/build-storybook/action.yml'
```
neha-tanti: **⚠️ New — path filter pins one file:** A second file added to the action directory would not trigger a redeploy, silently staling the published site. Use `.github/actions/build-storybook/**`, matching `ci.yml`'s `.github/actions/**` filter.

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up)

### Verdict: NEEDS DISCUSSION

### Unresolved Issues

| # | Original Issue | File | Status |
| --- | --- | --- | --- |
| 1 | `enablement: true` cannot provision Pages with `GITHUB_TOKEN` | `deploy-storybook.yml` | ✅ Fixed — step removed entirely |
| 2 | `workflow_dispatch` accepts any ref | `deploy-storybook.yml:37` | ✅ Fixed — ref guard plus the default `github-pages` branch protection |
| 3 | Build steps duplicate `ci.yml`'s `build-storybook` | `.github/actions/build-storybook/action.yml` | ✅ Fixed — composite action consumed by both |
| 4 | `FOOTER_LINKS` exported only for the test | `Footer.tsx` | ✅ Fixed — unexported |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/_components/Footer.tsx:1` | React import is inert in Next, load-bearing only for vitest | Set `esbuild: { jsx: 'automatic' }` on the `unit` project (inline on the SDE2 review) |
| 2 | ⚠️ Minor | `.github/workflows/deploy-storybook.yml:19` | Deploy trigger pins one action file instead of the directory | Use `.github/actions/build-storybook/**` |

### Verification

| Check | Result |
| --- | --- |
| `deploy-pages@v4` without `configure-pages` | No dependency; removing the step is sound |
| Subpath serving at `/nexus/` | Built Storybook emits relative `./assets/…` paths, so no base-path outputs needed |
| `github-pages` environment | GitHub auto-protects it with the default branch; combined with the ref guard, the dispatch hole is closed |
| Composite action authoring + both callers | `using: composite` and `shell:` correct; `ci.yml`'s job behaviourally identical |
| `.github/actions/**` in `ci.yml` filter | Present |

---

apps/docs/app/_components/Footer.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { renderToStaticMarkup } from 'react-dom/server';
+
+import { describe, expect, it } from 'vitest';
+
+import { Footer } from './Footer';
+
+function renderFooterAnchors(): { href: string; text: string }[] {
+  const html = renderToStaticMarkup(createElement(Footer));
+
+  return [...html.matchAll(/<a\b[^>]*href="([^"]*)"[^>]*>(.*?)<\/a>/g)].map(
```
neha-tanti: ⚠️ **New:** The `unit` project runs in jsdom, so a regex HTML parser is unnecessary indirection. Set `document.body.innerHTML` to the markup and use `querySelectorAll('a')` — that drops the regex and the `= ''` defaults, and lets you assert `target` and `rel` too.

apps/docs/app/_components/Footer.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    expect(anchors.length).toBeGreaterThan(0);
+
+    for (const anchor of anchors) {
+      expect(anchor.href).toMatch(/^https:\/\//);
+    }
+  });
+
+  it('tells assistive tech that each link opens a new tab', () => {
+    for (const anchor of renderFooterAnchors()) {
```
neha-tanti, line 33: ⚠️ **New:** This loop passes vacuously if `renderFooterAnchors()` returns nothing — only the first test guards the count. Move `expect(anchors.length).toBeGreaterThan(0)` into the helper so both tests fail for a Footer that renders no links.

apps/docs/app/_components/Footer.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
```
neha-tanti: ⚠️ **New (both lanes):** This import is inert under Next's automatic JSX runtime and exists only because the `unit` vitest project transforms `.tsx` classically. Add `esbuild: { jsx: 'automatic' }` to that project in `vitest.config.ts` and drop the import — verified locally, same two pre-existing suite failures either way.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up)

### Verdict: MINOR CHANGES

### Unresolved Issues

| # | Original Issue | File | Status |
| --- | --- | --- | --- |
| 1 | Tautological test restating the imported constant | `Footer.test.ts` | ✅ Fixed — renders `Footer`, asserts emitted anchors |
| 2 | `href` inferred as bare `string` | `Footer.tsx:3` | ✅ Fixed — template-literal type verified to reject `/docs` |
| 3 | Hardcoded `outline-offset-2` | `Footer.tsx:25` | ✅ Fixed — tokenised `(--focus-offset)` |
| 4 | No screen-reader cue for new-tab links | `Footer.tsx:28` | ✅ Fixed — `nx:sr-only` span |

### New Issues

| 1 | ⚠️ Minor | `Footer.test.ts:11` | Hand-rolled regex HTML parser in a jsdom project | Parse with `document.body.innerHTML` + `querySelectorAll('a')` |
| 2 | ⚠️ Minor | `Footer.test.ts:28` | Second test passes vacuously when zero anchors render | Move the count assertion into `renderFooterAnchors` |
| 3 | ⚠️ Minor | `Footer.tsx:1` | Production import exists only to satisfy the test harness | `esbuild: { jsx: 'automatic' }` on the `unit` project |

### Verification

| Check | Result |
| --- | --- |
| `pnpm vitest run --project=unit .../Footer.test.ts` | 2 passed |
| `pnpm --filter @nexus_ds/docs typecheck` | Only the two pre-existing `cn` errors (stale local `packages/react/dist`) |
| `npx eslint` on both Footer files | Clean |
| Template-literal type constraint | Inserting `{ label: 'Relative', href: '/docs' }` fails with TS2322 — constraint holds |
| React-import claim | Confirmed: removing it fails inside `renderToStaticMarkup`; `esbuild: { jsx: 'automatic' }` makes it unnecessary with no change to the suite's two pre-existing failures |
| YAML parse — `action.yml`, `deploy-storybook.yml`, `ci.yml` | All clean |

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up 2)

### Verdict: NEEDS DISCUSSION

### Unresolved Issues

| # | Original Issue | File | Status |
| --- | --- | --- | --- |
| 1 | Test-only React import in production source | `Footer.tsx` | ✅ Fixed — import removed, JSX transform set at the project |
| 2 | Deploy trigger pinned one action file | `deploy-storybook.yml:19` | ✅ Fixed — `.github/actions/build-storybook/**` |

### New Issues

| 1 | ⚠️ Minor | `vitest.config.ts:55` | `jsxImportSource: 'react'` is esbuild's default; setting it clamps the value workspace-wide | Drop the key; keep only `jsx: 'automatic'` |

### Blast-radius verification

The `unit` project globs `apps/**` and `packages/**/src/**`, so the `esbuild` override was checked against every workspace it reaches:

| Surface | Result |
| --- | --- |
| `packages/react`, `packages/test-utils`, `apps/console` | Already set `jsx: "react-jsx"` in tsconfig — the override is a no-op for them |
| `packages/core` | No JSX |
| `storybook` project | Unaffected — declares no `esbuild`, and `extends: true` pulls from a root config that has none, so it keeps its tsconfig-derived `react-jsx` |
| Scope choice | Project-level is the narrower of the two available scopes; root would have reached the storybook project. A dedicated third project would duplicate `environment`/`globals`/`setupFiles` for a currently-no-op difference |
| Config surface | `esbuild` is correct for the pinned Vite 7.3.5 — `oxc` does not exist in this version; the esbuild→oxc deprecation is the v8 migration |
| Full `unit` suite | 2 failing files, both pre-existing (`audit-agent-drift`, `generate-tailwind-package`); `Footer.test.ts` 3/3 |
| Config comment vs `code-comments.md` | Acceptable — documents non-obvious cross-file behaviour, not a design tradeoff; matches the precedent at `vitest.config.ts:110` |

---

vitest.config.ts
```diff
@@ -49,6 +49,10 @@ export default defineConfig({
       // Unit tests (hooks, utilities) - jsdom
       {
         extends: true,
+        // apps/docs sets `jsx: preserve` for Next, which esbuild reads as the
+        // classic runtime. Components rendered in a unit test would otherwise
+        // need their own `import * as React`.
+        esbuild: { jsx: 'automatic', jsxImportSource: 'react' },
```
neha-tanti: **⚠️ New — redundant key (both lanes):** `jsxImportSource: 'react'` is already esbuild's default — a transform probe emits `react/jsx-runtime` without it, and the docs tests still pass 3/3 with the key removed. Setting it explicitly also clamps the value across every package in the `unit` glob, so a package's own tsconfig `jsxImportSource` could no longer win. Only `jsx: 'automatic'` is load-bearing.

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up 2)

### Verdict: MINOR CHANGES

### Unresolved Issues

| # | Original Issue | File | Status |
| --- | --- | --- | --- |
| 1 | Regex HTML parser in a jsdom project | `Footer.test.ts:9-15` | ✅ Fixed — `document.body.innerHTML` + `querySelectorAll('a')` |
| 2 | Second test passed vacuously on empty render | `Footer.test.ts:13` | ✅ Fixed — count assertion moved into the helper |
| 3 | Test-only React import in production source | `Footer.tsx` | ✅ Fixed — removed; no remaining `React` reference |

### New Issues

| 1 | ⚠️ Minor | `vitest.config.ts:55` | `jsxImportSource: 'react'` is already esbuild's default | Keep `jsx: 'automatic'`; drop the key (inline on the Architect review) |

### Verification

| Check | Result |
| --- | --- |
| `pnpm vitest run --project=unit .../Footer.test.ts` | 3 passed |
| `pnpm --filter @nexus_ds/docs typecheck` | `Footer.tsx` / `Footer.test.ts` clean; only the two stale-`dist` `cn` errors |
| esbuild probe without `jsxImportSource` | Emits `import { jsx } from "react/jsx-runtime"` — default confirmed as `react` |
| Removing `jsxImportSource` | 3/3 still pass. Removing the whole `esbuild` key → all 3 fail. `jsx: 'automatic'` is the load-bearing half |
| Test isolation | Helper *assigns* `innerHTML` rather than appending, and all three tests route through it, so no leakage — despite jsdom being per-file and `setup.ts` doing no DOM reset |
| `expect()` inside the helper | Sound — vitest attributes failure to the enclosing `it`, stack points at `Footer.test.ts:13`; it guards the fixture, not the assertion under test |
| Type inference | `querySelectorAll('a')` hits lib.dom's `HTMLElementTagNameMap` overload → `HTMLAnchorElement[]`; `.target`/`.rel` typecheck legitimately, not by widening |
| `getAttribute('href')` vs `.href` | Correct — `.href` would resolve against the document base URL; `.target`/`.rel` are plain reflected attributes |

---

neha-tanti (commented):
## 🤖 Automated Code Review — Principal Architect (Follow-up 3)

### Verdict: APPROVED

### Unresolved Issues

| # | Original Issue | File | Status |
| --- | --- | --- | --- |
| 1 | `jsxImportSource: 'react'` redundant and clamps workspace-wide | `vitest.config.ts:55` | ✅ Fixed |

### New Issues

_None._

### Verification

| Check | Result |
| --- | --- |
| Clamp mechanism | Verified against installed Vite 7.3.5 (`chunks/config.js:6100-6106`): `transformWithEsbuild` nulls the file's tsconfig field whenever the matching esbuild option is set. Dropping the key restores per-package tsconfig resolution |
| Behaviour today | No package in the `unit` glob sets a non-`react` `jsxImportSource`, so the change is behaviour-neutral now and removes the future clamp |
| Surviving `jsx: 'automatic'` | Still nulls tsconfig `jsx` by the same mechanism — the intended override for `apps/docs`' `jsx: preserve`; the only other value in the glob (`react-jsx`) is equivalent to automatic |
| Comment at `vitest.config.ts:52-54` | Describes only the `jsx: 'automatic'` override, never mentioned `jsxImportSource` — not stale, and compliant with `code-comments.md` |
| Issue #678 acceptance criteria | All three met in code: no `href="#"` in `apps/docs/app`, footer points at the Pages URL, publishing runs on `push: main` |
| One-time Pages enablement | Correctly left as a repo setting — `configure-pages` `enablement` requires a token above `GITHUB_TOKEN`, so no code path is owed |
| Full `unit` suite | 38 passed / 2 failed files — the two known pre-existing failures |

---

neha-tanti (commented):
## 🤖 Automated Code Review — SDE2 (Follow-up 3)

### Verdict: APPROVED

> Combined verdict across both lanes is **APPROVE**. Posted as `COMMENT` because GitHub rejects `APPROVE` from the PR author and the `nexuslabs-ai-bot` account is not configured on this machine.

### Unresolved Issues

| # | Original Issue | File | Status |
| --- | --- | --- | --- |
| 1 | `jsxImportSource: 'react'` is already esbuild's default | `vitest.config.ts:55` | ✅ Fixed |

### New Issues

_None._

### Verification

| Check | Result |
| --- | --- |
| Transform after the change | `transformWithEsbuild` against a path inside `apps/docs` still emits `import { jsx } from "react/jsx-runtime"` with the key gone |
| Competing config | No `jsxImportSource` set in any tsconfig or build config repo-wide, so esbuild's `react` default is the only source |
| Config still load-bearing | Removing the whole `esbuild` line fails all 3 tests with `ReferenceError: React is not defined`; restored, tree clean |
| `Footer.test.ts` | 3 passed |
| `pnpm --filter @nexus_ds/docs typecheck` | Neither Footer file errors; only the two pre-existing stale-`dist` `cn` errors |
| ESLint on both Footer files | Clean |
| Comment at `vitest.config.ts:52-54` | Both claims verified empirically; documents toolchain mechanism, not design rationale — compliant |

### Final sweep — `Footer.tsx` / `Footer.test.ts`

| Aspect | Result |
| --- | --- |
| Dead-href class of bug | Unrepresentable — the `https://${string}` template-literal type rejects it at compile time |
| `target="_blank"` safety | `rel="noreferrer"` implies `noopener` per spec |
| Accessible name | `sr-only` span sits inside the anchor, so it reads "Storybook (opens in a new tab)" |
| Focus contract | `nx:outline-offset-(--focus-offset)` matches `components.md`; the old inline copy used the untokenised `outline-offset-2` |
| `ripple-effect.md` | The `isExternal` ternary was deleted rather than carried over |
| `nx:font-mono nx:text-[11px]` | Byte-identical to the existing docs convention at `apps/docs/app/page.tsx:176` — consistent, not a new magic number |
| `nx:sm:` usage | Correct for an app-level page-shell footer under `responsive.md`; not a component-internal `@container` case |

---

PR #693, opened 2026-09-09
fix(tooling): chrome-devtools MCP hardcodes a macOS browser path

### LINKED ISSUE #684: fix(tooling): chrome-devtools MCP hardcodes a macOS browser path

neha-tanti:
## Summary

- `.mcp.json` now resolves the `chrome-devtools-mcp` executable path from `$NEXUS_BROWSER_PATH`, with the current macOS Brave path as the `${VAR:-default}` fallback — so macOS contributors need no change, and everyone else sets one env var.
- The override lives in `.claude/settings.local.json`, which is already gitignored (`.gitignore:45`), so no contributor edits a checked-in file.
- Documented as a new "Browser MCP (chrome-devtools)" section in `CONTRIBUTING.md`, directly after the docs-MCP section.

## GitHub Issue

Closes #684

## Test Plan

- [x] `.mcp.json` still parses; the arg reads `--executable-path=${NEXUS_BROWSER_PATH:-/Applications/Brave Browser.app/Contents/MacOS/Brave Browser}`
- [x] `npx prettier --check .mcp.json CONTRIBUTING.md` passes
- [x] With `NEXUS_BROWSER_PATH` set in `.claude/settings.local.json` on Windows, the variable reaches the tool environment and the target binary exists
- [ ] macOS contributor with Brave installed and no `NEXUS_BROWSER_PATH` set: browser tools still launch (default branch of the expansion)

### DIFF
--- a/.codex/config.toml
+++ b/.codex/config.toml
@@ -1,9 +1,6 @@
 [mcp_servers.chrome-devtools]
 command = "npx"
-args = [
-    "chrome-devtools-mcp@latest",
-    "--executable-path=/Applications/Brave Browser.app/Contents/MacOS/Brave Browser",
-]
+args = ["chrome-devtools-mcp@latest"]
 
 [mcp_servers.figma]
 url = "https://mcp.figma.com/mcp"
--- a/.mcp.json
+++ b/.mcp.json
@@ -8,7 +8,7 @@
       "command": "npx",
       "args": [
         "chrome-devtools-mcp@latest",
-        "--executable-path=/Applications/Brave Browser.app/Contents/MacOS/Brave Browser"
+        "--executable-path=${NEXUS_BROWSER_PATH:-/Applications/Brave Browser.app/Contents/MacOS/Brave Browser}"
       ]
     },
     "nexus-docs-mcp": {
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -347,6 +347,8 @@ args = [
 
 Both servers stay registered and expose the same tool set, so the choice is per-session: point the agent at the `chrome-devtools-local` tools. `--isolated` is what lets it run at all — without it both servers launch into `~/.cache/chrome-devtools-mcp/chrome-profile` and whichever starts second fails with `The browser is already running for …`.
 
+That shared profile is also why only one agent drives the browser at a time: neither `.mcp.json` nor `.codex/config.toml` passes `--isolated`, so Claude Code and Codex cannot hold a browser open at once. If you need both, add `--isolated` to whichever you reach for second.
+
 ---
 
 ## Releasing
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -319,11 +319,26 @@ It launches the browser at `$NEXUS_BROWSER_PATH`, defaulting to Brave's macOS lo
 }
 ```
 
-`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`; Ubuntu snap installs land at `/snap/bin/chromium`.
+`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
 
-A wrong or missing path does not fail at startup — the server connects, then the first browser tool call returns `Browser was not found at the configured executablePath (...)`. Run `claude mcp list` to confirm the value expanded.
+A wrong or missing path does not fail at startup — the server connects, and the first browser tool call returns `Browser was not found at the configured executablePath (...)`. That call is the only real check: `claude mcp list` reports the server healthy either way, and prints the `${…}` template rather than the value it expanded to.
 
-Codex reads `.codex/config.toml`, which has no environment-variable expansion, so it passes no `--executable-path` at all and `chrome-devtools-mcp` falls back to your system Chrome on every platform.
+### Codex
+
+Codex reads `.codex/config.toml`, which has no environment-variable expansion ([openai/codex#2680](https://github.com/openai/codex/issues/2680)), so there is no equivalent of `$NEXUS_BROWSER_PATH` there. It passes no `--executable-path` at all and `chrome-devtools-mcp` falls back to your system Chrome on every platform.
+
+The two configs therefore default to different browsers — Brave under Claude Code on macOS, Chrome under Codex — and that is deliberate: pinning Brave is what keeps macOS contributors' Claude Code setup unchanged, and Codex cannot express the same default without re-introducing the hardcoded path this section exists to remove.
+
+To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml` and use that one — a distinct name avoids depending on how Codex ranks project config against user config:
+
+```toml
+[mcp_servers.chrome-devtools-local]
+command = "npx"
+args = [
+    "chrome-devtools-mcp@latest",
+    "--executable-path=/usr/bin/brave-browser",
+]
+```
 
 ---
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -305,6 +305,24 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 
 ---
 
+## Browser MCP (chrome-devtools)
+
+`.mcp.json` also wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots, [`ui-audit`](.agents/skills/ui-audit-guide/SKILL.md), and checking a change in the running app.
+
+It launches the browser at `$NEXUS_BROWSER_PATH`, which defaults to Brave's macOS location — so on macOS there is nothing to set. Elsewhere, point it at your own Chromium-family binary in `.claude/settings.local.json`. That file is gitignored, so `.mcp.json` stays untouched:
+
+```json
+{
+  "env": {
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
+  }
+}
+```
+
+Restart Claude Code after editing it. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
+
+---
+
 ## Releasing
 
 Releases are driven by [changesets](https://github.com/changesets/changesets) and the [`Release`](.github/workflows/release.yml) workflow. Only the two runtime packages publish to npm; everything else is either internal or copy/own.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -307,6 +307,8 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 
 ## Browser MCP (chrome-devtools)
 
+### Claude Code
+
 `.mcp.json` wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots and for checking a change in the running app.
 
 It launches the browser at `$NEXUS_BROWSER_PATH`, defaulting to Brave's macOS location — so a macOS contributor with Brave installed there sets nothing. Everyone else (and macOS contributors who use a different browser) points it at their own binary by adding an `env` block to `~/.claude/settings.json`, alongside whatever keys are already in the file:
@@ -329,7 +331,9 @@ Codex reads `.codex/config.toml`, which has no environment-variable expansion ([
 
 The two configs therefore default to different browsers — Brave under Claude Code on macOS, Chrome under Codex — and that is deliberate: pinning Brave is what keeps macOS contributors' Claude Code setup unchanged, and Codex cannot express the same default without re-introducing the hardcoded path this section exists to remove.
 
-To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml` and use that one — a distinct name avoids depending on how Codex ranks project config against user config:
+This is a change for macOS: `.codex/config.toml` used to pin Brave, so Codex now drives system Chrome instead — and nothing at all if Chrome is not installed. Use the override below to put it back on Brave.
+
+To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml` — a distinct name avoids depending on how Codex ranks project config against user config:
 
 ```toml
 [mcp_servers.chrome-devtools-local]
@@ -338,6 +342,13 @@ args = [
     "chrome-devtools-mcp@latest",
     "--executable-path=/usr/bin/brave-browser",
 ]
+```
+
+Both servers then load and expose the same tool set, so picking the right one is a per-session choice rather than something the config settles. To take the project server out of the running, park it with the documented `enabled` key — subject to the same untested precedence question as everything else in this section:
+
+```toml
+[mcp_servers.chrome-devtools]
+enabled = false
 ```
 
 ---
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -333,7 +333,7 @@ The two configs therefore default to different browsers — Brave under Claude C
 
 This is a change for macOS: `.codex/config.toml` used to pin Brave, so Codex now drives system Chrome instead — and nothing at all if Chrome is not installed. Use the override below to put it back on Brave.
 
-To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml` — a distinct name avoids depending on how Codex ranks project config against user config:
+To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml`. Codex ranks the project's `.codex/config.toml` above your user config, so redefining `chrome-devtools` there would not take effect — a distinct name sidesteps that ordering entirely:
 
 ```toml
 [mcp_servers.chrome-devtools-local]
@@ -341,15 +341,11 @@ command = "npx"
 args = [
     "chrome-devtools-mcp@latest",
     "--executable-path=/usr/bin/brave-browser",
+    "--isolated",
 ]
 ```
 
-Both servers then load and expose the same tool set, so picking the right one is a per-session choice rather than something the config settles. To take the project server out of the running, park it with the documented `enabled` key — subject to the same untested precedence question as everything else in this section:
-
-```toml
-[mcp_servers.chrome-devtools]
-enabled = false
-```
+Both servers stay registered and expose the same tool set, so the choice is per-session: point the agent at the `chrome-devtools-local` tools. `--isolated` is what lets it run at all — without it both servers launch into `~/.cache/chrome-devtools-mcp/chrome-profile` and whichever starts second fails with `The browser is already running for …`.
 
 ---
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -347,7 +347,7 @@ args = [
 
 Both servers stay registered and expose the same tool set, so the choice is per-session: point the agent at the `chrome-devtools-local` tools. `--isolated` is what lets it run at all — without it both servers launch into `~/.cache/chrome-devtools-mcp/chrome-profile` and whichever starts second fails with `The browser is already running for …`.
 
-That shared profile is also why only one agent drives the browser at a time: neither `.mcp.json` nor `.codex/config.toml` passes `--isolated`, so Claude Code and Codex cannot hold a browser open at once. If you need both, add `--isolated` to whichever you reach for second.
+That shared profile is also why only one agent drives the browser at a time: neither `.mcp.json` nor `.codex/config.toml` passes `--isolated`, so Claude Code and Codex cannot hold a browser open at once. To run both, give one of them an isolated server of its own — the Codex block above, or on the Claude Code side `claude mcp add -s local chrome-devtools-local -- npx chrome-devtools-mcp@latest --isolated`, which registers it for you alone rather than in the tracked `.mcp.json`.
 
 ---
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -307,11 +307,11 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 
 ## Browser MCP (chrome-devtools)
 
+`chrome-devtools-mcp` drives a real Chromium-family browser for screenshots and for checking a change in the running app.
+
 ### Claude Code
 
-`.mcp.json` wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots and for checking a change in the running app.
-
-It launches the browser at `$NEXUS_BROWSER_PATH`, defaulting to Brave's macOS location — so a macOS contributor with Brave installed there sets nothing. Everyone else (and macOS contributors who use a different browser) points it at their own binary by adding an `env` block to `~/.claude/settings.json`, alongside whatever keys are already in the file:
+`.mcp.json` launches the browser at `$NEXUS_BROWSER_PATH`, defaulting to Brave's macOS location — so a macOS contributor with Brave installed there sets nothing. Everyone else adds an `env` block to `~/.claude/settings.json`, alongside whatever keys are already in the file:
 
 ```json
 {
@@ -321,19 +321,15 @@ It launches the browser at `$NEXUS_BROWSER_PATH`, defaulting to Brave's macOS lo
 }
 ```
 
-`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
+Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`. That file is per-machine, so every checkout picks it up; use the gitignored `.claude/settings.local.json` to override a single worktree. Restart Claude Code after editing either.
 
-A wrong or missing path does not fail at startup — the server connects, and the first browser tool call returns `Browser was not found at the configured executablePath (...)`. That call is the only real check: `claude mcp list` reports the server healthy either way, and prints the `${…}` template rather than the value it expanded to.
+A wrong path fails lazily — the server connects and `claude mcp list` reports it healthy, but the first browser tool call returns `Browser was not found at the configured executablePath (...)`. That call is the only real check.
 
 ### Codex
 
-Codex reads `.codex/config.toml`, which has no environment-variable expansion ([openai/codex#2680](https://github.com/openai/codex/issues/2680)), so there is no equivalent of `$NEXUS_BROWSER_PATH` there. It passes no `--executable-path` at all and `chrome-devtools-mcp` falls back to your system Chrome on every platform.
+`.codex/config.toml` has no environment-variable expansion ([openai/codex#2680](https://github.com/openai/codex/issues/2680)), so it passes no `--executable-path` and falls back to system Chrome on every platform. It used to pin Brave, so macOS Codex users now get Chrome — or nothing, if Chrome is not installed.
 
-The two configs therefore default to different browsers — Brave under Claude Code on macOS, Chrome under Codex — and that is deliberate: pinning Brave is what keeps macOS contributors' Claude Code setup unchanged, and Codex cannot express the same default without re-introducing the hardcoded path this section exists to remove.
-
-This is a change for macOS: `.codex/config.toml` used to pin Brave, so Codex now drives system Chrome instead — and nothing at all if Chrome is not installed. Use the override below to put it back on Brave.
-
-To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml`. Codex ranks the project's `.codex/config.toml` above your user config, so redefining `chrome-devtools` there would not take effect — a distinct name sidesteps that ordering entirely:
+To pick your own browser, add a second server under its own name to `~/.codex/config.toml`. Codex ranks the project's `.codex/config.toml` above your user config, so redefining `chrome-devtools` there would not take effect:
 
 ```toml
 [mcp_servers.chrome-devtools-local]
@@ -345,9 +341,7 @@ args = [
 ]
 ```
 
-Both servers stay registered and expose the same tool set, so the choice is per-session: point the agent at the `chrome-devtools-local` tools. `--isolated` is what lets it run at all — without it both servers launch into `~/.cache/chrome-devtools-mcp/chrome-profile` and whichever starts second fails with `The browser is already running for …`.
-
-That shared profile is also why only one agent drives the browser at a time: neither `.mcp.json` nor `.codex/config.toml` passes `--isolated`, so Claude Code and Codex cannot hold a browser open at once. To run both, give one of them an isolated server of its own — the Codex block above, or on the Claude Code side `claude mcp add -s local chrome-devtools-local -- npx chrome-devtools-mcp@latest --isolated`, which registers it for you alone rather than in the tracked `.mcp.json`.
+Both servers stay registered with the same tools, so point the agent at the `chrome-devtools-local` ones. `--isolated` is what lets it run: without it every server launches into `~/.cache/chrome-devtools-mcp/chrome-profile` and whichever starts second fails with `The browser is already running for …`. That shared profile is also why Claude Code and Codex cannot hold a browser open at once — to run both, give one an isolated server of its own, either the block above or `claude mcp add -s local chrome-devtools-local -- npx chrome-devtools-mcp@latest --isolated`, which stays out of the tracked `.mcp.json`.
 
 ---
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -321,7 +321,7 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 }
 ```
 
-Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`. That file is per-machine, so every checkout picks it up; use the gitignored `.claude/settings.local.json` to override a single worktree. Restart Claude Code after editing either.
+`~/.claude/settings.json` is per-machine, so every checkout picks it up; use the gitignored `.claude/settings.local.json` to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
 
 A wrong path fails lazily — the server connects and `claude mcp list` reports it healthy, but the first browser tool call returns `Browser was not found at the configured executablePath (...)`. That call is the only real check.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -307,9 +307,9 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 
 ## Browser MCP (chrome-devtools)
 
-`.mcp.json` also wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots, [`ui-audit`](.agents/skills/ui-audit-guide/SKILL.md), and checking a change in the running app.
+`.mcp.json` wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots and for checking a change in the running app.
 
-It launches the browser at `$NEXUS_BROWSER_PATH`, which defaults to Brave's macOS location — so on macOS there is nothing to set. Elsewhere, point it at your own Chromium-family binary in `.claude/settings.local.json`. That file is gitignored, so `.mcp.json` stays untouched:
+It launches the browser at `$NEXUS_BROWSER_PATH`, defaulting to Brave's macOS location — so a macOS contributor with Brave installed there sets nothing. Everyone else (and macOS contributors who use a different browser) points it at their own binary by adding an `env` block to `~/.claude/settings.json`, alongside whatever keys are already in the file:
 
 ```json
 {
@@ -319,7 +319,11 @@ It launches the browser at `$NEXUS_BROWSER_PATH`, which defaults to Brave's macO
 }
 ```
 
-Restart Claude Code after editing it. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
+`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`; Ubuntu snap installs land at `/snap/bin/chromium`.
+
+A wrong or missing path does not fail at startup — the server connects, then the first browser tool call returns `Browser was not found at the configured executablePath (...)`. Run `claude mcp list` to confirm the value expanded.
+
+Codex reads `.codex/config.toml`, which has no environment-variable expansion, so it passes no `--executable-path` at all and `chrome-devtools-mcp` falls back to your system Chrome on every platform.
 
 ---
 


CONTRIBUTING.md
```diff
@@ -305,6 +305,24 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
 
 ---
 
+## Browser MCP (chrome-devtools)
+
+`.mcp.json` also wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots, [`ui-audit`](.agents/skills/ui-audit-guide/SKILL.md), and checking a change in the running app.
```
neha-tanti: **Minor — wrong attribution:** `.agents/skills/ui-audit-guide/SKILL.md:30` says the skill uses Playwright MCP, and explicitly notes this repo ships chrome-devtools *instead*. Citing `ui-audit` as a chrome-devtools consumer contradicts it. Drop the link.

CONTRIBUTING.md
```diff
@@ -305,6 +305,24 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+It launches the browser at `$NEXUS_BROWSER_PATH`, which defaults to Brave's macOS location — so on macOS there is nothing to set. Elsewhere, point it at your own Chromium-family binary in `.claude/settings.local.json`. That file is gitignored, so `.mcp.json` stays untouched:
```
neha-tanti: **Minor — scope layer:** a browser path is per-machine, not per-checkout. `.claude/settings.local.json` is project-scoped, so every worktree needs its own copy. Document `~/.claude/settings.json` `env` as the home instead.

.mcp.json
```diff
@@ -8,7 +8,7 @@
       "command": "npx",
       "args": [
         "chrome-devtools-mcp@latest",
-        "--executable-path=/Applications/Brave Browser.app/Contents/MacOS/Brave Browser"
+        "--executable-path=${NEXUS_BROWSER_PATH:-/Applications/Brave Browser.app/Contents/MacOS/Brave Browser}"
```
neha-tanti, line 11: **Discussion:** `.codex/config.toml:5` holds this same literal path and is git-tracked, so Codex users on Linux/Windows are still broken. #684 isn't closed until both configs are fixed.

neha-tanti (commented):

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Configuration boundaries | ⚠️ | Override lives in a project-scoped file; the value is per-machine |
| Cross-platform contract | ⚠️ | Default is still one OS + one browser baked into shared config |
| Issue coverage | ❌ | `.codex/config.toml` carries the identical bug, unfixed |

### Verdict: CHANGES REQUESTED

### Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `.codex/config.toml:5` | Identical hardcoded macOS Brave path, tracked in git, left unfixed |
| 2 | ⚠️ Minor | `CONTRIBUTING.md:312` | Machine-scoped value documented in a project-scoped settings file |
| 3 | ⚠️ Minor | `CONTRIBUTING.md:310` | `ui-audit` cited as a chrome-devtools consumer; the skill uses Playwright MCP |
| 4 | 💬 Discussion | `.mcp.json:11` | Literal default vs. letting the server auto-detect |

### Architectural Concerns

**#684 is only half fixed** (`.codex/config.toml:5`) — the repo ships two committed MCP configs declaring the same three servers, and Codex users on Linux/Windows still hit the exact failure the issue describes. Apply the equivalent override there in this PR.

**Override scope** (`CONTRIBUTING.md:312`) — a browser binary path is a property of the machine, not the checkout, so `~/.claude/settings.json` `env` is the correct home; `.claude/settings.local.json` forces a fresh copy per worktree.

**Default contract** (`.mcp.json:11`) — `${VAR:-default}` cannot express "omit the flag", so the shared default is permanently one OS and one browser. Omitting `--executable-path` would let Puppeteer resolve system Chrome per-OS, but that switches Mac contributors from Brave to Chrome, which #684 explicitly rules out — so the literal default is defensible as long as the Codex config is fixed too.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,24 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+```json
+{
+  "env": {
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
```
neha-tanti, line 319: **Blocking — this example does not work.** Reproduced end-to-end: the space in `Program Files` splits the arg and the server reports `Browser was not found at the configured executablePath (C:/Program)`. Windows forces `shell: true` (Node won't spawn `npx.cmd` otherwise), and Node doesn't re-quote args. Both `C:/PROGRA~1/...` and embedding quotes in the value work — verified.

CONTRIBUTING.md
```diff
@@ -305,6 +305,24 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
 ---
 
+## Browser MCP (chrome-devtools)
+
+`.mcp.json` also wires up `chrome-devtools-mcp`, which drives a real Chromium-family browser for screenshots, [`ui-audit`](.agents/skills/ui-audit-guide/SKILL.md), and checking a change in the running app.
+
+It launches the browser at `$NEXUS_BROWSER_PATH`, which defaults to Brave's macOS location — so on macOS there is nothing to set. Elsewhere, point it at your own Chromium-family binary in `.claude/settings.local.json`. That file is gitignored, so `.mcp.json` stays untouched:
+
+```json
+{
```
neha-tanti, line 317: **Minor — snippet clobbers keys:** shown as a whole file, so pasting it wipes existing ones (this repo's own `.claude/settings.local.json` already has `outputStyle`). Reword to "add an `env` block" and show it alongside another key.

CONTRIBUTING.md
```diff
@@ -305,6 +305,24 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+  "env": {
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
+  }
+}
+```
+
+Restart Claude Code after editing it. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
```
neha-tanti: **Minor — no failure symptom.** A wrong/unset path doesn't fail at startup; the server connects and errors on the first browser tool. Add that symptom and `claude mcp list` as the check. Also scope "nothing to set" to Brave users — macOS contributors on Chrome hit the same failure.

neha-tanti (commented):

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Mechanism correctness | ✅ | `${VAR:-default}` in `args` and settings `env` → expansion both verified |
| Documented example | ❌ | The Windows example is broken; reproduced end-to-end |
| Doc accuracy | ⚠️ | Snippet clobbers existing keys; macOS claim overstated; no failure symptom |

### Verdict: CHANGES REQUESTED

### Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `CONTRIBUTING.md:317` | Space in the Windows path truncates the arg; browser never launches |
| 2 | ⚠️ Minor | `CONTRIBUTING.md:315` | Snippet is a whole file — pasting it drops existing keys such as `outputStyle` |
| 3 | ⚠️ Minor | `CONTRIBUTING.md:322` | Failure is lazy, not at startup; symptom undocumented, macOS claim overstated |

### Verification performed

Drove `chrome-devtools-mcp@latest` over stdio (initialize → `list_pages`) with `shell: true`, which Windows forces because Node refuses to spawn `npx.cmd` unshelled (`EINVAL`):

| Path form | Result |
| --- | --- |
| `C:/Program Files/BraveSoftware/.../brave.exe` | ❌ `Browser was not found at the configured executablePath (C:/Program)` |
| `C:/PROGRA~1/BraveSoftware/.../brave.exe` | ✅ browser launched, pages returned |
| `"C:/Program Files/BraveSoftware/.../brave.exe"` (quotes inside the value) | ✅ browser launched, pages returned |

The macOS default is unaffected despite also containing spaces — on macOS `npx` is a real binary, so it spawns without a shell and no splitting occurs. This is a Windows-only break, and it hits the one example the doc gives.

Confirmed working as claimed: `${VAR:-default}` expansion in `.mcp.json` `args`; `env` from `.claude/settings.local.json` reaching the MCP subprocess (`$NEXUS_BROWSER_PATH` is live in this session); `.gitignore:45` ignoring `settings.local.json`; the `.agents/skills/ui-audit-guide/SKILL.md` relative link resolving from repo root; the three Linux paths.

---

.mcp.json
neha-tanti, line 11: **❌ New — divergent defaults:** this pins Brave-on-macOS while `.codex/config.toml` now auto-detects Chrome, so one repo drives two different browsers for the same work. Either unify on `${NEXUS_BROWSER_PATH:-}` (and amend #684's scope sentence), or document the split as deliberate.

.codex/config.toml
```diff
@@ -1,9 +1,6 @@
 [mcp_servers.chrome-devtools]
 command = "npx"
-args = [
-    "chrome-devtools-mcp@latest",
-    "--executable-path=/Applications/Brave Browser.app/Contents/MacOS/Brave Browser",
-]
+args = ["chrome-devtools-mcp@latest"]
```
neha-tanti, line 3: **⚠️ Partial — no Codex override:** dropping the flag fixes the macOS pin, but a contributor with Brave/Chromium and no system Chrome still has no per-machine escape hatch. A user-level `[mcp_servers.chrome-devtools-local]` block is the workaround; precedence is undocumented and untested here.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Codex config unpinned | ⚠️ | Path removed, but no override path remains for Codex users |
| Cross-tool consistency | ❌ | The two committed configs now default to different browsers |
| Issue coverage | ⚠️ | AC met for Claude Code; partially met for Codex |

### Verdict: CHANGES REQUESTED

### Unresolved Issues

| 1 | Hardcoded macOS Brave path in the Codex config | `.codex/config.toml:3` | ⚠️ Partial | Path removed; no way to select a non-Chrome binary |
| 2 | Machine-scoped value in a project-scoped file | `CONTRIBUTING.md:322` | ✅ Fixed | Now documents `~/.claude/settings.json` |
| 3 | `ui-audit` cited as a chrome-devtools consumer | `CONTRIBUTING.md:310` | ✅ Fixed | Link dropped |
| 4 | Literal default vs. auto-detection | `.mcp.json:11` | 🔄 Changed | Resolved for Codex only; configs now diverge |

### New Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `.mcp.json:11` | The two committed configs encode different default browsers |
| 2 | ⚠️ Minor | `.codex/config.toml:3` | Codex users without system Chrome must edit a tracked file |

### Architectural Concerns

**Divergent default browser** (`.mcp.json:11`) — Codex cannot express the Brave-on-macOS default, so the repo now drives Brave under one agent and Chrome under the other for the same screenshot work. Either unify both on auto-detection and amend #684's scope sentence, or accept the split explicitly in the docs as a known consequence of Codex's missing expansion.

**No Codex override path** (`.codex/config.toml:3`) — a contributor with Brave or Chromium but no system Chrome has no per-machine escape hatch under Codex, which is the shape of failure #684 exists to remove. A user-level `[mcp_servers.chrome-devtools-local]` block is the available workaround; note that Codex's project-vs-user precedence is not covered by official docs and was not testable here (codex CLI is not installed on this machine).

### Research Flags

| --- | --- | --- |
| Codex config | `${VAR}` expansion is an open request (openai/codex#2680), not a shipped feature | Do not document expansion for `.codex/config.toml` |
| `chrome-devtools-mcp` | `--config <json>` exists on main but not in 1.9.0 | Not a usable symmetric override mechanism today |

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,28 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+{
+  "env": {
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
+  }
+}
+```
+
+`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`; Ubuntu snap installs land at `/snap/bin/chromium`.
+
+A wrong or missing path does not fail at startup — the server connects, then the first browser tool call returns `Browser was not found at the configured executablePath (...)`. Run `claude mcp list` to confirm the value expanded.
```
neha-tanti: **❌ New — this check cannot check.** Ran `claude mcp list`: it prints the raw `${NEXUS_BROWSER_PATH:-...}` template, not the expanded value, and per the preceding sentence the server reports healthy even with a bad path. Drop it, or point at the first browser tool call as the real verification.

CONTRIBUTING.md
```diff
@@ -305,6 +305,28 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+```json
+{
+  "env": {
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
+  }
+}
+```
+
+`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`; Ubuntu snap installs land at `/snap/bin/chromium`.
```
neha-tanti: **⚠️ New — doubtful path:** snapd's `home` interface denies hidden dirs in `$HOME`, and `chrome-devtools-mcp` defaults its profile to `~/.cache/chrome-devtools-mcp`. `/snap/bin/chromium` was added unverified and could not be tested from Windows. Drop it or note the caveat.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issues | ✅ | All three resolved; the blocking one was a retracted false positive |
| Doc accuracy | ❌ | The verification step doesn't verify; one Linux path is doubtful |

### Verdict: CHANGES REQUESTED

### Unresolved Issues

| 1 | Spaced Windows path truncates the arg | `CONTRIBUTING.md:317` | ✅ Retracted | False positive; MCP SDK uses `cross-spawn` |
| 2 | Snippet clobbers existing settings keys | `CONTRIBUTING.md:312` | ✅ Fixed | Prose now says to add `env` alongside existing keys |
| 3 | Lazy failure undocumented; macOS claim overstated | `CONTRIBUTING.md:324` | ✅ Fixed | Symptom documented; claim scoped to Brave |

### New Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `CONTRIBUTING.md:324` | `claude mcp list` prints the raw `${...}` template, so it cannot confirm expansion |
| 2 | ⚠️ Minor | `CONTRIBUTING.md:322` | `/snap/bin/chromium` may be blocked by snap confinement; unverified |

### Verification performed

Ran `claude mcp list` in this repo. It printed the arg **unexpanded**:

```
chrome-devtools: npx chrome-devtools-mcp@latest --executable-path=${NEXUS_BROWSER_PATH:-/Applications/Brave Browser.app/Contents/MacOS/Brave Browser}
```

So the sentence telling contributors to use it as the check is wrong — and per the sentence immediately before it, the server reports healthy even with a bad path. The first browser tool call is the only real check.

Confirmed accurate: the error string `Browser was not found at the configured executablePath (...)` is verbatim from `chrome-devtools-mcp@1.9.0`; the lazy timing is right (`#getContext()` is reached only via `ToolHandler`); `~/.claude/settings.json` `env` is real and per-machine; the retraction of the spaced-path finding is correct (`@modelcontextprotocol/sdk` depends on `cross-spawn ^7.0.5`); the Codex sentence matches what `.codex/config.toml` now does; the three `/usr/bin/*` Linux paths are standard.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,43 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+A wrong or missing path does not fail at startup — the server connects, and the first browser tool call returns `Browser was not found at the configured executablePath (...)`. That call is the only real check: `claude mcp list` reports the server healthy either way, and prints the `${…}` template rather than the value it expanded to.
+
+### Codex
+
+Codex reads `.codex/config.toml`, which has no environment-variable expansion ([openai/codex#2680](https://github.com/openai/codex/issues/2680)), so there is no equivalent of `$NEXUS_BROWSER_PATH` there. It passes no `--executable-path` at all and `chrome-devtools-mcp` falls back to your system Chrome on every platform.
+
+The two configs therefore default to different browsers — Brave under Claude Code on macOS, Chrome under Codex — and that is deliberate: pinning Brave is what keeps macOS contributors' Claude Code setup unchanged, and Codex cannot express the same default without re-introducing the hardcoded path this section exists to remove.
```
neha-tanti: **❌ New — macOS Codex regression unstated:** `.codex/config.toml` pinned Brave before this PR and now resolves system Chrome, so a macOS Codex contributor changes browser silently — or gets nothing if Chrome isn't installed. Say so here and point at the escape hatch below.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Codex escape hatch | ✅ | Separately-named server is sound given untestable precedence |
| Split documentation | ⚠️ | Honest about Claude Code; silent on the Codex-side consequence |
| Issue coverage | ⚠️ | AC "macOS contributors need no change" now holds only for Claude Code |

### Verdict: CHANGES REQUESTED

### Unresolved Issues

| 1 | Two configs encode different default browsers | `CONTRIBUTING.md:330` | ⚠️ Partial | Split documented; macOS Codex Brave→Chrome change omitted |
| 2 | Codex users have no per-machine escape hatch | `.codex/config.toml:3` | ✅ Fixed | Separately-named server documented |

### New Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `CONTRIBUTING.md:330` | The macOS Codex behaviour change is a real regression and goes unstated |

### Architectural Concerns

**macOS Codex regression is undocumented** (`CONTRIBUTING.md:330`) — before this PR `.codex/config.toml` pinned Brave; it now resolves system Chrome, so a macOS Codex contributor silently changes browser and gets nothing at all if Chrome is absent. The paragraph scopes its "unchanged" claim to Claude Code, which is accurate but leaves the reader to infer macOS is unaffected everywhere.

This is a consequence of the owner's decision to document the split rather than unify the defaults, not a challenge to it — the fix is a sentence naming the Codex-side change and pointing at the escape hatch below, not a config change.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,43 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml` and use that one — a distinct name avoids depending on how Codex ranks project config against user config:
```
neha-tanti: **⚠️ New — additive, not a replacement:** adding `chrome-devtools-local` doesn't retire the repo's `chrome-devtools` server; both load and expose identical tool sets, so "use that one" is a per-session human choice. Mention the documented `enabled = false` key, with the same precedence caveat this section already carries.

CONTRIBUTING.md
```diff
@@ -305,6 +305,43 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
+  }
+}
+```
+
+`~/.claude/settings.json` is per-machine, which is what a browser path is — set it once and every checkout picks it up. Use the project-local `.claude/settings.local.json` (gitignored) only to override a single worktree. Restart Claude Code after editing either. Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`.
+
+A wrong or missing path does not fail at startup — the server connects, and the first browser tool call returns `Browser was not found at the configured executablePath (...)`. That call is the only real check: `claude mcp list` reports the server healthy either way, and prints the `${…}` template rather than the value it expanded to.
+
+### Codex
```
neha-tanti, line 328: **⚠️ New — heading asymmetry:** lines 310-324 are Claude Code-specific but unlabeled, so this `### Codex` heading makes the first half read as section preamble rather than its counterpart. Add a matching `### Claude Code` above line 310.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issues | ✅ | Both resolved |
| Factual accuracy | ✅ | Every claim in the new text verified |
| Completeness | ⚠️ | Escape hatch is additive; heading levels asymmetric |

### Verdict: MINOR CHANGES

### Unresolved Issues

| 1 | `claude mcp list` documented as the expansion check | `CONTRIBUTING.md:324` | ✅ Fixed | Now states it prints the template and reports healthy regardless |
| 2 | Unverified `/snap/bin/chromium` path | `CONTRIBUTING.md:322` | ✅ Fixed | Removed |

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `CONTRIBUTING.md:332` | "use that one" is not self-executing — the project server stays registered |
| 2 | ⚠️ Minor | `CONTRIBUTING.md:326` | `### Codex` added, but the Claude Code half above has no matching heading |

### Verification performed

Every factual claim in the new text checks out: the TOML snippet parses, and `[mcp_servers.NAME]` with `command` / `args` matches Codex's documented STDIO schema; openai/codex#2680 is real, open, and is specifically about `${ENV_VAR}` expansion in Codex MCP config; the line-330 description of the split matches what `.mcp.json` and `.codex/config.toml` actually do; prettier passes.

One mechanism worth adding: `enabled` **is** a documented `[mcp_servers.NAME]` key (`enabled = false` parks a server without deleting it), so it is the natural companion to the escape hatch. Note that whether a user-level `enabled = false` beats the project entry runs into the same project-vs-user precedence question the section already hedges on, and that could not be tested here — so document it as the mechanism, not as a guaranteed override.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,54 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+args = [
+    "chrome-devtools-mcp@latest",
+    "--executable-path=/usr/bin/brave-browser",
+]
+```
+
+Both servers then load and expose the same tool set, so picking the right one is a per-session choice rather than something the config settles. To take the project server out of the running, park it with the documented `enabled` key — subject to the same untested precedence question as everything else in this section:
+
+```toml
+[mcp_servers.chrome-devtools]
```
neha-tanti: **⚠️ New — no file anchor:** the snippet above says where to put `chrome-devtools-local`; this one says nothing. It must be `~/.codex/config.toml` — dropped into the repo's tracked `.codex/config.toml` it disables the browser MCP for every contributor. Name the file in the sentence above.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| macOS Codex regression | ✅ | Now stated explicitly and pointed at the override |
| Escape-hatch soundness | ❌ | Disable mechanism depends on the precedence the section avoids |
| Issue coverage | ⚠️ | AC holds for Claude Code; Codex restore path unproven |

### Verdict: CHANGES REQUESTED

### Unresolved Issues

| 1 | macOS Codex behaviour change unstated | `CONTRIBUTING.md:334` | ✅ Fixed | Names the Brave→Chrome change and the absent-Chrome case |

### New Issues

| --- | --- | --- | --- |
| 1 | ❌ Blocking | `CONTRIBUTING.md:347` | `enabled = false` is ranked below the tracked project entry, so the documented disable may be inert |
| 2 | ⚠️ Minor | `CONTRIBUTING.md:350` | The snippet names no owning file; applied to the tracked config it disables the server for everyone |

### Architectural Concerns

**The disable step reintroduces the dependency the escape hatch was built to avoid** (`CONTRIBUTING.md:347`) — `chrome-devtools-local` uses a distinct name precisely so the override never rides on project-vs-user ranking, and `enabled = false` puts it back one paragraph later against an entry Codex ranks higher. Since line 334's macOS behaviour change is only acceptable because the override restores Brave, an override that may not apply leaves #684's "macOS contributors need no change" unmet for Codex.

**Codex's own ordering is published** — project `.codex/config.toml` outranks `~/.codex/config.toml`, closest-to-cwd first, so the section's "untested precedence question" understates what is knowable and overstates what is safe. Either confirm the merge behaviour and state the result, or drop the `enabled = false` block and leave the distinct-name server as the sole documented path.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,54 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+### Codex
+
+Codex reads `.codex/config.toml`, which has no environment-variable expansion ([openai/codex#2680](https://github.com/openai/codex/issues/2680)), so there is no equivalent of `$NEXUS_BROWSER_PATH` there. It passes no `--executable-path` at all and `chrome-devtools-mcp` falls back to your system Chrome on every platform.
+
+The two configs therefore default to different browsers — Brave under Claude Code on macOS, Chrome under Codex — and that is deliberate: pinning Brave is what keeps macOS contributors' Claude Code setup unchanged, and Codex cannot express the same default without re-introducing the hardcoded path this section exists to remove.
+
+This is a change for macOS: `.codex/config.toml` used to pin Brave, so Codex now drives system Chrome instead — and nothing at all if Chrome is not installed. Use the override below to put it back on Brave.
+
+To drive a different browser under Codex, add a second server under its own name to `~/.codex/config.toml` — a distinct name avoids depending on how Codex ranks project config against user config:
```
neha-tanti: **⚠️ New — precedence is documented:** Codex ranks project `.codex/config.toml` above `~/.codex/config.toml`, closest-to-cwd first. The distinct-name advice still holds, but drop "untested" — the only genuinely open question is whether table merging is per-key or whole-table.

CONTRIBUTING.md
```diff
@@ -305,6 +305,54 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+```toml
+[mcp_servers.chrome-devtools-local]
+command = "npx"
+args = [
+    "chrome-devtools-mcp@latest",
+    "--executable-path=/usr/bin/brave-browser",
+]
+```
+
+Both servers then load and expose the same tool set, so picking the right one is a per-session choice rather than something the config settles. To take the project server out of the running, park it with the documented `enabled` key — subject to the same untested precedence question as everything else in this section:
```
neha-tanti: **⚠️ New — not a free choice:** both servers default to `~/.cache/chrome-devtools-mcp/chrome-profile`, so whichever launches first throws `The browser is already running... Use --isolated` at the other. Add `--isolated` to the `chrome-devtools-local` args; that also makes disabling the project server load-bearing rather than optional.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issues | ✅ | Both resolved |
| Factual accuracy | ⚠️ | `enabled` key and Brave history verified; precedence claim overstated |
| Completeness | ⚠️ | Two servers cannot both run — shared profile dir |

### Verdict: MINOR CHANGES

### Unresolved Issues

| 1 | Escape hatch additive, not a replacement | `CONTRIBUTING.md:347` | ✅ Fixed | Both servers named; `enabled` key documented |
| 2 | Heading asymmetry under Browser MCP | `CONTRIBUTING.md:310` | ✅ Fixed | `### Claude Code` added |

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `CONTRIBUTING.md:347` | Not a per-session choice — both servers share one profile dir, so the second fails to launch |
| 2 | ⚠️ Minor | `CONTRIBUTING.md:336` | Codex publishes this ranking; "untested" overstates the unknown |

### Verification performed

Checked against git history, Codex's published config docs, and the installed `chrome-devtools-mcp@1.9.0` source:

| Claim | Result |
| --- | --- |
| "`.codex/config.toml` used to pin Brave" | ✅ `ae86e62` removed `--executable-path=/Applications/Brave Browser.app/...` |
| "nothing at all if Chrome is not installed" | ✅ `browser.js:146-150` sets `channel: 'chrome'` — resolves an installed Chrome, never downloads |
| "the documented `enabled` key" | ✅ Codex docs: "`enabled` (optional): Set `false` to disable a server without deleting it." |
| Both TOML snippets parse | ✅ |
| `npx prettier --check CONTRIBUTING.md` | ✅ |

Not testable here: the `codex` CLI is not installed on this machine, so precedence and `enabled` handling rest on published docs, not observed behaviour.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,50 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+[mcp_servers.chrome-devtools-local]
+command = "npx"
+args = [
+    "chrome-devtools-mcp@latest",
+    "--executable-path=/usr/bin/brave-browser",
+    "--isolated",
+]
+```
+
+Both servers stay registered and expose the same tool set, so the choice is per-session: point the agent at the `chrome-devtools-local` tools. `--isolated` is what lets it run at all — without it both servers launch into `~/.cache/chrome-devtools-mcp/chrome-profile` and whichever starts second fails with `The browser is already running for …`.
```
neha-tanti: **⚠️ New — collision is broader than the hatch:** `.mcp.json` and `.codex/config.toml` both launch into that same default profile dir, so running Claude Code and Codex browser tools concurrently fails the same way. Add `--isolated` to one tracked config, or state the constraint here.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Escape-hatch soundness | ✅ | Unreliable disable removed; distinct name is now the sole path |
| Config-file boundaries | ⚠️ | Shared profile dir also collides across the two tracked configs |
| Issue coverage | ✅ | #684 met for both agents |

### Verdict: NEEDS DISCUSSION

### Unresolved Issues

| 1 | `enabled = false` ranked below the project entry | `CONTRIBUTING.md:336` | ✅ Fixed | Block removed; precedence now stated as the reason |
| 2 | Disable snippet named no owning file | — | ✅ Fixed | Snippet gone |

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `CONTRIBUTING.md:348` | The same profile collision applies between the two tracked configs, which the section does not mention |

### Architectural Concerns

**The collision is not confined to the escape hatch** (`CONTRIBUTING.md:348`) — `.mcp.json` and `.codex/config.toml` both launch `chrome-devtools-mcp` with no `--isolated` and no `--user-data-dir`, so a contributor driving browser tools from Claude Code and Codex in the same window hits the identical failure the new paragraph describes. Either add `--isolated` to one tracked config or state the one-at-a-time constraint here.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issues | ✅ | Both resolved |
| Factual accuracy | ✅ | Every changed claim verified against the installed package and Codex docs |

### Verdict: APPROVED

### Unresolved Issues

| 1 | Shared profile dir blocks the second server | `CONTRIBUTING.md:344` | ✅ Fixed | `--isolated` added; collision now explained at 348 |
| 2 | "Untested" overstated the precedence unknown | `CONTRIBUTING.md:336` | ✅ Fixed | Documented ordering stated as the reason for the distinct name |

### Verification performed

| Claim | Result |
| --- | --- |
| `--isolated` is a real `chrome-devtools-mcp@1.9.0` flag | ✅ boolean, no default (`config/browser-options.js`) |
| `--isolated` + `--executable-path` do not conflict | ✅ ran both against the bin; control case `--isolated --user-data-dir` correctly errored as mutually exclusive |
| `--isolated` skips the shared profile dir | ✅ `browser.js:123-131` gates the `~/.cache/chrome-devtools-mcp/chrome-profile` assignment on `!isolated && !userDataDir` |
| Error string is verbatim | ✅ `browser.js:191` |
| Codex precedence claim | ✅ project config outranks user config, per-key merge; the project entry sets `command` + `args`, so "would not take effect" holds |
| TOML parses with the added arg | ✅ trailing comma is valid TOML |
| `npx prettier --check CONTRIBUTING.md` | ✅ |

Not testable here: the `codex` CLI is not installed, so precedence rests on OpenAI's published config docs rather than observed behaviour. Those docs add one caveat the prose omits — the project layer is skipped for untrusted projects — which does not change the advice, since a distinct server name works under either trust state.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Config-file boundaries | ✅ | Both tracked configs stay machine-agnostic |
| Concurrency constraint | ✅ | Documented rather than pushed into a tracked config |
| Issue coverage | ✅ | #684 met for both agents |

### Verdict: APPROVED

### Unresolved Issues

| 1 | Profile collision also spans the two tracked configs | `CONTRIBUTING.md:350` | ✅ Fixed | Constraint stated; remedy already shown above |

_No new issues._

Every per-machine value still lives in a user-scope file (`~/.claude/settings.json`, `~/.codex/config.toml`); the concurrency workaround does not push a machine-specific flag into either tracked config.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,52 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+That shared profile is also why only one agent drives the browser at a time: neither `.mcp.json` nor `.codex/config.toml` passes `--isolated`, so Claude Code and Codex cannot hold a browser open at once. If you need both, add `--isolated` to whichever you reach for second.
```
neha-tanti: **⚠️ New — one-sided remedy:** for Codex the `chrome-devtools-local` recipe above already carries `--isolated`. For Claude Code the only in-repo spot is the tracked `.mcp.json`, which line 324 steers contributors away from. Point at a local-scope server instead: `claude mcp add -s local chrome-devtools-local -- npx chrome-devtools-mcp@latest --isolated`.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issues | ✅ | All resolved |
| Factual accuracy | ✅ | Both claims in the new sentence verified in source |
| Actionability | ⚠️ | The remedy has no Claude Code-side location |

### Verdict: MINOR CHANGES

### Unresolved Issues

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `CONTRIBUTING.md:350` | "whichever you reach for second" names no Claude Code-side place to put `--isolated` |

### Verification performed

| Claim | Result |
| --- | --- |
| Neither tracked config passes `--isolated` or `--user-data-dir` | ✅ `.mcp.json` passes only `--executable-path=${NEXUS_BROWSER_PATH:-…}`; `.codex/config.toml` only `chrome-devtools-mcp@latest` |
| "cannot hold a browser open at once" | ✅ failure is at concurrent launch — `userDataDir` is `mkdir`-ed inside `launch()` (`browser.js:128-133`), reached only via `ToolHandler` → `ensureBrowserLaunched`, so a merely-registered server locks nothing |
| Error path | ✅ `browser.js:191` throws `The browser is already running for ${userDataDir}.` |
| No contradiction with lines 347-348 or the divergent-defaults paragraph | ✅ |
| `npx prettier --check CONTRIBUTING.md` | ✅ |
| `claude mcp add -s local\|user` is real | ✅ confirmed via `claude mcp add --help` |

The `codex` CLI is not installed here, so nothing about Codex's runtime config merge was re-tested this round.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Config-file ownership | ✅ | `claude mcp add -s local` writes `~/.claude.json`, not the tracked `.mcp.json` |
| Issue coverage | ✅ | #684 criteria unchanged and still met |

### Verdict: APPROVED

### Unresolved Issues

_No new issues._

The remedy now matches the ownership split the rest of the section uses — `~/.claude/settings.json` for `NEXUS_BROWSER_PATH`, `~/.codex/config.toml` for the Codex isolated server, `~/.claude.json` for the local Claude Code one. `.mcp.json` and `.codex/config.toml` are byte-identical across `71b4f5d..96f698d`.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issue | ✅ | Claude Code-side remedy now named |
| Factual accuracy | ✅ | CLI signature, scope value, and scope description all verified |

### Verdict: APPROVED

### Unresolved Issues

| 1 | Remedy named no Claude Code-side location | `CONTRIBUTING.md:350` | ✅ Fixed | `claude mcp add -s local` recipe added |

_No new issues._

### Verification performed

| Claim | Result |
| --- | --- |
| `claude mcp add` signature | ✅ `[options] <name> <commandOrUrl> [args...]`; the `--` separator form matches the CLI's own stdio example (v2.1.229) |
| `-s local` is a real scope | ✅ `-s, --scope` accepts `local \| user \| project` |
| "registers it for you alone rather than in the tracked `.mcp.json`" | ✅ local scope stores under `~/.claude.json` per-project and stays private to you; `project` scope is the one that writes `.mcp.json` |
| `npx prettier --check CONTRIBUTING.md` | ✅ |

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Information loss | ✅ | Every operational fact survives the trim |
| Config ownership | ✅ | Split still conveyed across the four scopes |
| Issue coverage | ✅ | #684 unaffected; both configs untouched |

### Verdict: APPROVED

### Unresolved Issues

_No new issues._

Facts checked as surviving: the env-block example, Linux path list, restart requirement, the lazy-failure symptom with its verbatim error string, `claude mcp list` being an invalid check, the `openai/codex#2680` link, the macOS Codex behaviour change including the absent-Chrome case, the distinct-server-name requirement with its precedence reason, `--isolated` with the `The browser is already running for …` string, and both remedies for the one-agent-at-a-time constraint.

Dropping "or missing" from the failure sentence is a correctness improvement rather than a loss — `${NEXUS_BROWSER_PATH:-…}` resolves an unset variable to the Brave default, so a wrong path is the only reachable failure.

---

CONTRIBUTING.md
```diff
@@ -305,6 +305,46 @@ The rule that mandates querying it lives in [`.claude/rules/docs-mcp.md`](.claud
...
+
+```json
+{
+  "env": {
+    "NEXUS_BROWSER_PATH": "C:/Program Files/BraveSoftware/Brave-Browser/Application/brave.exe"
+  }
+}
+```
+
+Typical Linux paths: `/usr/bin/brave-browser`, `/usr/bin/google-chrome`, `/usr/bin/chromium`. That file is per-machine, so every checkout picks it up; use the gitignored `.claude/settings.local.json` to override a single worktree. Restart Claude Code after editing either.
```
neha-tanti: **⚠️ New — dangling referent:** "That file" now directly follows three browser binary paths, and its antecedent (`~/.claude/settings.json`) is a paragraph and a JSON block back. Name the file explicitly.

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issues | ✅ | All resolved |
| Rewrite accuracy | ✅ | No claim became false in the compression |
| Referent clarity | ⚠️ | One pronoun lost its antecedent to the reflow |

### Verdict: MINOR CHANGES

### Unresolved Issues

### New Issues

| --- | --- | --- | --- |
| 1 | ⚠️ Minor | `CONTRIBUTING.md:324` | "That file" now follows three browser binary paths, with its antecedent a paragraph and a JSON block back |

### Verification performed

| Line | Claim | Result |
| --- | --- | --- |
| 314 | "`.mcp.json` launches the browser at `$NEXUS_BROWSER_PATH`" | ✅ `${NEXUS_BROWSER_PATH:-…Brave Browser}` matches "defaulting to Brave's macOS location" |
| 324 | "That file is per-machine…" | ⚠️ referent degraded — see finding; the rest is intact |
| 326 | Lazy failure | ✅ the dropped template-printing detail was diagnostic colour, not a premise |
| 330 | Codex paragraph merge | ✅ `main:.codex/config.toml` did pin the Brave macOS path; the current config passes no `--executable-path`; the "so" chain introduces no false causality |
| 344 | "every server launches into `~/.cache/…`" | ✅ generalises correctly from "both servers"; the stated conclusion and reader action are unchanged |
| — | `npx prettier --check CONTRIBUTING.md` | ✅ |

Nothing removed was load-bearing for correctness — the cuts were rationale.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — Principal Architect

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Config ownership | ✅ | Unchanged; the edit only resolves an antecedent |
| Issue coverage | ✅ | #684 holds; no config file moved in `0dfc86e` |

### Verdict: APPROVED

### Unresolved Issues

_No new issues._

`~/.claude/settings.json` remains the per-machine layer and `.claude/settings.local.json` the gitignored per-worktree override, exactly as before. `$NEXUS_BROWSER_PATH` in `.mcp.json` and no `--executable-path` in `.codex/config.toml` are untouched.

---

neha-tanti (commented):
## 🤖 Automated Follow-up Review — SDE2

### Summary

| Category | Status | Notes |
| --- | --- | --- |
| Previous issue | ✅ | Referent named explicitly |
| Reordering | ✅ | No referent broken by the move |

### Verdict: APPROVED

### Unresolved Issues

| 1 | "That file" had no nearby antecedent | `CONTRIBUTING.md:324` | ✅ Fixed | Sentence now opens with the file name |

_No new issues._

### Verification performed

| Claim | Result |
| --- | --- |
| "either" still resolves correctly | ✅ points at the two files named in the immediately preceding clause |
| Linux path list survives the move | ✅ no competing referent in the section — `NEXUS_BROWSER_PATH` is the only path-valued item under this heading |
| `npx prettier --check CONTRIBUTING.md` | ✅ |

---

INNOVATIVEGAMER:
Closing — we don't need this change.

No reflection on the work; the per-machine browser MCP path just isn't something we want to take on right now. Leaving the branch in place in case we revisit it.

#685 is being closed alongside this as the earlier version of the same fix.

Issue #696, opened 2026-09-10
refactor(docs): read rails and routes from the manifest, delete the three maps


neha-tanti:
## Context

With the manifest generated (#695), the three hand-maintained maps are still the live inputs. Until they are deleted they can still drift from the generated file, which is worse than before — now there are four things that have to agree.

## Goal

The manifest is the only input to navigation and routing, and the three arrays are gone.

## Scope

Rewire every consumer to read the manifest:

- `LeftRail` and `Breadcrumb` (via `SectionLayout`) — section and sub-page labels
- `generateStaticParams` in `app/[section]/page.tsx` and `app/[section]/[sub]/page.tsx`
- The page-resolution path in `app/[section]/[sub]/page.tsx` — MDX loader and hand-built page lookup
- `TopNav` and `SearchPalette` also read `SECTIONS` today; they have to move too, or the arrays cannot be deleted

Then delete `SECTIONS`, `MDX_PAGES` and `REAL_PAGES`.

## Acceptance criteria

- `SECTIONS`, `MDX_PAGES` and `REAL_PAGES` do not appear anywhere in the codebase.
- The site builds and navigates exactly as before — same routes, same rail, same breadcrumbs, same MDX and hand-built page resolution.
- Adding a page to the manifest makes it reachable with no other edit.

## Notes

Ticket **2 of 3 in Stage 2** of the docs build order. Depends on #695. Do this in the same person's pass as #695 — a half-migrated state where both the manifest and the arrays are live is the worst of both.

Issue #679, opened 2026-08-27
chore(repo): declare docs build outputs in turbo.json


neha-tanti:
## Context

`turbo.json` declares no build outputs for the docs app. With nothing declared, Turbo has nothing to cache, so the docs app rebuilds on every run — locally and in CI — even when no docs source changed.

## Goal

The docs build participates in Turbo's cache.

## Scope

- Add the docs build output paths to the docs `build` task in `turbo.json`.
- Confirm the declared paths match what the build actually emits, including the static output.
- Check that nothing needed at runtime is left out of the declared set — an under-declared output is worse than none, because it caches a broken artifact.

## Acceptance criteria

- A second `turbo build` with no changes reports a cache hit for the docs app.
- Editing a docs source file busts the cache and triggers a real rebuild.
- The cached artifact serves correctly — a restored build is not missing files.

## Notes

Ticket **7 of 7 in Stage 1** of the docs build order. Independent; touches only `turbo.json`.

PR #698, opened 2026-09-10
feat(docs): generate one page manifest from the docs filesystem

### LINKED ISSUE #695: feat(docs): generate one page manifest from the docs filesystem
## Context

The docs IA is currently spread across three hand-maintained maps that all have to agree:

- `SECTIONS` in `apps/docs/app/_lib/sections.ts` — sections, sub-pages, labels, order
- `MDX_PAGES` in `apps/docs/app/_lib/real-pages.tsx` — `${section}/${sub}` → lazy MDX import
- `REAL_PAGES` in `apps/docs/app/_lib/real-pages.tsx` — `${section}/${sub}` → hand-built component

Adding a page means editing all three by hand. Miss one and the build still passes — the page just cannot be reached. This has to be fixed before the 64 component pages arrive, not after: three maps mean three chances to be wrong on every page added.

## Goal

One generated manifest, produced by a build script that walks the docs filesystem, describing every page the site has.

## Scope

- A build script that walks the docs sources (`content/`, `app/_pages/`, and the section registry) and emits a single manifest.
- Each entry carries at minimum: route, title/label, section, and order.
- The manifest is generated output, wired into the docs build so it regenerates before Next builds.
- This ticket only produces the manifest. Nothing reads it yet — the rewire is the next ticket.

## Acceptance criteria

- The generated manifest lists every page the site currently has, matching `SECTIONS`, `MDX_PAGES` and `REAL_PAGES` combined — no page missing, no page invented.
- Adding a file on disk and re-running the generator adds it to the manifest with no hand-editing.
- The generator is deterministic: two runs with no source change produce byte-identical output.

## Notes

Ticket **1 of 3 in Stage 2** of the docs build order. Nothing depends on this being merged to start reading it, but the rewire ticket does.

neha-tanti:
## Summary

- Adds a build script that walks the docs sources and emits the site's page manifest.
  - `apps/docs/scripts/page-manifest.mjs` — the builder (`buildPageManifest(docsRoot)` → formatted TS source, keyed by output path)
  - `apps/docs/scripts/generate-page-manifest.mjs` — the CLI that writes it
  - Follows the existing `engine-snapshot-matrix.mjs` / `generate-engine-snapshot.mjs` split in `packages/core/scripts/`.
- Wires the generator ahead of both `next build` and `next dev`, adds a standalone `pnpm --filter @nexus_ds/docs generate:manifest`, and a CI step that regenerates and fails on a stale checked-in manifest.

### Two modules, split at the client boundary

The nav is client code — `TopNav` and `SearchPalette` import the IA at module scope from `'use client'` files — while page bodies are dynamic imports of server modules. Those can't share a module, so the generator emits two:

| Output | Holds | Side |
| --- | --- | --- |
| `app/_lib/page-manifest.generated.ts` | `PAGE_MANIFEST` — `route`, `slug`, `label`, `nested?`, `kind`, `file` | Serializable; safe from a client component |
| `app/_lib/page-content.generated.ts` | `PAGE_LOADERS` (route → dynamic import) and `PAGE_WIREFRAMES` (route → registry `lede` / `blocks`) | `import 'server-only'` |

7 sections, 31 pages. `kind` is a discriminated union: `mdx` / `component` carry a `file`, `placeholder` carries `file: null` and gets its body from `PAGE_WIREFRAMES`, so no consumer null-checks a page that has a source file.

`Block` moved out of `sections.ts` into `app/_lib/blocks.ts` — the generated server-only module needs the type, and `sections.ts` is what #696 deletes.

### Sources and precedence

| Source | Contributes |
| --- | --- |
| `app/_lib/sections.ts` | section order / title / href, sub-page order / label / nested, and the `lede` / `blocks` wireframe |
| `content/{section}/{slug}.mdx` | MDX pages |
| `app/_pages/{section}/{slug}.tsx` | hand-built pages |

Resolution order is `mdx → component → placeholder`, mirroring what `app/[section]/[sub]/page.tsx` already does at runtime. Pages found on disk but absent from the registry are appended to their section in slug order with a label derived from the slug, so a new file needs no hand-edit anywhere. A file at any other depth throws — a silent drop is the exact failure mode this ticket exists to remove. A route claimed by two sources throws too, and so does a `nested` label whose page now exists (`DropdownMenu` and `dropdown-menu.tsx` are compared on a folded key, not on display text).

### `app/_pages/` restructured to mirror routes

`ColorShowcase.tsx` → `foundations/color.tsx`, `MultiBrand.tsx` → `theming/multi-brand.tsx`, and so on, each now a default export. This is load-bearing rather than cosmetic: `ColorShowcase` does not derive from `foundations/color`, so without the move the "adding a file on disk adds it to the manifest" criterion cannot hold for hand-built pages — the mapping would stay hand-maintained, which is the thing being deleted. `ColorScales.tsx` was a client island used by the Color page, not a page itself, so it moved to `_components/`.

`REAL_PAGES` was re-pointed at the new paths and stays live until #696.

## GitHub Issue

Closes #695

## Test Plan

- [x] `pnpm --filter @nexus_ds/docs build` — 44 static pages, the same 31 sub-routes as before
- [x] `pnpm --filter @nexus_ds/docs typecheck`
- [x] `pnpm lint` — clean at `--max-warnings 0`
- [x] `pnpm vitest run --project=unit apps/docs` — 29 tests, including the 11 in `apps/docs/scripts/page-manifest.test.js`:
  - the manifest matches `SECTIONS` label-for-label and position-for-position, and tolerates a page that exists only on disk
  - its `mdx` / `component` routes equal the `MDX_PAGES` / `REAL_PAGES` keys exactly, and every non-placeholder `file` exists on disk (AC 1: no page missing, no page invented)
  - `PAGE_WIREFRAMES` covers exactly the placeholder routes, and the client-safe module contains no `import(` and no `lede:`
  - the checked-in modules equal a fresh build, and two consecutive builds are byte-identical (AC 3)
  - a fixture docs tree proves a new `.mdx` file, and a whole new section, appear with no hand-editing (AC 2)
  - a misplaced page file, a doubly-claimed route, and a stale `nested` label are each rejected with a pointed message

## Notes

Ticket 1 of 3 in Stage 2. Per the scope, nothing reads the manifest yet — #696 does the rewire and deletes the three maps. That ticket is worth doing in the same pass: until the arrays are gone there are four things that have to agree rather than three, and the registry-parity test above is what holds that line in the meantime.

Two changes outside `apps/docs`:

- `vitest.config.ts` gains `apps/**/scripts/**/*.test.{js,ts}` in the unit project, matching the `packages/**/scripts/**` pattern already there, plus a unit-project-scoped `server-only` alias (the package throws unless resolved under the `react-server` condition; scoping it keeps the browser project failing loudly on a client component that imports it).
- `.github/workflows/ci.yml` regenerates the manifest and fails on a dirty `apps/docs/app/_lib`.

Turbo's docs `build` inputs still don't cover `content/**`, `app/**`, or `scripts/**`, so a cache hit can skip regeneration. That is #679's scope — it owns the docs cache contract end to end, and the CI freshness step above covers the gap until it lands.

`jiti` is added as an explicit docs devDependency — the registry is TypeScript and the repo runs Node 20, which has no type stripping. `server-only` is added as a docs dependency for the generated content module.


neha-tanti (commented):

### Issues

| 1 | ❌ Blocking | `apps/docs/scripts/page-manifest.mjs:136` | `load` thunks share a module with the nav data client components must import | Emit two modules: serializable IA data, plus a server-only loader map |
| 2 | ❌ Blocking | `apps/docs/scripts/page-manifest.mjs:198` | `lede`/`blocks` excluded while `nested` is copied through | Copy the registry's page-scoped render fields through too |
| 3 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:85` | Specifier base `'app/_lib'` is not derived from `MANIFEST_FILE` | `path.dirname(MANIFEST_FILE)` |
| 4 | ⚠️ Minor | `apps/docs/app/_lib/page-manifest.generated.ts:18` | `order` duplicates array position in a nested manifest | |
| 5 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:45` | `_`-prefix co-location escape hatch is undocumented | |
| 6 | ⚠️ Minor | `apps/docs/package.json:9` | Turbo's docs `build` inputs don't cover `content/**`, `app/**`, `scripts/**` | Cite #679, which owns the docs cache contract |

### Architectural Concerns

**Manifest is not client-safe** (`apps/docs/scripts/page-manifest.mjs:136`) — `TopNav` and `SearchPalette` import the IA registry at module scope from a `'use client'` file, and `LeftRail` receives its section as a prop from a server layout, so #696 cannot consume this shape: `load` is a function and dynamic-imports server page modules. Split the emitted output into a serializable data module (`route`/`slug`/`label`/`order`/`nested`/`kind`/`file`) and a server-only route→loader module.

**Placeholder content stays hand-maintained** (`apps/docs/scripts/page-manifest.mjs:198`) — 24 of 31 pages render from `SECTIONS.subs[].lede` / `.blocks` via `SubPageView`, so #696's "SECTIONS does not appear anywhere" AC cannot hold against this manifest. Copy those fields through the generator like `nested` already is, leaving `sections.ts` as a generator-only authoring input.

**Route depth is a hard two-level invariant** (`apps/docs/scripts/page-manifest.mjs:64`) — the registry's `nested` labels (`Button`, `Input`, …) are the 64 component pages in waiting, and a `components/inputs/button` layout throws in the collector and changes every `route` consumer. Commit to flat `components/{name}` pages now and drop `nested` as those pages land.

---

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+  label: string;
+  /** Position within the section, zero-based. */
+  order: number;
+  /** Non-interactive labels rendered under this page in the left rail. */
+  nested?: readonly string[];
+  kind: ManifestPageKind;
+  /** Source file relative to \`apps/docs\`; \`null\` for placeholder pages. */
+  file: string | null;
+  /** Lazy module loader; \`null\` for placeholder pages. */
+  load: ManifestPageLoader | null;
```
neha-tanti: **Blocking — loaders and nav data in one module:** `TopNav`/`SearchPalette` import the registry at module scope from `'use client'` files, and `LeftRail` gets its section as a prop from a server layout. `load` functions can't cross that boundary. Emit loaders as a separate server-only module.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+    const source = sources.find((candidate) => candidate.pages.has(key));
+    const page = {
+      route: `/${key}`,
+      slug,
+      label: sub?.label ?? humanize(slug),
+      order,
+      kind: source?.kind ?? 'placeholder',
+      file: source?.pages.get(key) ?? null,
+    };
+    if (sub?.nested) {
```
neha-tanti, line 292: **Blocking — registry fields split arbitrarily:** `nested` is copied through but `lede`/`blocks` aren't, and `SubPageView` renders those for 24 of 31 pages. #696 can't delete `SECTIONS` until they're in the manifest too.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+      found.set(`${entry.name}/${slug}`, `${dir}/${entry.name}/${file.name}`);
+    }
+  }
+
+  return found;
+}
+
+/** Import specifier for a docs-root-relative file, resolved from `app/_lib/`. */
+function specifierFor(file, keepExtension) {
+  const relative = toPosix(path.relative('app/_lib', file));
```
neha-tanti: **Minor — output path duplicated:** the `'app/_lib'` base here has to agree with `MANIFEST_FILE`, but nothing ties them together. Derive it with `path.dirname(MANIFEST_FILE)`.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+
+function toPosix(relative) {
+  return relative.split(path.sep).join('/');
+}
+
+function readDirSorted(dir) {
+  return fs
+    .readdirSync(dir, { withFileTypes: true })
+    .filter(
+      (entry) => !entry.name.startsWith('_') && !entry.name.startsWith('.')
```
neha-tanti, line 52: **Minor — undocumented escape hatch:** this `_` skip is the only way to co-locate a page-local island under `_pages/`. Say so in the header comment; with 64 component pages, `_components/` is otherwise the sole home for every demo.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+function collectPages(docsRoot, { dir, ext }) {
+  const found = new Map();
+  const root = path.join(docsRoot, dir);
+  if (!fs.existsSync(root)) {
+    return found;
+  }
+
+  for (const entry of readDirSorted(root)) {
+    if (!entry.isDirectory()) {
+      throw new Error(
```
neha-tanti, line 66: **Minor — depth-2 is a hard invariant:** the registry’s `nested` labels are the 64 component pages in waiting. Confirm they land as `components/{name}`, not `components/inputs/{name}`, or this throw plus every `route` consumer has to change.

apps/docs/app/_lib/page-manifest.generated.ts
```diff
@@ -0,0 +1,384 @@
...
+// eslint-disable-next-line @nexus_ds/no-render-prop-types -- `default: ComponentType` is the shape of a lazily-imported page module, not a component-as-prop.
+export type ManifestPageLoader = () => Promise<{ default: ComponentType }>;
+
+export type ManifestPage = {
+  /** Route path, e.g. `/foundations/color`. */
+  route: string;
+  slug: string;
+  label: string;
+  /** Position within the section, zero-based. */
+  order: number;
```
neha-tanti: **Minor — `order` duplicates array position:** in a nested manifest the index already is the order, giving consumers two ways to sort. Drop both `order` fields, or emit a flat page list where the field earns its place.

apps/docs/package.json
```diff
@@ -5,9 +5,10 @@
   "type": "module",
   "description": "Documentation site for the Nexus Design System",
   "scripts": {
-    "dev": "next dev --turbopack",
-    "build": "next build",
+    "dev": "pnpm generate:manifest && next dev --turbopack",
+    "build": "pnpm generate:manifest && next build",
```
neha-tanti, line 9: **Minor — generator sources aren't hashed:** turbo's `build` inputs are `src/**` plus token files; docs has no `src/`, so `content/**`, `app/**`, and `scripts/**` never bust the cache and a hit skips regeneration entirely. #679 owns this.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,141 @@
...
+
+  it('is checked in current, and rebuilds byte-identically', async () => {
+    const committed = fs.readFileSync(
+      path.join(docsRoot, MANIFEST_FILE),
+      'utf8'
+    );
+    const first = await buildPageManifest(docsRoot);
+    const second = await buildPageManifest(docsRoot);
+
+    expect(first).toBe(committed);
```
neha-tanti: **Blocking — CRLF comparison:** on a Windows checkout (repo default `core.autocrlf=true`, no `.gitattributes`) the committed file is CRLF while prettier emits LF, so this assertion fails — reproduced on this branch. Normalise `\r\n` before comparing, or check in a `.gitattributes` `eol=lf` entry for the generated file.

neha-tanti (commented):
## SDE2 Review — Verdict: CHANGES REQUIRED

### Issues

| 1 | ❌ Blocking | `apps/docs/scripts/page-manifest.test.js:97` | Byte-identical test fails on any Windows checkout — CRLF vs LF | Normalise `\r\n` before comparing, or add a `.gitattributes` `eol=lf` entry |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:226` | Prettier plugin specifier resolves from `process.cwd()`, not the config file | Strip `plugins` from the resolved config |
| 3 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:96` | `specifier` interpolated into quotes unescaped; `'` in a filename emits invalid TS | `JSON.stringify(specifier)` |
| 4 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:189` | A route present in both sources silently drops the `.tsx` | Throw when a key appears in more than one source |
| 5 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:209` | `slug in SECTIONS` matches prototype keys; a `constructor/` section is dropped | `Object.hasOwn(SECTIONS, slug)` |
| 6 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:71` | Throw message calls a directory "not a `.mdx` page" | Split the `isFile` and extension checks |
| 7 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:47` | `localeCompare` listing sort never reaches output; contradicts the determinism comment | Drop the sort; output order comes from the explicit `.sort()` calls |
| 8 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:134` | `kind` does not narrow `file` / `load`; consumers must null-check `mdx` pages | Emit a union discriminated on `kind` |
| 9 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:71` | Element-wise route comparison fails on `MDX_PAGES` ordering, not correctness | Compare sorted arrays or sets |
| 10 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:111` | Quote rewrite masks that fixture builds use prettier defaults, not repo config | Pass explicit format options into `buildPageManifest` |

### Research Flags

| --- | --- | --- |
| prettier 3 | `resolveConfig` returns `plugins` as bare specifiers; `format` resolves them against `process.cwd()`, so a foreign cwd throws `ERR_MODULE_NOT_FOUND`. | Omit `plugins` from the options passed to `format` — the manifest has no class strings to sort. |

---

_Formal REQUEST_CHANGES state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a REQUEST_CHANGES self-review. The verdict above stands._

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+      order,
+      pages: orderedSlugs(sectionSlug).map((slug, index) =>
+        buildPage(sectionSlug, slug, index)
+      ),
+    })
+  );
+
+  const outputPath = path.join(docsRoot, MANIFEST_FILE);
+  return prettier.format(renderModule(manifest), {
+    ...(await prettier.resolveConfig(outputPath)),
```
neha-tanti: **Minor — plugin resolution is cwd-relative:** `resolveConfig` returns `plugins: ['prettier-plugin-tailwindcss']` as a bare specifier, and `format` resolves it from `process.cwd()`, not the config file. Invoking from outside the repo throws `ERR_MODULE_NOT_FOUND`. Drop `plugins` from the resolved config — the manifest has no class strings.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+    ? relative
+    : relative.slice(0, relative.lastIndexOf('.'));
+}
+
+function renderPage({ page, specifier }) {
+  const fields = Object.entries(page).map(
+    ([key, value]) => `${key}: ${JSON.stringify(value)},`
+  );
+  const load = specifier
+    ? `load: () => import('${specifier}'),`
```
neha-tanti: **Minor — unescaped specifier:** interpolating into single quotes breaks on a filename containing `'`; `bob's.mdx` emits invalid TS and surfaces as an opaque prettier parse error. Use `JSON.stringify(specifier)`, as the other fields already do.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+      .sort();
+    return [...registered, ...extra];
+  }
+
+  function buildPage(sectionSlug, slug, order) {
+    const key = `${sectionSlug}/${slug}`;
+    const sub = SECTIONS[sectionSlug]?.subs.find(
+      (entry) => entry.slug === slug
+    );
+    const source = sources.find((candidate) => candidate.pages.has(key));
```
neha-tanti: **Minor — duplicate route silently shadowed:** with both `content/x/y.mdx` and `app/_pages/x/y.tsx` present, the `.tsx` disappears from the manifest with no signal — the failure mode this generator exists to remove. Throw when a key resolves in more than one source.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+    }
+    const specifier = source
+      ? specifierFor(page.file, source.keepExtension)
+      : null;
+    return { page, specifier };
+  }
+
+  /** Registry order first, then sections that exist only on disk, in slug order. */
+  const extraSections = [...new Set(keysOnDisk.map((key) => key.split('/')[0]))]
+    .filter((slug) => !(slug in SECTIONS))
```
neha-tanti: **Minor — `in` matches prototype keys:** a section directory named `constructor` or `toString` is silently dropped from the manifest. Use `Object.hasOwn(SECTIONS, slug)`, and the same for the `SECTIONS[sectionSlug]` lookups above.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+  for (const entry of readDirSorted(root)) {
+    if (!entry.isDirectory()) {
+      throw new Error(
+        `${dir}/${entry.name} is not inside a section folder — docs pages live at {section}/{slug}${ext}.`
+      );
+    }
+
+    for (const file of readDirSorted(path.join(root, entry.name))) {
+      if (!file.isFile() || !file.name.endsWith(ext)) {
+        throw new Error(
```
neha-tanti, line 74: **Minor — inaccurate message:** this branch also fires for directories, so a nested `content/foundations/parts/` is reported as "not a .mdx page". Split the `!file.isFile()` and extension checks so each throws the message that matches it.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+  return relative.split(path.sep).join('/');
+}
+
+function readDirSorted(dir) {
+  return fs
+    .readdirSync(dir, { withFileTypes: true })
+    .filter(
+      (entry) => !entry.name.startsWith('_') && !entry.name.startsWith('.')
+    )
+    .sort((a, b) => a.name.localeCompare(b.name, 'en'));
```
neha-tanti: **Minor — dead sort:** listing order never reaches the output; the emitted order comes from registry order plus the `.sort()` calls in `orderedSlugs` / `extraSections`, which use code-unit order rather than this collation. Drop the sort and the determinism claim on line 157.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,229 @@
...
+  route: string;
+  slug: string;
+  label: string;
+  /** Position within the section, zero-based. */
+  order: number;
+  /** Non-interactive labels rendered under this page in the left rail. */
+  nested?: readonly string[];
+  kind: ManifestPageKind;
+  /** Source file relative to \`apps/docs\`; \`null\` for placeholder pages. */
+  file: string | null;
```
neha-tanti: **Minor — `kind` doesn't narrow `file` / `load`:** consumers must null-check even for `kind: 'mdx'`. Emit a union instead: `{ kind: 'placeholder'; file: null; load: null }` | `{ kind: 'mdx' | 'component'; file: string; load: ManifestPageLoader }`.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,141 @@
...
+      }))
+    );
+
+    expect(
+      pages.map(({ route, label, order }) => ({ route, label, order }))
+    ).toEqual(expected);
+  });
+
+  it('resolves MDX and hand-built pages to their source files', () => {
+    expect(routesOfKind('mdx')).toEqual(
```
neha-tanti, line 94: **Minor — order coupling:** this compares route arrays element-wise, so it fails whenever `MDX_PAGES` insertion order diverges from registry order even though both are correct. Compare sorted arrays or sets, so the test fails only for a missing or invented page.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,141 @@
...
+    const root = writeFixture({
+      'app/_lib/sections.ts': FIXTURE_REGISTRY,
+      'content/foundations/color.mdx': '# Color\n',
+      'content/foundations/motion.mdx': '# Motion\n',
+      'app/_pages/theming/multi-brand.tsx': 'export default function P() {}\n',
+    });
+
+    // A throwaway root has no prettier config to resolve, so the emitted
+    // quotes are prettier's double-quote default rather than the repo's.
+    const source = (await buildPageManifest(root)).replaceAll('"', "'");
```
neha-tanti: **Minor — normalisation hides a gap:** a fixture root resolves no prettier config, so this path exercises prettier defaults rather than the repo’s. Pass explicit format options into `buildPageManifest` instead of rewriting every double quote in the output.

neha-tanti (commented):
## Principal Architect Follow-up — Verdict: NEEDS DISCUSSION

### Unresolved Issues

| 1 | `load` thunks share a module with client-imported nav data | `apps/docs/scripts/page-manifest.mjs:169` | ✅ Fixed | Split into `PAGE_MANIFEST` and server-only `PAGE_LOADERS`. |
| 2 | `lede`/`blocks` excluded while `nested` copied through | `apps/docs/scripts/page-manifest.mjs:317` | ✅ Fixed | `PAGE_WIREFRAMES` carries both for every placeholder. |
| 3 | Specifier base not derived from the output path | `apps/docs/scripts/page-manifest.mjs:141` | ✅ Fixed | `path.dirname(CONTENT_FILE)`. |
| 4 | `order` duplicates array position | `apps/docs/app/_lib/page-manifest.generated.ts` | ✅ Fixed | Field dropped from both levels. |
| 5 | `_`-prefix co-location escape hatch undocumented | `apps/docs/scripts/page-manifest.mjs:19` | ✅ Fixed | Header comment states it. |
| 6 | Turbo docs `build` inputs miss `content/**`, `app/**`, `scripts/**` | `turbo.json:5` | ❌ Still Present | #679 owns the docs cache contract; cite it in the PR body. |

Route depth: the header now commits to flat `components/{name}`, and `assertNestedLabelsHaveNoPage` enforces the transition.

### New Issues

| 1 | ⚠️ Minor | `vitest.config.ts:38` | Root-level alias disables `server-only` for every vitest project | Scope the alias to the `unit` project |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:221` | Generated server-only module type-depends on the registry #696 deletes | Move `Block` to its own module |
| 3 | ⚠️ Minor | PR description | Body still documents `order`, `load` thunks, and one output module | Rewrite for the split, `server-only`, and the CI gate |

---

vitest.config.ts
```diff
@@ -32,6 +32,15 @@ export default defineConfig({
         find: /^@\//,
         replacement: `${path.resolve(__dirname, './packages/react/src')}/`,
       },
+      // `server-only` throws unless resolved under the `react-server`
+      // condition, which no test runs under; point it at its own RSC build.
+      {
+        find: /^server-only$/,
```
neha-tanti: **⚠️ New:** This alias sits on the root config, so both projects lose the `server-only` guard — including the browser project, where a client component importing it should fail loudly. Scope the alias to the `unit` project instead.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,368 @@
...
+  );
+  const wireframeEntries = wireframes.map(
+    ({ route, lede, blocks }) =>
+      `${JSON.stringify(route)}: { lede: ${JSON.stringify(lede)}, blocks: ${JSON.stringify(blocks)} },`
+  );
+  return `${HEADER}
+
+import type { ComponentType } from 'react';
+
+import type { Block } from './sections';
```
neha-tanti: **⚠️ New:** The generated server-only module type-depends on `sections.ts`, the registry #696 deletes. Move `Block` into its own module (e.g. `app/_lib/blocks.ts`) and emit the import against that.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,368 @@
...
+    }
+  }
+}
+
+/**
+ * The registry's `nested` labels stand in for pages that do not exist yet. Once
+ * one does, the label and the page would both show in the left rail.
+ */
+function assertNestedLabelsHaveNoPage(section) {
+  const labels = new Set(section.pages.map((page) => page.label));
```
neha-tanti: **⚠️ New — blocking:** Comparing display labels means `DropdownMenu` never matches the humanized `Dropdown menu` that a `dropdown-menu.tsx` page produces, so the guard silently passes for most registry `nested` entries. Normalise both sides to lowercase alphanumerics before comparing.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,233 @@
...
+
+  it('rejects a nested label that now has a page of its own', async () => {
+    const root = writeFixture({
+      'app/_lib/sections.ts': `export const SECTIONS = {
+  components: {
+    slug: 'components',
+    title: 'Components',
+    href: '/components',
+    subs: [
+      { slug: 'inputs', label: 'Inputs', nested: ['Button'], lede: '', blocks: [] },
```
neha-tanti: **⚠️ New:** The fixture's single-word `Button` is the one case where label matching works. Use a two-word label like `DropdownMenu` with `app/_pages/components/dropdown-menu.tsx` so the test covers the labels the real registry actually holds.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,368 @@
...
+    };
+    if (sub?.nested) {
+      page.nested = sub.nested;
+    }
+
+    const source = sources.find((candidate) => candidate.pages.has(key));
+    if (!source) {
+      // No file on disk, so the page is registry-only and renders its wireframe.
+      Object.assign(page, { kind: 'placeholder', file: null });
+      return { page, wireframe: { lede: sub.lede, blocks: sub.blocks } };
```
neha-tanti: **⚠️ New:** `sub.lede` is unguarded while `sub?.label` ten lines above is optional. If `sub` can be undefined this throws a raw TypeError; if it cannot, drop the `?.`. Either way a sub missing `lede`/`blocks` emits `undefined` into a `PageWireframe`-typed record — throw with the section/slug like the other input errors.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,368 @@
...
+      .filter((key) => key.startsWith(`${sectionSlug}/`))
+      .map((key) => key.slice(sectionSlug.length + 1))
+      .filter((slug) => !registered.includes(slug))
+      .sort();
+    return [...registered, ...extra];
+  }
+
+  function buildPage(sectionSlug, slug) {
+    const key = `${sectionSlug}/${slug}`;
+    const sub = sectionFor(sectionSlug)?.subs.find(
```
neha-tanti: **⚠️ New:** `?.subs.find` throws `Cannot read properties of undefined` when a registry section omits `subs`, while `orderedSlugs` guards the same access with `?? []`. Guard consistently here too.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,233 @@
...
+    expect(manifest).not.toContain('lede:');
+    expect(readOnDisk(CONTENT_FILE)).toContain("import 'server-only';");
+  });
+
+  it('is checked in current, and rebuilds byte-identically', async () => {
+    const first = await buildPageManifest(docsRoot);
+    const second = await buildPageManifest(docsRoot);
+
+    for (const file of [MANIFEST_FILE, CONTENT_FILE]) {
+      expect(toLf(first[file])).toBe(readOnDisk(file));
```
neha-tanti: **⚠️ New:** `.prettierrc` sets `endOfLine: 'lf'` and the line-34 comment says prettier always emits LF, so `toLf` on the freshly built string is dead. Compare `first[file]` directly.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,233 @@
...
+    const expected = Object.values(SECTIONS).flatMap((section) =>
+      section.subs.map((sub) => ({
+        route: `/${section.slug}/${sub.slug}`,
+        label: sub.label,
+        nested: sub.nested,
+      }))
+    );
+
+    expect(
+      pages.map(({ route, label, nested }) => ({ route, label, nested }))
```
neha-tanti: **⚠️ New:** Exact equality against registry subs fails the first time someone drops in an unregistered page — the flow the module doc advertises. Assert the registry pages appear in order as a subset instead.

.github/workflows/ci.yml
```diff
@@ -318,6 +318,23 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Regenerate docs page manifest
+        id: regen-manifest
+        run: pnpm --filter @nexus_ds/docs generate:manifest
+
+      - name: Check docs page manifest freshness
+        if: ${{ !cancelled() && steps.regen-manifest.conclusion == 'success' }}
```
neha-tanti: **⚠️ New:** A step's default `if` is `success()`, which already skips it when `regen-manifest` fails, and the step has no `continue-on-error`. The explicit condition adds no behaviour — drop the line.

neha-tanti (commented):
## SDE2 Follow-up — Verdict: CHANGES REQUIRED

### Unresolved Issues

All previous issues resolved — CRLF normalisation (`toLf`), `plugins` stripped in `resolveFormatOptions`, `JSON.stringify(specifier)`, `assertOneSourcePerRoute`, `Object.hasOwn`, split directory/extension throws, dead sort removed, discriminated `kind` union, sorted route comparison, explicit fixture format options.

### New Issues

| 1 | ❌ Blocking | `apps/docs/scripts/page-manifest.mjs:132` | Nested-label guard compares display labels, so `DropdownMenu` never matches `Dropdown menu` | Normalise both sides to lowercase alphanumerics |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:221` | Fixture's single-word `Button` masks the gap in #1 | Use a two-word label plus `dropdown-menu.tsx` |
| 3 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:313` | `sub.lede` unguarded while `sub?.label` is optional ten lines up; emits `undefined` into a typed record | Throw with section/slug like the other input errors |
| 4 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:297` | `?.subs.find` throws when a section omits `subs`; `orderedSlugs` guards it | Guard consistently |
| 5 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:138` | `toLf` on freshly built output is dead — prettier emits LF | Compare `first[file]` directly |
| 6 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:81` | Exact equality breaks the documented "just drop a file in" flow | Assert registry pages as an ordered subset |
| 7 | ⚠️ Minor | `.github/workflows/ci.yml:326` | `if:` guard is equivalent to the default `success()` | Drop the line |

Verified locally: `pnpm vitest run --project=unit apps/docs` → 29 passed; `tsc --noEmit` and `eslint` clean on the changed files.

---

_Formal REQUEST_CHANGES state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a REQUEST_CHANGES self-review. The verdict above stands._

apps/docs/app/_lib/blocks.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Wireframe content blocks — the placeholder body a docs page renders until a
+ * real page file lands. Kept out of `sections.ts` so the generated
+ * `page-content.generated.ts` does not type-depend on the section registry.
+ */
+
+export type PlaceholderVariant =
```
neha-tanti, line 7: **⚠️ New — duplicated union:** `_components/Placeholder.tsx:6` declares a byte-identical `PlaceholderVariant`. Its `Record<PlaceholderVariant, string>` is exhaustive against its own copy, so adding a variant here compiles clean and renders `undefined`. Delete the copy and import from `../_lib/blocks`.

neha-tanti (commented):
## Principal Architect Follow-up — Verdict: NEEDS DISCUSSION

### Unresolved Issues

| 1 | Docs `build` inputs miss `content/**`, `app/**`, `scripts/**` | `turbo.json:5` | 🔄 Changed | #679 is open and its ACs cover cache-busting; citation accepted. |
| 2 | Root-level `server-only` alias disabled the guard for every project | `vitest.config.ts:36` | ✅ Fixed | Scoped to the `unit` project; 29/29 still pass. |
| 3 | Generated server-only module type-depended on the registry | `apps/docs/app/_lib/blocks.ts` | ✅ Fixed | `Block` now lives outside `sections.ts`. |
| 4 | PR description stale | PR body | ✅ Fixed | Rewritten for the split, `server-only`, and the CI gate. |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/_lib/blocks.ts:7` | `PlaceholderVariant` duplicated in `_components/Placeholder.tsx:6` | Delete the copy in `Placeholder.tsx`; import from `../_lib/blocks` |

---

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,375 @@
...
+        throw new Error(
+          `${key} resolves to both ${existing} and ${file} — a route has one source file.`
+        );
+      }
+      seen.set(key, file);
+    }
+  }
+}
+
+/** `DropdownMenu`, `dropdown-menu` and `Show / Hide` all fold to one key. */
```
neha-tanti: **⚠️ New — misleading comment:** `Show / Hide` folds to `showhide`, not to the same key as `DropdownMenu` / `dropdown-menu`. Reword so it doesn't claim all three collapse to one key.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,375 @@
...
+function comparisonKey(value) {
+  return value.toLowerCase().replace(/[^a-z0-9]/g, '');
+}
+
+/**
+ * The registry's `nested` labels stand in for pages that do not exist yet. Once
+ * one does, the label and the page would both show in the left rail.
+ */
+function assertNestedLabelsHaveNoPage(section) {
+  const slugs = new Set(section.pages.map((page) => comparisonKey(page.slug)));
```
neha-tanti: **⚠️ New — label matches slip through:** the set now holds only slugs, so a nested label equal to an existing page's *label* (rather than its slug) no longer trips the guard. Fold both slug and label into the set.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,375 @@
...
+      label: sub?.label ?? humanize(slug),
+    };
+    if (sub?.nested) {
+      page.nested = sub.nested;
+    }
+
+    const source = sources.find((candidate) => candidate.pages.has(key));
+    if (!source) {
+      // No file on disk, so the page is registry-only and renders its wireframe.
+      if (sub.lede === undefined || sub.blocks === undefined) {
```
neha-tanti: **⚠️ Partial:** `sub` is still dereferenced bare while line 305 uses `sub?.label`. Use `sub?.lede === undefined || sub?.blocks === undefined` so a missing registry entry hits this error rather than a TypeError.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,242 @@
...
+    ],
+  },
+} satisfies Record<string, unknown>;
+`,
+      'app/_pages/components/dropdown-menu.tsx':
+        'export default function P() {}\n',
+    });
+
+    await expect(buildPageManifest(root)).rejects.toThrow(
+      'components lists "DropdownMenu" both as a page and as a nested label'
```
neha-tanti: **⚠️ New — untested throw:** every other generator error has a fixture test; the new missing-`lede`/`blocks` throw has none. Add a fixture whose registry sub omits both and has no page file.

neha-tanti (commented):
## SDE2 Follow-up — Verdict: MINOR CHANGES

### Unresolved Issues

| 3 | `sub.lede` unguarded while `sub?.label` is optional | `apps/docs/scripts/page-manifest.mjs:314` | ⚠️ Partial | New throw guards the fields, not `sub` itself; still bare deref. |

Previous issues 1, 2, 4, 5, 6 and 7 are ✅ Fixed. `toLf` is not dead — `readOnDisk` still uses it.

### New Issues

| 1 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:127` | Docstring claims three unlike values fold to one key | Reword |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:137` | Guard now folds slugs only; label-only duplicates slip through | Fold both `page.slug` and `page.label` into the set |
| 3 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:239` | New missing-`lede`/`blocks` throw has no fixture test | Add one, like every other generator error |

Verified: `pnpm vitest run --project=unit apps/docs` 29/29, docs typecheck and `pnpm lint` clean. Vitest 4 `extends: true` concatenates the project `resolve.alias` with the root array rather than replacing it — confirmed by a root-aliased and a project-aliased import both resolving in the same `unit` run. `comparisonKey()` produces no false positive against the current registry.

---

_Formal review state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a non-COMMENT self-review. The verdict above stands._

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. `PlaceholderVariant` is declared once, in `apps/docs/app/_lib/blocks.ts`; `blocks.ts` stays type-only, so its four importers span the client boundary without pulling server code. The nested-label guard keys off the label the left rail actually renders, with slug as a superset, and stays section-scoped. `turbo.json` is untouched since `12cbcd8`, so the #679 citation still stands.

---

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,260 @@
...
+} satisfies Record<string, unknown>;
+`,
+    });
+
+    await expect(buildPageManifest(root)).rejects.toThrow(
+      'foundations/color has no page file'
+    );
+  });
+
+  it('rejects a nested label that now has a page of its own', async () => {
```
neha-tanti, line 239: ⚠️ **New:** This fixture collides on slug (`dropdown-menu`), so it passes with the old slug-only set — reverting the label fold leaves all 12 tests green. Add a case where the page slug differs from its label, e.g. `{ slug: 'menus', label: 'DropdownMenu' }` with `nested: ['DropdownMenu']` elsewhere.

neha-tanti (commented):
## SDE2 Follow-up — Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved.

### New Issues

| 1 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:239` | Label fold has no test — reverting it keeps 12/12 green | Add a fixture where the colliding page's slug differs from its label |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:147` | Throw names the nested label's owner, not the page that collided | Make `taken` a `Map` from key to page slug and name that slug |

Verified: `pnpm vitest run --project=unit apps/docs` 30/30, docs typecheck and `pnpm lint` clean. The generator succeeds against the shipped registry, so the widened guard produces no false positive — every nested-label key was checked against every page slug and label in its section; the closest pair, `inputs` / `input`, does not fold together. Mutation-checked both directions: blanking the wireframe guard fails the new fixture test, but reverting the label fold to slug-only leaves all 12 tests green.

---

_Formal review state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a non-COMMENT self-review. The verdict above stands._

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,380 @@
...
+    section.pages.flatMap((page) => [
+      comparisonKey(page.slug),
+      comparisonKey(page.label),
+    ])
+  );
+  for (const page of section.pages) {
+    for (const label of page.nested ?? []) {
+      if (taken.has(comparisonKey(label))) {
+        throw new Error(
+          `${section.slug} lists "${label}" both as a page and as a nested label under ${page.slug} — drop the nested label now the page exists.`
```
neha-tanti: ⚠️ **New:** `page.slug` here is the page owning the nested label, not the one that collided. With the label fold the collision can be a page whose slug looks nothing like the label, leaving no pointer. Make `taken` a `Map` from key to page slug and name that slug.

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. The `Set` → `Map` swap is contained inside `assertNestedLabelsHaveNoPage`; no manifest or content module shape, client-boundary split, specifier derivation, or `server-only` boundary changed, so none of the previously-fixed architectural issues regress.

---

apps/docs/scripts/page-manifest.test.js
```diff
@@ -0,0 +1,281 @@
...
+    ],
+  },
+} satisfies Record<string, unknown>;
+`,
+      'app/_pages/components/dropdown-menu.tsx':
+        'export default function P() {}\n',
+    });
+
+    await expect(buildPageManifest(root)).rejects.toThrow(
+      'but components/dropdown-menu is now a page'
```
neha-tanti: **⚠️ New:** Narrowing to the second clause dropped the only assertion on the owner slug — replacing `under ${page.slug}` with a literal keeps all 13 tests green. Widen both assertions to `lists "DropdownMenu" as a nested label under overlays, but components/dropdown-menu is now a page`.

neha-tanti (commented):
## SDE2 Follow-up — Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved. Mutation-checked both: dropping the label key from `pageFor` fails the new fixture, and naming `page.slug` instead of `existing` in the throw fails two tests.

### New Issues

| 1 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:257` | Narrowed assertion no longer pins the owner page in the message | Widen to `lists "DropdownMenu" as a nested label under overlays, but …` |

Verified: `pnpm vitest run --project=unit apps/docs` 31/31 (`page-manifest.test.js` 13/13), docs typecheck and `eslint` clean on both changed files.

---

_Formal review state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a non-COMMENT self-review. The verdict above stands._

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. The commit touches only `page-manifest.test.js` assertion strings; no source file, module boundary, or `server-only` surface changed, so nothing regresses. The guard builds `pageFor` over all pages before scanning nested labels, so both widened assertions are order-independent.

---

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. The widened assertions now bite: replacing `under ${page.slug}` with a literal fails 2 tests, where before this commit the same mutation left all 13 green.

Verified at `27d72f3`:

| Check | Result |
| --- | --- |
| `pnpm vitest run --project=unit apps/docs` | 31 passed (4 files); `page-manifest.test.js` 13/13 |
| Mutation: `under ${page.slug}` → literal | 2 failed / 11 passed |
| Mutation: drop the `page.label` key from `pageFor` | 1 failed / 12 passed |
| Mutation: Map value `page.slug` → `page.label` | 2 failed / 11 passed |
| `tsc --noEmit` (docs) | clean |
| `eslint apps/docs/scripts/page-manifest.test.js` | clean |

Working tree restored after mutation checks.

---

_Formal review state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a non-COMMENT self-review. The verdict above stands._

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. The flat `components/{name}` contract dropped from the header stays enforced at both failure points: a nested path hits `collectPages`' two-levels-deep error, and the correct flat page hits `assertNestedLabelsHaveNoPage`, which names the stale label and the colliding page. Replacing the `({ entries, ...section })` rest-destructure with four named fields also stops a future field on the intermediate from spreading silently into the client-imported `PAGE_MANIFEST`.

---

neha-tanti (commented):
## SDE2 Follow-up — Verdict: MINOR CHANGES

### Unresolved Issues

All previous issues resolved.

### New Issues

| 1 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:342` | Inlined `flatMap` wraps the `for...of` header across 3 lines | Keep `const entries = built.flatMap(…)` and iterate over `entries` |

Behaviour-preservation confirmed on the three risks in this refactor:

| Risk | Result |
| --- | --- |
| Emitted key order | `nested` is assigned to `base` before the spread, so `{ ...base, kind, file }` yields `route, slug, label, nested?, kind, file` — same as the prior `Object.assign` order |
| Inverted `buildPage` branch | Registry-only path keeps the same missing-`lede`/`blocks` throw and message |
| `if (specifier)` / `if (wireframe)` guards | Same predicate as the removed `.filter()`; `specifierFor` cannot return `''`, and each entry carries exactly one key |

No removed comment took away non-obvious logic — the deleted docstrings restated the function name or the error text, and the `comparisonKey` fold example and `plugins`-deletion reason were both kept.

Verified at `128eb45`: `pnpm vitest run --project=unit apps/docs` 31 passed (`page-manifest.test.js` 13); regenerating produces **no drift** in `apps/docs/app/_lib/`; docs typecheck, eslint and `prettier --check` clean.

---

_Formal review state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a non-COMMENT self-review. The verdict above stands._

apps/docs/scripts/page-manifest.mjs
```diff
@@ -0,0 +1,370 @@
...
+    slug: section.slug,
+    title: section.title,
+    href: section.href,
+    pages: section.entries.map((entry) => entry.page),
+  }));
+  manifest.forEach(assertNestedLabelsHaveNoPage);
+
+  const loaders = [];
+  const wireframes = [];
+  for (const { page, specifier, wireframe } of built.flatMap(
```
neha-tanti: **⚠️ New — wrapped loop header:** Inlining the `flatMap` splits the destructured names from the collection across three prettier-wrapped lines, ending in `)) {`. Restore the named `const entries = built.flatMap((section) => section.entries);` above the loop and iterate over `entries`.

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. The commit is a local hoist of `built.flatMap(…)` into a `const` above the loop — no change to module boundaries, the `PAGE_MANIFEST` / `PAGE_LOADERS` / `PAGE_WIREFRAMES` split, specifier derivation, or the nested-label guard.

---

neha-tanti (commented):

### Unresolved Issues

All previous issues resolved. The loop header is one line again.

Shadowing check: `entries` is the only bare binding of that name in `buildPageManifest`'s scope — the other occurrences are the `entries` object key, `section.entries` property access, and `Object.entries` inside `renderPage`. It is read only after its `const`, so there is no TDZ hazard.

Verified at `469229e`:

| Check | Result |
| --- | --- |
| `pnpm vitest run --project=unit apps/docs` | 31 passed (4 files); `page-manifest.test.js` 13 |
| `generate:manifest` drift | No change under `apps/docs/app/_lib/` |
| Docs typecheck | Clean |
| `eslint` / `prettier --check` on the changed file | Clean |

---

_Formal review state unavailable: the `nexuslabs-ai-bot` account is not authenticated on this machine, and GitHub rejects a non-COMMENT self-review. The verdict above stands._

PR #699, opened 2026-09-10
refactor(docs): read rails and routes from the manifest, delete the three maps

neha-tanti:
## Summary

Rewires every consumer of the docs IA onto the manifest #695 generates, then deletes the hand-maintained maps.

| Consumer | Reads now |
| --- | --- |
| `generateStaticParams` in `app/[section]/page.tsx` and `app/[section]/[sub]/page.tsx` | `PAGE_MANIFEST` |
| Page resolution in `app/[section]/[sub]/page.tsx` | `PAGE_MANIFEST` + `PAGE_LOADERS` |
| `LeftRail` / `Breadcrumb` (via `SectionLayout`) | `ManifestSection` |
| `TopNav`, `SearchPalette` | `PAGE_MANIFEST` |
| `PageWireframeView` (was `SubPageView`) | `PAGE_WIREFRAMES` |

`app/_lib/manifest.ts` is the one lookup over the manifest — `getSection` / `getPage`, replacing `getSection` / `getSubPage` / `getDefaultSub`.

### Resolution behaviour is unchanged

The old route tried MDX → hand-built → registry placeholder by string key. It now switches on the manifest's `kind`, which is the same three cases with the lookups already resolved:

- `placeholder` → `PageWireframeView`, breadcrumb included
- `component` → the loader's module rendered bare; those pages draw their own breadcrumb
- `mdx` → the loader's module behind a route-provided breadcrumb, so authors write content only

Rail links are `page.route` rather than a re-derived `${section.href}/${sub.slug}`.

### What got deleted, and what did not

`MDX_PAGES` and `REAL_PAGES` are gone outright — `PAGE_LOADERS` is generated from the same files. `SERVER_ROUTE_PAGES` was the only other export of `real-pages.tsx` and is unrelated to page resolution, so it moved to `app/_lib/server-route-pages.ts`.

`SECTIONS` is the one that could not simply be deleted: it is the generator's own input, and 20 of the 31 pages are `kind: 'placeholder'`, whose `lede` / `blocks` bodies exist nowhere on disk. Deleting the export would delete those page bodies. So the data moved out of the app instead — `apps/docs/scripts/page-registry.ts`, exporting `PAGE_REGISTRY`, sitting next to the generator that consumes it. Nothing under `app/` imports it, so the manifest is the only input to navigation and routing, and there is no second live map to drift: the registry is generated *from*, never compared *against*.

[...]

PR #700, opened 2026-09-10
fix(docs): derive the home page counts from the manifest

### LINKED ISSUE #697: fix(docs): derive the home page counts from the manifest
## Context

`apps/docs/app/page.tsx` states the size of the catalogue in hardcoded strings — the `{ n: '17', l: 'Components' }` stat tile, and the section cards below it (`'6 pages'`, `'5 groups · 17 components'`, `'3 pages'`, `'5 pages'`). Every one of these goes stale the moment a page is added, and nothing on the site notices.

Once the manifest exists (#695), the real numbers are already available at build time.

## Goal

The home page reports what the site actually contains, without anyone editing it.

## Scope

- Replace the hardcoded `{ n: '17', l: 'Components' }` stat with a count read from the manifest.
- Do the same for the per-section counts on the section cards — same failure mode, same page, same fix.
- No new hand-maintained number is introduced in the process.

## Acceptance criteria

- The home page says **64** components.
- Adding a component page changes the number without anyone editing `app/page.tsx`.
- The section-card counts match the manifest too.

## Notes

Ticket **3 of 3 in Stage 2** of the docs build order. Depends on #695; cleanest after #696.

neha-tanti:
## Summary

- The `Components` stat tile and every section card on the home page now come from the generated page manifest. `SECTION_CARDS` is built by mapping `PAGE_MANIFEST` — title, href, count and blurb all derived — so adding a section or a page moves the page with no edit to `app/page.tsx`. The `Bases` tile counts `BASE_CHIPS` for the same reason.
- Splits the registry's `nested` field in two: `components` for the component names a group page covers, `nested` for in-page headings. They looked alike in the rail but mean different things, and only one of them is a component count. The rail renders the union of both; the generator validates both; only `components` is counted.
- Adds `unit?: 'components'` to a registry section, carried through to the manifest. A section states what it is counted in rather than having it inferred from the shape of its pages, so the Components card stays correct when the per-component pages (#701) land flat.
- Adds `requireSection(slug)` to `app/_lib/manifest.ts` — throws naming the slug when the manifest has no such section, so a renamed or removed section fails the build instead of rendering a dead card. The home page uses it for both the `components` stat and the Getting Started exclusion.
- Adds `app/_lib/home-counts.ts` (`countComponents`, `describeSize`) with tests, plus tests for `requireSection` and for the section unit reaching the manifest.

Two of the hardcoded numbers were already stale, which is the failure mode the issue describes. Verified against a real `next build` of the home page:

| Card | Before | Now |
| --- | --- | --- |
| Components stat | `17` | `15` |
| Foundations | `6 pages` | `6 pages` |
| Components | `5 groups · 17 components` | `5 groups · 15 components` |
| Theming | `3 pages` | `4 pages` |
| Tools | `5 pages` | `5 pages` |
| Guidance | `3 pages` | `3 pages` |
| For AI agents | `3 pages` | `3 pages` |

The card blurbs are derived too, so `Theming` now lists Appearance and `Foundations` no longer abbreviates `Radius · Borders · Shadows`.

## GitHub Issue

Closes #697

## Stacked on

Based on `neha-tanti/issue-696-manifest-rails-routes` (#699), which carries the manifest from #695. Retarget to `main` once that merges.

## On the "64 components" acceptance criterion

The page says **15**, not 64. The manifest today has 5 component *group* pages carrying 15 component names; the 64 per-component pages do not exist yet. That work is now tracked as **#701** — when those pages land the number moves on its own, which is the criterion that actually matters here. Deriving 64 by counting `packages/react` instead was rejected: the issue says to read the manifest, and that count is 60–63 anyway.

#701 also covers the flat-vs-nested decision for those pages. Either shape renders correctly today: `describeSize` branches on the section's declared `unit`, so a flat section reads `64 components` rather than `64 pages` or `64 groups · 64 components`.

## Test Plan

- [x] `pnpm --filter @nexus_ds/docs typecheck`
- [x] `pnpm --filter @nexus_ds/docs build` — 44 static pages, home page counts asserted against the emitted HTML
- [x] `npx vitest run --project unit apps/docs` — 44 passing across 6 files
- [x] `npx eslint` + `prettier --check` on the changed files
- [x] Re-ran `pnpm --filter @nexus_ds/docs generate:manifest` — regenerated output is checked in and rebuilds byte-identically


apps/docs/app/page.tsx
```diff
@@ -36,42 +39,38 @@ const AUDIENCES = [
...
-    title: 'Components',
+    slug: 'components',
     desc: 'Inputs · Containers · Navigation · Display · Primitives',
-    href: '/components',
   },
   {
-    count: '3 pages',
-    title: 'Theming',
+    slug: 'theming',
     desc: 'Multi-brand · density modes · consumer overrides',
```
neha-tanti: **Blocking — stale blurb:** `theming` has four pages; this lists three (Appearance is missing), so the card now renders `4 pages` over a three-item list. Derive `desc` from `section.pages` labels instead of restating them by hand — that is the same staleness the PR removes one line up.

apps/docs/app/page.tsx
```diff
@@ -36,42 +39,38 @@ const AUDIENCES = [
...
-    desc: 'llms.txt · rules mirror · authoring',
-    href: '/agents',
-  },
-];
+  { slug: 'agents', desc: 'llms.txt · rules mirror · authoring' },
+].map(({ slug, desc }) => {
+  const section = requireSection(slug);
+  return {
+    count:
+      slug === 'components'
```
neha-tanti: **Blocking — `components` special case:** `N groups ·` is pinned to today's grouped shape and reads `64 groups · 64 components` once flat per-component pages land. No tracked issue owns that work, so derive the clause here from whether the section's pages carry nested names.

**Code quality:** this is also the file's only `slug === '...'` string branch inside the map — deriving it removes the discriminator outright.

apps/docs/app/_lib/manifest.ts
```diff
@@ -9,3 +9,25 @@ export { type ManifestPage, type ManifestSection, PAGE_MANIFEST };
...
+    );
+  }
+  return section;
+}
+
+/**
+ * Entries the section's left rail lists: a page's nested names when it groups
+ * several, otherwise the page itself.
+ */
+export function countRailEntries(section: ManifestSection): number {
```
neha-tanti: **Blocking — `nested` carries two meanings:** `foundations/color` uses it for in-page headings, `components/*` for component names. The `Components` stat therefore counts a rail-rendering hint. Split the registry field so the count reads something that means components.

apps/docs/app/page.tsx
```diff
@@ -36,42 +39,38 @@ const AUDIENCES = [
 
 const SECTION_CARDS = [
```
neha-tanti: **Minor — hand-maintained section subset:** adding a section to the registry silently produces no card here. Build the list from `PAGE_MANIFEST` (minus `getting-started`) keyed by a slug→blurb record, so a section without a blurb fails the build.

apps/docs/app/_lib/manifest.ts
```diff
@@ -9,3 +9,25 @@ export { type ManifestPage, type ManifestSection, PAGE_MANIFEST };
 export function getSection(slug: string): ManifestSection | undefined {
   return PAGE_MANIFEST.find((section) => section.slug === slug);
 }
+
+/** `getSection` for callers that link to a fixed section and cannot render without it. */
+export function requireSection(slug: string): ManifestSection {
```
neha-tanti, line 14: **Minor — client-graph placement:** `manifest.ts` is imported by `TopNav` and `LeftRail` (both `'use client'`), so these server-only home-page helpers land in the client module graph. Keep them in `app/page.tsx`.

neha-tanti (commented):
## 🤖 Principal Architect Review — #700

### Verdict: CHANGES REQUIRED

*(Posted as `COMMENT`: the `nexuslabs-ai-bot` account is not authenticated on this machine and GitHub rejects `REQUEST_CHANGES` on a self-authored PR.)*

### Issues

| 1 | ❌ Blocking | `apps/docs/app/page.tsx:51` | `theming` blurb lists 3 pages under a derived `4 pages` count | Derive `desc` from `section.pages` labels |
| 2 | ❌ Blocking | `apps/docs/app/page.tsx:66` | `slug === 'components'` hardcodes the grouped shape; no issue tracks the break | Derive the `groups` clause from whether pages carry `nested` |
| 3 | ❌ Blocking | `apps/docs/app/_lib/manifest.ts:28` | Component stat counts `nested`, a rail hint also used for in-page headings | Split the registry field so the count reads component identity |
| 4 | ⚠️ Minor | `apps/docs/app/page.tsx:40` | Card list is a hand-maintained subset; a new section silently gets no card | Build from `PAGE_MANIFEST` with a slug→blurb record |
| 5 | ⚠️ Minor | `apps/docs/app/_lib/manifest.ts:14` | Home-page-only helpers added to a module in the client bundle graph | Keep them in `app/page.tsx` |

### Architectural Concerns

**`countRailEntries` sits on the wrong boundary** (`apps/docs/app/_lib/manifest.ts:28`) — it is named for `LeftRail`, which never calls it, and its only consumer is a home-page catalogue statistic. Move it into `app/page.tsx` and name it for the number it produces.

**`Closes #697` with the 64-component criterion unmet** (`apps/docs/app/page.tsx:13`) — no tracked issue owns the Stage 3 per-component pages, so the PR body's deferral has nothing to point at. Amend #697's acceptance criteria or open the Stage 3 issue before merging.

---

neha-tanti (commented):
## 🤖 SDE2 Review — #700

### Verdict: MINOR CHANGES

*(Combined verdict across both agents is CHANGES REQUIRED — see the Principal Architect review. Posted as `COMMENT` because the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Issues

| 1 | ⚠️ Minor | `apps/docs/app/_lib/manifest.ts:25` | `countRailEntries` name and JSDoc contradict what `LeftRail` renders | Rail lists the page link *and* its nested labels |
| 2 | ⚠️ Minor | `apps/docs/app/_lib/manifest.ts:30` | `?? 1` lets a `nested: []` entry count a real page as zero | `page.nested?.length || 1` |
| 3 | ⚠️ Minor | `apps/docs/app/page.tsx:68` | A one-page section renders "1 pages" | Pluralize off the count |
| 4 | ⚠️ Minor | `apps/docs/app/page.tsx:16` | Bases tile hardcodes `'5'` while `BASE_CHIPS` sits in the same file | `String(BASE_CHIPS.length)` |
| 5 | ⚠️ Minor | `apps/docs/app/_lib/manifest.test.ts:47` | No coverage for `nested: []` | Add it beside the empty-section case |

### Verification

| Command | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ passes |
| `npx vitest run --project unit apps/docs/app/_lib/manifest.test.ts` | ✅ 4 passed |

---

apps/docs/app/_lib/manifest.ts
```diff
@@ -9,3 +9,25 @@ export { type ManifestPage, type ManifestSection, PAGE_MANIFEST };
...
+  if (!section) {
+    throw new Error(
+      `No '${slug}' section in the page manifest — add it to apps/docs/page-registry, or stop linking to it.`
+    );
+  }
+  return section;
+}
+
+/**
+ * Entries the section's left rail lists: a page's nested names when it groups
```
neha-tanti: **Minor — name/doc contradict LeftRail:** the rail renders the page link *and* its nested labels, so `components` lists 20 entries, not 15. This counts leaf subjects. Rename it for what it counts and drop the rail claim from the JSDoc.

apps/docs/app/_lib/manifest.ts
```diff
@@ -9,3 +9,25 @@ export { type ManifestPage, type ManifestSection, PAGE_MANIFEST };
...
+ * several, otherwise the page itself.
+ */
+export function countRailEntries(section: ManifestSection): number {
+  return section.pages.reduce(
+    (total, page) => total + (page.nested?.length ?? 1),
```
neha-tanti: **Minor — empty `nested` counts as zero:** `?? 1` only guards `undefined`, so a registry entry with `nested: []` drops its page from the count — `page-manifest.mjs:298` copies `[]` through, and LeftRail still renders the page. Use `page.nested?.length || 1`.

apps/docs/app/page.tsx
```diff
@@ -8,10 +8,13 @@ import {
   CardHeader,
   CardTitle,
 } from './_components/nexus';
+import { countRailEntries, requireSection } from './_lib/manifest';
+
+const COMPONENT_COUNT = countRailEntries(requireSection('components'));
 
 const STATS = [
   { n: '5', l: 'Bases' },
```
neha-tanti: **Minor — stale sibling:** `BASE_CHIPS` in this file is the source of truth for this 5. Use `String(BASE_CHIPS.length)` so the tile cannot drift the way `Components` just did.

apps/docs/app/page.tsx
```diff
@@ -36,42 +39,38 @@ const AUDIENCES = [
...
-  },
-];
+  { slug: 'agents', desc: 'llms.txt · rules mirror · authoring' },
+].map(({ slug, desc }) => {
+  const section = requireSection(slug);
+  return {
+    count:
+      slug === 'components'
+        ? `${section.pages.length} groups · ${COMPONENT_COUNT} components`
+        : `${section.pages.length} pages`,
```
neha-tanti: **Minor — `1 pages`:** the count is manifest-driven now, so a future one-page section renders bad grammar. Pluralize: `${n} ${n === 1 ? 'page' : 'pages'}`.

apps/docs/app/_lib/manifest.test.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+describe('countRailEntries', () => {
+  it('counts a grouping page once per nested name and an ungrouped page once', () => {
+    expect(countRailEntries(SECTION)).toBe(5);
+  });
+
+  it('counts nothing for a section with no pages', () => {
+    expect(countRailEntries({ ...SECTION, pages: [] })).toBe(0);
+  });
```
neha-tanti: **Minor — missing case:** nothing covers `nested: []`, which today returns 0 for a page the rail still lists. Add it beside the empty-section case.

apps/docs/page-registry/index.ts
```diff
@@ -10,8 +10,10 @@ import type { Block } from './blocks';
 export type RegistryPage = {
   slug: string;
   label: string;
-  /** Optional nested labels rendered inline in the left rail (non-interactive). */
+  /** Optional in-page headings rendered inline in the left rail (non-interactive). */
   nested?: string[];
+  /** Components a group page covers, rendered inline in the left rail (non-interactive). */
+  components?: string[];
```
neha-tanti, line 16: **⚠️ Partial — two optional siblings:** Nothing stops a page carrying both `components` and `nested`. LeftRail's `components ?? nested` drops the headings, while the generator's `railLabelsOf` concatenates both — three consumers, two invariants. Model one field with a kind, or a union so carrying both is a type error.

apps/docs/app/_components/LeftRail.tsx
```diff
@@ -19,6 +19,7 @@ export function LeftRail({ section }: { section: ManifestSection }) {
       <ul className="nx:list-none nx:p-0 nx:m-0">
         {section.pages.map((page) => {
           const active = pathname === page.route;
+          const railLabels = page.components ?? page.nested;
```
neha-tanti: **❌ New — silent drop:** `components ?? nested` discards `nested` when a page carries both, so in-page headings vanish from the rail with no build error — the generator validates both. Use `[...(page.components ?? []), ...(page.nested ?? [])]` and guard on `.length`, or make both-fields impossible in `RegistryPage`.

apps/docs/app/_lib/home-counts.ts
```diff
@@ -0,0 +1,19 @@
...
+export function countComponents(section: ManifestSection): number {
+  return section.pages.reduce(
+    (total, page) => total + (page.components?.length || 1),
+    0
+  );
+}
+
+/** How big a section is, in the units it is organised by. */
+export function describeSize(section: ManifestSection): string {
+  const groups = section.pages.filter((page) => page.components?.length).length;
```
neha-tanti: **❌ New — inferred unit:** The section's unit is read off page shape. When the per-component pages land flat, `groups` is 0 and this card reads "64 pages" while the stat tile says 64 Components. Put the unit on the section in the registry.

apps/docs/app/page.tsx
```diff
@@ -34,47 +39,15 @@ const AUDIENCES = [
...
-    desc: 'llms.txt · rules mirror · authoring',
-    href: '/agents',
-  },
-];
-
-const BLUE_RAMP = [50, 100, 200, 300, 400, 500, 600, 700, 800, 900, 950];
-const BASE_CHIPS = ['slate', 'stone', 'neutral', 'gray', 'zinc'];
+// Getting Started is reached through the audience cards above, not a section card.
+const SECTION_CARDS = PAGE_MANIFEST.filter(
+  (section) => section.slug !== 'getting-started'
```
neha-tanti: **⚠️ New — hardcoded slug:** Renaming `getting-started` silently adds a card here, unlike `requireSection` which fails loudly. Mark the exclusion on the section in `page-registry` instead.

neha-tanti (commented):
## 🤖 Principal Architect Follow-up — #700

### Verdict: CHANGES REQUIRED

*(Posted as `COMMENT`: the `nexuslabs-ai-bot` account is not authenticated on this machine and GitHub rejects `REQUEST_CHANGES` on a self-authored PR.)*

### Unresolved Issues

| 1 | Stale `theming` blurb under a derived count | `apps/docs/app/page.tsx:48` | ✅ Fixed | `desc` now joins `section.pages` labels. |
| 2 | `slug === 'components'` hardcoded the grouped shape | `apps/docs/app/_lib/home-counts.ts:13` | ✅ Fixed | Discriminator gone; clause derived from page shape. |
| 3 | Component stat counted `nested`, a rail hint | `apps/docs/page-registry/index.ts:16` | ⚠️ Partial | Split done; nothing forbids a page carrying both fields. |
| 4 | Card list a hand-maintained subset | `apps/docs/app/page.tsx:43` | ✅ Fixed | Built from `PAGE_MANIFEST`. |
| 5 | Home-page helpers in the client bundle graph | `apps/docs/app/_lib/home-counts.ts:1` | ✅ Fixed | Own module; `manifest.ts` import is type-only. |
| 6 | `Closes #697` with the 64-component criterion unmet | PR description | ❌ Still Present | No open issue owns Stage 3 per-component pages. |

### New Issues

| 1 | ❌ Blocking | `apps/docs/app/_lib/home-counts.ts:13` | Section unit inferred from page shape, wrong at the target state | Carry the unit on the section in `page-registry`, not off `components?.length` |
| 2 | ❌ Blocking | `apps/docs/app/_components/LeftRail.tsx:22` | `components ?? nested` silently drops one list | Make carrying both a type error in `RegistryPage`, or render the union |
| 3 | ⚠️ Minor | `apps/docs/app/page.tsx:44` | Hardcoded `getting-started` exclusion, silent on rename | Flag the section in `page-registry` so a rename fails the build |
| 4 | ⚠️ Minor | PR description | Documents the removed `countRailEntries` and `manifest.test.ts` | Rewrite the body for the shipped shape; drop the untracked "one-line fix later" |

### Architectural Concerns

**Stage 3 has nothing to point at** — `gh issue list` shows no open issue for the per-component pages, so the PR body's "when they land" deferral is untracked. Open that issue or amend #697's acceptance criteria before merging (`.claude/rules/no-follow-up-deferral.md`).

---

apps/docs/app/_lib/home-counts.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (groups > 0) {
+    return `${groups} groups · ${countComponents(section)} components`;
```
neha-tanti: **⚠️ New:** The page branch pluralizes but this one doesn't — a section with one group page renders "1 groups · 3 components". Pluralize `groups` (and `components`) the same way, and cover the one-group case in `home-counts.test.ts`.

apps/docs/app/page.tsx
```diff
@@ -34,47 +39,15 @@ const AUDIENCES = [
...
+).map((section) => ({
+  count: describeSize(section),
+  title: section.title,
+  desc: section.pages.map((page) => page.label).join(', '),
```
neha-tanti: **⚠️ New:** Foundations renders "Spacing, Radius · Borders · Shadows, Layering" — the `, ` separator is ambiguous against the ` · ` inside a label (`nx: prefix (Tailwind)` in Tools too). The old blurbs used ` · ` throughout; join with that instead.

neha-tanti (commented):
## 🤖 SDE2 Follow-up — #700

### Verdict: MINOR CHANGES

*(Combined verdict across both agents is CHANGES REQUIRED — see the Principal Architect follow-up. Posted as `COMMENT` because the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All five previous SDE2 issues resolved.

| 1 | `countRailEntries` name/JSDoc contradicted LeftRail | `apps/docs/app/_lib/home-counts.ts:4` | ✅ Fixed | Renamed `countComponents`, counts the split field. |
| 2 | `?? 1` counted an empty rail list as zero | `apps/docs/app/_lib/home-counts.ts:6` | ✅ Fixed | Truthy fallback; generator drops empty arrays too. |
| 3 | One-page section rendered "1 pages" | `apps/docs/app/_lib/home-counts.ts:18` | ✅ Fixed | Pluralized, with a test. |
| 4 | Bases tile hardcoded `'5'` | `apps/docs/app/page.tsx:18` | ✅ Fixed | `String(BASE_CHIPS.length)`. |
| 5 | No coverage for an empty rail list | `apps/docs/app/_lib/home-counts.test.ts:36` | ✅ Fixed | Covered for both helpers. |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/_lib/home-counts.ts:15` | Grouped branch says "1 groups" at one group | Pluralize like the page branch |
| 2 | ⚠️ Minor | `apps/docs/app/page.tsx:48` | `, ` separator collides with the ` · ` inside a label | Join with ` · ` |
| 3 | ⚠️ Minor | `apps/docs/app/_lib/manifest.ts:14` | `requireSection` lost its tests with `manifest.test.ts` | Re-add both cases to `home-counts.test.ts` |

### Verification

| Command | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ passes |
| `npx vitest run --project unit apps/docs/app/_lib/home-counts.test.ts` | ✅ 7 passed |
| `npx vitest run --project unit apps/docs/scripts/page-manifest.test.js` | ✅ 14 passed |

---

apps/docs/app/_lib/manifest.ts
```diff
@@ -9,3 +9,14 @@ export { type ManifestPage, type ManifestSection, PAGE_MANIFEST };
 export function getSection(slug: string): ManifestSection | undefined {
   return PAGE_MANIFEST.find((section) => section.slug === slug);
 }
+
+/** `getSection` for callers that link to a fixed section and cannot render without it. */
+export function requireSection(slug: string): ManifestSection {
```
neha-tanti, line 14: **⚠️ New:** Deleting `manifest.test.ts` dropped the only coverage of `requireSection` — both the happy path and the throw-with-slug message are now untested. Move those two cases into `home-counts.test.ts`.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -344,9 +359,10 @@ export async function buildPageManifest(docsRoot, formatOptions) {
     slug: section.slug,
     title: section.title,
     href: section.href,
+    ...(section.unit ? { unit: section.unit } : {}),
```
neha-tanti: **❌ New — unit/shape desync unguarded:** A section without `unit` whose pages carry `components` renders those names in the rail while the home card counts pages. The generator already asserts wireframe and rail-label desyncs; add the matching throw when `components` appears outside a `unit: 'components'` section.

**Code quality:** this conditional spread is also dead — `renderSection` already skips a falsy `unit`, and the returned manifest never escapes `buildPageManifest`. Use `unit: section.unit` and keep the single guard at the render site.

apps/docs/app/_lib/home-counts.ts
```diff
@@ -0,0 +1,23 @@
...
+export function countComponents(section: ManifestSection): number {
+  return section.pages.reduce(
+    (total, page) => total + (page.components?.length || 1),
+    0
+  );
+}
+
+/** How big a section is, in the unit it declares in the page registry. */
+export function describeSize(section: ManifestSection): string {
+  if (section.unit !== 'components') {
```
neha-tanti: **⚠️ New — non-exhaustive unit branch:** `!== 'components'` treats every future unit as pages. Adding `'tokens'` to the union compiles clean and silently mislabels the card. Switch on `section.unit` with an exhaustive default so a new member fails typecheck.

neha-tanti (commented):
## 🤖 Principal Architect Follow-up (round 2) — #700

### Verdict: CHANGES REQUIRED

*(Posted as `COMMENT`: the `nexuslabs-ai-bot` account is not authenticated on this machine and GitHub rejects `REQUEST_CHANGES` on a self-authored PR.)*

### Unresolved Issues

All six previous architectural issues resolved.

| 1 | `components` / `nested` — three consumers, two invariants | `apps/docs/app/_components/LeftRail.tsx:22` | ✅ Fixed | Rail unions both; matches the generator's `railLabelsOf`. |
| 2 | Counting unit inferred from page shape | `apps/docs/app/_lib/home-counts.ts:17` | ✅ Fixed | Section declares `unit`; flat Stage 3 shape reads correctly. |
| 3 | `components ?? nested` drops a list | `apps/docs/app/_components/LeftRail.tsx:22` | ✅ Fixed | Union-render was one of the two offered options. |
| 4 | Hardcoded `getting-started` slug | `apps/docs/app/page.tsx:43` | ✅ Fixed | `requireSection` throws at module eval on rename. |
| 5 | PR description documented removed code | PR description | ✅ Fixed | Body rewritten for the shipped shape. |
| 6 | `Closes #697` with the 64-component criterion unmet | PR description | ✅ Fixed | #701 owns Stage 3 and is cited — meets the tracked-issue exception. |

### New Issues

| 1 | ❌ Blocking | `apps/docs/scripts/page-manifest.mjs:362` | `unit` and page `components` can contradict, with nothing validating the pair | Throw when `components` appears outside a `unit: 'components'` section |
| 2 | ⚠️ Minor | `apps/docs/app/_lib/home-counts.ts:17` | Negated equality on a union — a second unit silently reads as pages | Switch on `section.unit` so an unhandled member fails typecheck |

### Architectural Concerns

**`unit` is on the right entity, read the wrong way** — the section is the thing being counted, and declaring it there is what decouples the label from page shape. The single-member union is defensible as a model; reading it with negated equality is what makes its one extensibility point silently wrong on first extension.

---

neha-tanti (commented):
## 🤖 SDE2 Follow-up (round 2) — #700

### Verdict: MINOR CHANGES

*(Combined verdict across both agents is CHANGES REQUIRED — see the Principal Architect follow-up. Posted as `COMMENT` because the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

| 1 | Grouped branch said "1 groups" | `apps/docs/app/_lib/home-counts.ts:20` | ✅ Fixed | `plural()` covers page, group and component. |
| 2 | Page labels collide with in-label ` · ` | `apps/docs/app/page.tsx:50` | ⚠️ Partial | Foundations reads as 8 items under a "6 pages" count. |
| 3 | `requireSection` lost its coverage | `apps/docs/app/_lib/manifest.test.ts` | ✅ Fixed | Happy path and throw message both covered. |
| 4 | Rail dropped `nested`, rendered empty `<ul>` | `apps/docs/app/_components/LeftRail.tsx:22` | ✅ Fixed | Unions both lists, guards on `.length`. |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/page.tsx:46` | Object-identity filter where a slug compare reads plainly | `section.slug !== AUDIENCE_SECTION.slug` |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.test.js:76` | New test subsumes the section-order test above it | Fold `unit` into the existing order test |

`describeSize` handles zero pages, all-empty `components` lists, and singular/plural for every noun; the reworked `home-counts.test.ts` kept all prior cases (7 → 9).

### Verification

| Command | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ passes |
| `npx vitest run --project unit apps/docs` | ✅ 44 passed across 6 files |
| `npx eslint apps/docs` | ✅ clean |
| `npx prettier --check` (9 changed files) | ✅ all match |

---

apps/docs/app/page.tsx
```diff
@@ -34,47 +39,17 @@ const AUDIENCES = [
...
-    href: '/agents',
-  },
-];
+// Getting Started is reached through the audience cards above, not a section card.
+const AUDIENCE_SECTION = requireSection('getting-started');
 
-const BLUE_RAMP = [50, 100, 200, 300, 400, 500, 600, 700, 800, 900, 950];
-const BASE_CHIPS = ['slate', 'stone', 'neutral', 'gray', 'zinc'];
+const SECTION_CARDS = PAGE_MANIFEST.filter(
+  (section) => section !== AUDIENCE_SECTION
```
neha-tanti: **⚠️ New — identity-based exclusion:** Identity holds only because `requireSection` returns the same `PAGE_MANIFEST` element, which its contract does not promise. Compare `section.slug !== AUDIENCE_SECTION.slug` — same existence assertion, no reliance on reference equality.

apps/docs/app/page.tsx
```diff
@@ -34,47 +39,17 @@ const AUDIENCES = [
...
+).map((section) => ({
+  count: describeSize(section),
+  title: section.title,
+  desc: section.pages.map((page) => page.label).join(' · '),
```
neha-tanti: **⚠️ Partial:** The joiner now matches the separator inside `Radius · Borders · Shadows`, so the Foundations card reads as 8 entries beside its "6 pages" count. Rename that registry label (e.g. `Radius, borders & shadows`) so the joiner stays unambiguous.

apps/docs/scripts/page-manifest.test.js
```diff
@@ -73,11 +73,27 @@ describe('page manifest', () => {
     ).toEqual(Object.keys(PAGE_REGISTRY));
   });
 
-  it('lists every registry page, with its label, nesting and position', () => {
+  it('carries each section’s counting unit through from the registry', () => {
```
neha-tanti: **⚠️ New:** This assertion implies the section-order test directly above, and copies its `registered` Set. Fold `unit` into that test's mapped tuple and delete this one.

apps/docs/app/_pages/foundations/radius.tsx
```diff
@@ -56,12 +56,10 @@ export default function Radius() {
         items={[
           { label: 'Home', href: '/' },
           { label: 'Foundations', href: '/foundations' },
-          { label: 'Radius · Borders · Shadows' },
+          { label: 'Radius, borders & shadows' },
```
neha-tanti: **⚠️ New:** This breadcrumb hand-copies `section.title` + `page.label`, which the route shell and `PageWireframeView` already derive from the manifest. That third copy is why this rename touched four sites. Render it once in `[section]/[page]/page.tsx` above the `kind` branch; drop it from the seven page components.

apps/docs/page-registry/index.ts
```diff
@@ -120,7 +124,7 @@ export const PAGE_REGISTRY = {
       },
       {
         slug: 'radius',
-        label: 'Radius · Borders · Shadows',
+        label: 'Radius, borders & shadows',
```
neha-tanti, line 127: **⚠️ New:** The rename only exists because the home card joins page labels with ` · `. Nothing stops the next label from reintroducing the collision. Assert in the generator that a registry label contains no ` · `, alongside the two guards this commit adds.

neha-tanti (commented):
## 🤖 Principal Architect Follow-up (round 3) — #700

### Verdict: NEEDS DISCUSSION

*(Posted as `COMMENT` — the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

| 1 | `unit` and page `components` could contradict, unguarded | `apps/docs/scripts/page-manifest.mjs:126` | ✅ Fixed | New assert has no hole at the layer it runs. |
| 2 | Negated equality on the `unit` union | `apps/docs/app/_lib/home-counts.ts:16` | ✅ Fixed | Widening the union reproduces `TS2366` at that line. |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/app/_pages/foundations/radius.tsx:59` | Breadcrumb hand-copies manifest data the route shell derives | Render it in `[section]/[page]/page.tsx` above the `kind` branch; drop it from the 7 pages |
| 2 | ⚠️ Minor | `apps/docs/page-registry/index.ts:127` | Label rename dodges the home card's ` · ` joiner; nothing enforces it | Assert in the generator that no registry label contains the joiner |

### Architectural Concerns

**Breadcrumb ownership** (`apps/docs/app/[section]/[page]/page.tsx`) — the same three-item trail is built from the manifest in `PageWireframeView` and the mdx branch, and by hand in seven `component` pages behind a `kind` discriminator. Lifting it above that branch removes the discriminator, the seven copies, and the class of desync this commit fixed by hand.

---

neha-tanti (commented):
## 🤖 SDE2 Follow-up (round 3) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine. See the Principal Architect follow-up for two minor open items.)*

### Unresolved Issues

All previous issues resolved.

| 1 | ` · ` joiner collided with the label's own separator | `apps/docs/page-registry/index.ts:127` | ✅ Fixed | Renamed at four sites; repo-wide grep finds no fifth copy. |
| 2 | Object-identity filter on the audience section | `apps/docs/app/page.tsx:46` | ✅ Fixed | Slug compare; `requireSection` still supplies the throw. |
| 3 | `unit` test duplicated the section-order test | `apps/docs/scripts/page-manifest.test.js:66` | ✅ Fixed | Folded; the merged assertion is strictly stronger. |

### Verification

| Command | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ passes |
| `npx vitest run --project unit apps/docs` | ✅ 44 passed across 6 files |
| `npx eslint apps/docs` | ✅ clean |
| `npx prettier --check` (7 changed files) | ✅ all match |
| Exhaustiveness probe — widen `unit` to `'components' \| 'tokens'` | ✅ `TS2366` at `home-counts.ts:16`, restored clean |

Four points checked and cleared without a row: the annotated-return `switch` is the standard exhaustiveness shape and needs no `assertNever`; the new throw's wording matches its sibling asserts; `if (page.components)` cannot differ from `?.length` because the generator never emits an empty array; and the two asserts are independent per-section checks, so ordering only picks which message surfaces on a double fault.

---

neha-tanti (commented):
## 🤖 Principal Architect Follow-up (round 4) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine. See the SDE2 follow-up for two minor open items.)*

### Unresolved Issues

All previous issues resolved.

| 1 | Breadcrumb hand-copied manifest data in 7 pages | `apps/docs/app/[section]/[page]/page.tsx:27` | ✅ Fixed | One owner per route shell; grep finds no fourth copy. |
| 2 | Label rename dodged the card joiner, unenforced | `apps/docs/scripts/page-manifest.mjs:129` | ✅ Fixed | Generator rejects the joiner; rebuild guard still passes. |

Three things were considered and deliberately not raised: `CARD_JOINER` duplicating the `' · '` literal (the middle dot is already a fixed docs separator, and threading a constant across the `.mjs`/app boundary adds more moving parts than it removes); `PageBody` living in the route file rather than `_components/` (both branches now take only `page`, own the same lookup-plus-resync-throw shape, and `PageBody` is unexported); and #701 impact (per-component pages route through `[section]/[page]` at the generator's two-level depth, so they inherit the crumb).

---

neha-tanti (commented):
## 🤖 SDE2 Follow-up (round 4) — #700

### Verdict: MINOR CHANGES

*(Posted as `COMMENT` — the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

### New Issues

| 1 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:127` | Joiner defined twice; the guard and `app/page.tsx:50` can drift | Emit it from the generated manifest module; import at the consumer |
| 2 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:133` | Message names a `getting-started` card the home page never renders | Drop the per-section wording from the error string |

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ pass |
| `npx vitest run --project unit apps/docs` | ✅ 45 tests / 6 files, incl. the new joiner fixture |
| `npx eslint apps/docs` | ✅ clean |
| `npx prettier --check` (11 changed files) | ✅ all match |
| `pnpm --filter @nexus_ds/docs build` | ✅ 44/44 static; only `/appearance-ssr` is ƒ, a pre-existing `force-dynamic` fixture |
| One `<nav>` per emitted page, all three kinds | ✅ 1 each in `foundations/radius`, `getting-started/install`, `components/inputs`; trails correct |
| `Exclude<ManifestPage, { kind: 'placeholder' }>` narrowing | ✅ `page.route` / `page.kind` usable, `file` typed `string`; `noUncheckedIndexedAccess` keeps the `!loadPage` throw reachable and before `await loadPage()` |
| Seven `_pages/` deletions | ✅ pure — no orphan imports, no stray blanks, every fragment still multi-child |

---

apps/docs/scripts/page-manifest.mjs
```diff
@@ -119,19 +119,49 @@ function comparisonKey(value) {
   return value.toLowerCase().replace(/[^a-z0-9]/g, '');
 }
 
-function assertNestedLabelsHaveNoPage(section) {
+function railLabelsOf(page) {
+  return [...(page.components ?? []), ...(page.nested ?? [])];
+}
+
+/** The home page's section cards join a section's page labels with this. */
+const CARD_JOINER = ' \u00b7 ';
```
neha-tanti: **⚠️ New — joiner defined twice:** `app/page.tsx:50` joins with a literal `' · '`; this guard hardcodes its own escaped copy, so the two can drift. Emit the joiner from the generated manifest module and import it at the consumer, and write it as a literal so a grep matches both.

apps/docs/scripts/page-manifest.mjs
```diff
@@ -119,19 +119,49 @@ function comparisonKey(value) {
...
+
+function assertLabelsAvoidTheCardJoiner(section) {
+  for (const page of section.pages) {
+    if (page.label.includes(CARD_JOINER)) {
+      throw new Error(
+        `${section.slug}/${page.slug} has "${CARD_JOINER}" inside its label, which is what the home page's ${section.slug} card joins page labels with — the card would read the label as several pages. Name the page without it.`
```
neha-tanti: **⚠️ New — message names a card that isn't rendered:** `app/page.tsx` excludes `getting-started` from the section cards, so this would tell a maintainer their `getting-started` card is broken. Reword generically: the home page joins a section's page labels with this.

neha-tanti (commented):
## 🤖 Principal Architect Follow-up (round 5) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine. See the SDE2 follow-up for one minor open item.)*

### Unresolved Issues

All previous issues resolved.

| 1 | Breadcrumb hand-copied manifest data in 7 pages | `apps/docs/app/[section]/[page]/page.tsx:33` | ✅ Fixed | Still manifest-derived; this commit did not disturb it. |
| 2 | Label rename dodged the card joiner, unenforced | `apps/docs/scripts/page-manifest.mjs:129` | ✅ Fixed | Guard survives the rework; one binding, no hole. |

### New Issues

None in this lane.

### Architectural Concerns

**Ownership boundary holds** (`apps/docs/scripts/page-manifest.mjs:394`) — the guard runs over every section using the same binding it emits at line 236, so the check and the render cannot diverge by construction rather than by convention. CI (`ci.yml:321`) keeps the emitted copy fresh.

**Client-graph placement is safe** (`apps/docs/app/_lib/manifest.ts:8`) — `CARD_JOINER` is a serializable string on a module already imported by `'use client'` components, and the server-only loaders stay behind `import 'server-only'` in `page-content.generated.ts`.

---

apps/docs/scripts/page-manifest.mjs
```diff
@@ -119,19 +119,49 @@ function comparisonKey(value) {
   return value.toLowerCase().replace(/[^a-z0-9]/g, '');
 }
 
-function assertNestedLabelsHaveNoPage(section) {
+function railLabelsOf(page) {
+  return [...(page.components ?? []), ...(page.nested ?? [])];
+}
+
+/** The separator the home page's section cards join page labels with. */
```
neha-tanti: **⚠️ New — doc drift:** This JSDoc and the emitted one at line 235 describe the same constant in different words. The commit removed the value duplication; the comment duplication is what is left. Drop this one — `assertLabelsAvoidTheCardJoiner` below already names what the joiner is for.

neha-tanti (commented):
## 🤖 SDE2 Follow-up (round 5) — #700

### Verdict: MINOR CHANGES

*(Posted as `COMMENT` — the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

| 1 | Joiner defined twice; guard and consumer could drift | `apps/docs/scripts/page-manifest.mjs:127` | ✅ Fixed | One declaration, emitted → re-exported → consumed. |
| 2 | Throw named a `getting-started` card never rendered | `apps/docs/scripts/page-manifest.mjs:133` | ✅ Fixed | Message drops `${section.slug}`'s card wording. |

### New Issues

| 1 | ⚠️ Minor | `apps/docs/scripts/page-manifest.mjs:126` | Generator-local JSDoc worded differently from the emitted one at `:235` | Drop the local comment; `assertLabelsAvoidTheCardJoiner` already names its purpose |

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ clean |
| `npx vitest run --project unit apps/docs` | ✅ 45 passed / 6 files |
| `npx eslint` (5 changed files) | ✅ clean |
| `npx prettier --check` (5 changed files) | ✅ all match |
| `generate:manifest` rebuild | ✅ `page-manifest.generated.ts` byte-identical |
| `JSON.stringify(CARD_JOINER)` emits a literal, not an escape | ✅ raw U+00B7 (`2220c2b72022`); grep ` · ` matches both ends |
| Reworked test fails for the right reason | ✅ removing the `manifest.forEach` guard call fails exactly that one test |

One point checked and cleared without a row: `home-counts.ts:24` also joins with a raw ` · `, but that separator joins count phrases rather than page labels — reusing `CARD_JOINER` there would couple two unrelated concerns.

---

neha-tanti (commented):
## 🤖 Principal Architect Follow-up (round 6) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

| 1 | Breadcrumb hand-copied manifest data in 7 pages | `apps/docs/app/[section]/[page]/page.tsx:33` | ✅ Fixed | Untouched by this commit; still manifest-derived. |
| 2 | Label rename dodged the card joiner, unenforced | `apps/docs/scripts/page-manifest.mjs:126` | ✅ Fixed | Guard still reads the declaration below the deleted line. |

### New Issues

None.

---

neha-tanti (commented):
## 🤖 SDE2 Follow-up (round 6) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

| 1 | Generator-local JSDoc worded differently from the emitted one | `apps/docs/scripts/page-manifest.mjs:126` | ✅ Fixed | Exactly that line deleted; emitted doc untouched. |

### New Issues

None.

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ clean |
| `npx vitest run --project unit apps/docs` | ✅ 45 passed / 6 files |
| `npx eslint apps/docs/scripts/page-manifest.mjs` | ✅ clean |
| `npx prettier --check` | ✅ matches |
| `generate:manifest` rebuild | ✅ `page-manifest.generated.ts` byte-identical |
| Guard test still fails for the right reason | ✅ `rejects a label carrying the separator the home card joins with` |
| Emitted JSDoc survives | ✅ intact above `export const CARD_JOINER` at `page-manifest.generated.ts:35` |

The `page-content.generated.ts` working-copy entry was confirmed an LF→CRLF notice with zero content lines changed, and the tree was restored.

---

neha-tanti (commented):
## 🤖 Principal Architect Follow-up (round 7) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

| 1 | Breadcrumb hand-copied manifest data in 7 pages | `apps/docs/app/[section]/[page]/page.tsx:32` | ✅ Fixed | Still drawn once from the manifest. |
| 2 | Label rename dodged the card joiner, unenforced | `apps/docs/scripts/page-manifest.mjs:128` | ✅ Fixed | Guard intact; `CARD_JOINER` still ships from the generated module. |

### New Issues

None.

### Architectural Concerns

**The JSDoc split falls on the right boundary** — `requireSection` has two call sites (`app/page.tsx` and its test) and no `'use client'` importer, so it is read by jumping to definition rather than by hover. The doc that was kept sits on the hand-authored data contract (`page-registry/index.ts`) and the generated type surface, which are read by hover and have no call site to consult.

**Server/client boundary unchanged** — `PAGE_MANIFEST` is still public nav metadata emitted as a plain serializable literal, and nothing in `f7ee041` moves data across the boundary.

---

neha-tanti (commented):
## 🤖 SDE2 Follow-up (round 7) — #700

### Verdict: APPROVED

*(Posted as `COMMENT`, not `APPROVE`: GitHub rejects an approving review on a self-authored PR and the `nexuslabs-ai-bot` account is not authenticated on this machine.)*

### Unresolved Issues

All previous issues resolved.

### New Issues

None.

### Verification

| Check | Result |
| --- | --- |
| `pnpm --filter @nexus_ds/docs typecheck` | ✅ clean |
| `npx vitest run --project unit apps/docs` | ✅ 45 passed / 6 files |
| `npx eslint` (4 changed files) | ✅ clean |
| `npx prettier --check` (4 changed files) | ✅ all match |
| `generate:manifest` rebuild | ✅ zero content lines changed |
| Comment accuracy — does `requireSection` fail the *build*? | ✅ `/` prerenders by default (no `dynamic` export, no `output` in `next.config.ts`), so module-scope evaluation throws during `next build` |

The comment line drawn by this commit checks out against `code-comments.md`. The two deleted blocks described rationale and callers, and what they carried is still reachable: `describeSize`'s tie to the registry is `switch (section.unit)` on the first line of its body, and `requireSection`'s throw already names the registry path. `countComponents` keeping its doc is correct — `page.components?.length || 1` is the non-obvious logic the rule exists for. The `page.tsx:42` rewording reads as "why not X" but its substance is what the line does: the value is definitionally `'getting-started'` and exists for the throw, so the comment prevents a reader flattening it to the literal.

---

PR #705, opened 2026-09-12
docs(install): rewrite install.mdx as the four-step setup page

neha-tanti:
> [!IMPORTANT]
> **Scope notice for reviewers — read before flagging missing commits.**
>
> This PR is **docs only**: `install.mdx`, `theme-setup.mdx`, `appearance.mdx`, and the
> `EXPECTED_HEADING_IDS` repin those heading renames require. Head is `1da639d2`.
>
> The four `test(docs):` commits `d7fb563c..83a2877e` were **deliberately removed** on
> 2026-09-16 and live in their own PR, tracked by #716. They are not missing, not
> unpushed, and must not be pushed back here. A previous follow-up round read their
> absence as a regression, asked for them to be restored, and the restore silently
> re-closed the split — please do not repeat that.
>
> **Review rounds 8–13 apply to the test harness, not to this PR.** Their stored
> `commit_id`s (`d7fb563c`, `bcdf7a51`, `e0f67457`, `83a2877e`) are intentionally
> unreachable from this head. Rounds 1–7 are the ones that reviewed this content.
>
> Nine open findings from rounds 11 and 13 are carried in #716 with their evidence;
> none were dropped.

---

## Summary

- Rewrites `apps/docs/content/getting-started/install.mdx` around the four things a pasted component reaches for that aren't in the file it came from, in order: **Tailwind 4** → **the token CSS layer** (+ the `color-scheme` policy) → **the appearance provider** → **the `cn` helper**.
- Deletes the two-delivery-model table, the "Copy/own is still landing" callout, and every `#541` reference. The page opens by stating copy/own plainly and says a pasted component receives no upstream updates.
- Closes with a short "Paste a component" section that walks `Button` and its three sibling files in, keeps the live `<Button>` preview, and names the two failure modes (unstyled → step 2 never reached the page; stuck light → step 3 isn't wrapping them).

## GitHub Issue

Closes #703

## Test Plan

- [x] `pnpm --filter @nexus_ds/docs build` — passes; `/getting-started/install` prerenders.
- [x] `prettier --check` on the page — clean.
- [x] Page renders correctly in the docs site, live Button preview included.
[...]

PR #718, opened 2026-09-17
test(docs): assert MDX anchor links resolve to real routes and ids

### LINKED ISSUE #716: test(docs): harden the MDX link-integrity test against App Router conventions
## Context

`apps/docs/mdx-options.test.ts` gained a link-integrity suite during review of #705. It walks `content/`, collects every in-repo `](/route#fragment)` link, and asserts that the route exists and that the fragment matches a heading id the MDX pipeline actually emits. It caught real breakage — the `#step-3-the-appearance-provider` pointer in `theming/appearance.mdx` would otherwise have gone stale silently.

The suite arrived as review fallout on a docs PR and grew across four commits and four review rounds, none of which touched the page #705 exists to rewrite. Splitting it out so the docs rewrite can merge against #703 and this can be finished against its own bar.

## Scope

Route existence is currently established by re-implementing Next's App Router conventions inside the test: walking `app/`, filtering `(group)`, `[dynamic]`, `_private`, and `@slot` segments, and deriving page filenames from `pageExtensions`. Every review round found another convention the walk does not model.

The design question worth settling first: keep modelling the routing spec, or assert against a narrower pinned surface. The docs app has three static routes; a pinned list may be the whole answer, and would retire most of the findings below.

## Open findings (PR #705, review round 11)

| # | Severity | File | Finding |
| --- | --- | --- | --- |
| 1 | ❌ Blocking | `mdx-options.test.ts` | Nothing asserts `staticRoutes()`. Setting `PAGE_EXTENSIONS = ['ts']` empties `STATIC_ROUTES` and all 13 tests still pass — links fall through to the `SECTIONS` branch in `routeExists`. |
| 2 | ⚠️ Minor | `mdx-options.test.ts` | `expect(config.pageExtensions).toEqual(PAGE_EXTENSIONS)` compares the same array reference to itself; `@next/mdx` shallow-copies the config. |
| 3 | ⚠️ Minor | `mdx-options.test.ts` | `FOOTNOTE_ID_PREFIX = 'user-content-'` over-filters. remark-rehype clobbers only footnote ids, but the bare prefix also matches an authored slug such as `user-content-policy`. Use `user-content-fn`. |
| 4 | ⚠️ Minor | `mdx-options.test.ts` | Interception markers read as route groups: the `startsWith('(')` filter drops `(..)photo`, so `app/feed/(..)photo/page.tsx` registers `/feed`. |
| 5 | ⚠️ Minor | `page-extensions.ts` | `.md` never compiles — `createMDX({ options })` passes no `extension` and `@next/mdx` defaults to `/\.mdx$/`. Listing `md` only makes Next treat `page.md` as a route with no loader. |
| 6 | ⚠️ Minor | `mdx-options.test.ts` | `SUB_PAGE_KEYS` re-implements `generateStaticParams` by hand, with a comment asserting the two agree. |
| 7 | ⚠️ Minor | `mdx-options.test.ts` | Key/slug divergence: keys derive from `section.slug`, but every handler resolves via `getSection()`, which indexes `SECTIONS` by object key. A section whose key and slug diverge builds a route that `notFound()`s while the suite stays green. Latent — all seven keys currently equal their slugs. |

## Acceptance criteria

- [ ] Settle the design question above before fixing findings individually.
- [ ] Route discovery has a failure mode — emptying or gutting it fails at least one test.
- [ ] No assertion compares a value to itself.
- [ ] The vacuity guards track authored anchors only, and each is proven by deleting the thing it guards.
- [ ] `page-extensions.ts` lists only extensions that actually compile, or `createMDX` is given a matching `extension`.
- [ ] One route authority: the test does not re-derive what `generateStaticParams` already emits.

## History

Carried over from PR #705 review rounds 8–11. Reviews: [round 11 architect](https://github.com/nexuslabs-ai/nexus/pull/705#pullrequestreview-5222601970), [round 11 SDE2](https://github.com/nexuslabs-ai/nexus/pull/705#pullrequestreview-5222603132).

INNOVATIVEGAMER:
> [!IMPORTANT]
> **Scope notice for reviewers.**
>
> This PR owns the MDX link-integrity test. It is **stacked on #705** — the pinned ids in
> `EXPECTED_HEADING_IDS` reference headings #705 introduces, so it cannot stand alone on
> `main`. Retarget to `main` after #705 merges.
>
> **Review rounds 8–13 on #705 apply to this code, not to that PR.** Their stored
> `commit_id`s are `d7fb563c`, `bcdf7a51`, `e0f67457`, and `83a2877e` — all reachable from
> this head. Do not ask #705 to restore them; that reverts the split.

## Summary

- Adds the anchor-link assertions to `apps/docs/mdx-options.test.ts`: every in-repo `](/route#fragment)` link in `content/` is checked against the route registries and against the heading ids the MDX pipeline actually emits.
- Splits `MDX_PAGES` out of `real-pages.tsx` so the test can read the registry.
- Adds `apps/docs/app/_lib/routes.ts` as the one owner of the docs route shape: both `generateStaticParams` bodies plus `sectionHref` / `subPageHref`. The test reads route authority from there instead of importing the page modules, so it no longer drags `Breadcrumb`, `SubPageView`, `REAL_PAGES`, and seven `_pages/*` components into a config test — and its `ROUTES` set is built from the same two functions the app renders hrefs with.
- Splits the app-directory walk into static routes and dynamic page directories, and pins the dynamic list. A third dynamic route now fails the suite until its params source is registered in `ROUTES`.
- Drops `slug` and `href` from `Section`. The object key is the only spelling of a section's route, so key/slug/href divergence is unrepresentable and the `keys every section by its own slug` assertion is retired.
- Types the section key. `sectionHref` / `subPageHref` take `SectionSlug` (`keyof typeof SECTIONS`), and `sections.ts` exports `SECTION_SLUGS` plus an `isSectionSlug` predicate for the callers that legitimately hold a runtime string. `getSection` takes a `SectionSlug` and returns a `Section`, which retires the `ALL_SECTIONS` cast and two guards that could not fire.
- `SectionLayout` and `LeftRail` take a section slug alone; `LeftRail` resolves `getSection` itself, the way `SubPageView` already does, so there is no slug/section pair to drift.
- The home page's section cards take a typed slug and read title, page count, and href from `SECTIONS`. The Theming card had drifted to `3 pages` against four subs, with a blurb that omitted Appearance.

## GitHub Issue

Closes #716

## Supersedes

Replaces #717, which GitHub auto-closed as "merged" when its base branch was fast-forwarded to this same head. No code was merged anywhere; the branch and commits are unchanged.

## Design notes

**Why `page-extensions.ts` and the `pageExtensions` config key are gone.** `@next/mdx` compiles `content/*.mdx` through its own `extension: /\.mdx$/` and never reads `pageExtensions` (see `node_modules/@next/mdx/index.js`). The `mdx` entry therefore only enabled a `page.mdx` route form no route uses; without it the value was Next's own default (`['tsx','ts','jsx','js']`), so the module and the config key both restated a framework default. The test now derives page filenames from that default directly, and asserts `config.pageExtensions` is unset so the assumption is enforced rather than commented.

**Why `generateStaticParams` delegates rather than re-exports.** `export const generateStaticParams = sectionParams` would also work at runtime, but the explicit one-line delegation is the form Next documents, so it stays out of the way of any static analysis.

**Why `route-params.ts` became `routes.ts`.** The module was introduced in this PR to hold the two `generateStaticParams` bodies. Review round 15 asked for the href builders to live beside them, at which point the module owns the route shape, not just the params — so the name follows the contents. There is no history cost: the file has never existed on `main`.

**Why the test shares the app's href builders.** Reading `sectionHref` / `subPageHref` does not make the link assertion tautological. `ROUTES` moves with the builder, but `content/` does not: pointing `subPageHref` at `/${section}/docs/${sub}` still fails with `theming/appearance.mdx links to /getting-started/install#… , which is not a route`. What the sharing buys is that the test can no longer check a shape the app has stopped emitting.

**Why the section key is typed and the sub slug is not.** `keyof typeof SECTIONS` is already a literal union, so typing the section param cost nothing. A matching union for `sub` is not free: the `subs` arrays are contextually typed by `SubPage`, so every `slug` widens to `string`, and recovering the literals needs `as const satisfies` plus `readonly` throughout `Section` / `SubPage` / `Block`, with `subPageHref` becoming generic in the section to look the sub union up. That is a heavier type surface than the one typo class it closes, and the sub half of every content link is already checked by the link assertion.

**Why `getSection` is a wrapper over an index.** It returns `SECTIONS[slug]` widened to the declared `Section`. Indexing directly yields the union of the seven literal entry types, where `nested` exists on some subs and not others — `sub.nested` then fails to compile in `LeftRail`. The wrapper is what keeps the optional fields readable at the call sites.

**Why `LeftRail` re-resolves the section.** It costs one `getSection` call and no guard: the prop is a `SectionSlug`, so the lookup cannot miss. Threading a `Section` object beside its slug would be two props that must agree, which is the invariant the architect flagged on `SECTIONS` itself.

**Why `[section]/page.tsx` guards twice.** The first guard is `isSectionSlug`, which narrows the URL segment before `getDefaultSub` and `subPageHref` see it — that narrowing is what makes the typed builders sound at a route boundary. The second is still `getDefaultSub`, which is `undefined` for a section with no subs; without it an empty section redirects to `/section/undefined`.

**Why the home cards derive title and count but not the blurb.** Title, page count, and href are registry facts, and the Theming card proved they drift when hand-written. The blurb is not derivable: joining sub labels with ` · ` renders Foundations as `Color · Typography · Spacing · Radius · Borders · Shadows · Layering · Responsive`, because one label carries its own separators. The Components card now reads `5 pages` rather than `5 groups · 17 components` — the component count is not a fact `SECTIONS` holds, and the stats row above the cards still carries it.

## Verification

- `npx vitest run --project=unit apps/docs/mdx-options.test.ts` — 15 passed.
- `npx vitest run --project=unit apps/docs` — 4 files, 26 tests, pass.
- `tsc --noEmit -p apps/docs/tsconfig.json` — exit 0. `eslint` and `prettier --check` on all 11 changed files — exit 0.
- Mutation evidence for the typed section key:
  - `slug: 'foundatons'` in `SECTION_CARDS` → `TS2820: Type '"foundatons"' is not assignable … Did you mean '"foundations"'?`
  - `sectionHref('agnts')` → `TS2345: Argument of type '"agnts"' is not assignable to parameter of type '"getting-started" | "agents" | …'`
  - Both spellings compiled clean before this round, and no test walks TSX hrefs, so neither had a failure mode.
- Earlier rounds:
  - `pageExtensions: ['tsx','mdx']` in `next.config.ts` → `expected [ 'tsx', 'mdx' ] to be undefined`.
  - `<Broken` appended to `theming/appearance.mdx` → `Error: failed to compile theming/appearance.mdx`, with the `97:1 Unexpected end of file…` parse error attached as `Caused by:`. Previously the parse error stood alone, naming a line and column but no file, and reported all tests as 0.
  - `subPageHref` emitting `/${section}/docs/${sub}` → the link test fails naming the linking file and the href.
  - Adding `app/[tag]/page.tsx` fails `pins every dynamic route to a registered params source`; breaking a cross-page fragment fails the link test with the linking file, href, and id list named.

## Test Plan

- [x] `npx vitest run --project=unit apps/docs/mdx-options.test.ts` — 15 passed at this head.
- [x] Full docs unit scope — 26 passed.
- [x] Typecheck, lint, and format clean on every changed file.
- [x] Every new/reshaped assertion mutation-tested, including the two shared href builders and the typed section key.


INNOVATIVEGAMER:
## All nine findings fixed

Pushed as `e9e00098`. The route-discovery design question from #716 resolved toward **one authority**: route existence now comes from the routes' own `generateStaticParams` rather than a re-implementation of them, which retired several findings rather than patching each.

| # | Finding | Fix |
| --- | --- | --- |
| 1 | `staticRoutes()` never asserted | Discovered set pinned to `['/', '/appearance-ssr', '/changelog']` |
| 2 | Tautological `pageExtensions` assertion | Pinned to the literal list |
| 3 | `user-content-` over-filtered authored slugs | Narrowed to `user-content-fn` |
| 4 | `(..)x` read as a route group | `(.` added to the non-routable prefixes |
| 5 | `.md` never compiles | Dropped from `PAGE_EXTENSIONS` |
| 6 | `SUB_PAGE_KEYS` re-derived `generateStaticParams` | Imported and mapped |
| 7 | Key/slug divergence | New assertion: every `SECTIONS` key equals its `slug` |
| 8 | `next.config` imported without the `TURBOPACK` stub | Merged into the stubbed test — one import, no cache coupling |
| 9 | `routeExists` resolved by key, not emitted routes | `routeExists` deleted; one `ROUTES` set built from both `generateStaticParams` |

Also trimmed the design-rationale comment blocks, per `code-comments.md`.

### Verification

Each fix mutation-tested — all were previously green under the mutation:

| Mutation | Before | After |
| --- | --- | --- |
| `PAGE_EXTENSIONS = ['ts']` (empties route discovery) | 13/13 passed | **2 tests fail** |
| Diverge a `SECTIONS` key from its slug (`theming` → `theming2`) | green | **fails** `SECTIONS key theming2: expected 'theming' to be 'theming2'` |
| Delete the sole authored anchor `#the-model` | — | **fails** `no authored same-page anchor checked` |
| Synthetic tree with `feed/(..)photo` | registered a false `/feed` | **excluded**; `(marketing)/pricing` → `/pricing`, `blog/page.mdx` → `/blog`, `_private`/`@slot`/`[section]` skipped |

`vitest run --project=unit apps/docs/mdx-options.test.ts` → 14/14. `pnpm --filter @nexus_ds/docs typecheck` clean. `eslint --max-warnings 0` and `prettier --check` clean.

INNOVATIVEGAMER:
Closing: the link-integrity test isn't worth its cost to the system as it stands.

The docs IA is still wireframe-stage placeholder content, so the class of bug this guards against is cheap to catch by other means and rare in practice. The test also grew a source-grep gate over TSX call sites that is fail-open by construction — it would need ongoing maintenance to stay honest, which is more upkeep than the bug class justifies right now.

The route refactor that accumulated here (typed section keys, `routes.ts` as the single href owner, un-exported `SECTIONS`) is not lost — it can be re-proposed on its own merits if we want it, separate from the test that motivated it.

Branch `neha-tanti/issue-716-mdx-link-integrity-test` is left intact.

